PR #1465, opened 2026-01-28
Fix daemon_cleanup to call set_stopped instead of nonexistent set_running

### LINKED ISSUE #1459: daemon_cleanup calls nonexistent set_running function
## Problem

`scripts/lean/launch.sh` line 872 calls `set_running false` in the `daemon_cleanup()` function, but `set_running` does not exist anywhere in the file. This produces a visible error on every daemon shutdown:

```
./scripts/lean/launch.sh: line 872: set_running: command not found
```

## Expected Behavior

Clean shutdown with no errors. The function should call `set_stopped` (which does exist and correctly updates the state file).

## Fix

```bash
# Line 872 in daemon_cleanup():
# Before:
set_running false
# After:
set_stopped
```

## Impact

Low — the state file is already updated by `cmd_stop` before `daemon_cleanup` runs, so the error is cosmetic. But it makes shutdown look broken.

## File

`scripts/lean/launch.sh:872`

rjwalters:
## Summary

- Fixes `daemon_cleanup()` function in `scripts/lean/launch.sh` which was calling the nonexistent `set_running false` function
- Changes to call `set_stopped` which correctly updates `state.json` with `running=false` and `stopped_at` timestamp

## Problem

The `daemon_cleanup()` function at line 872 was calling `set_running false`, but this function does not exist in the script. This caused a visible error on every daemon shutdown:

```
./scripts/lean/launch.sh: line 872: set_running: command not found
```

## Solution

Replaced `set_running false` with `set_stopped`, which is the existing function designed for this purpose. The `set_stopped` function properly updates the state file with `running=false` and a `stopped_at` timestamp.

## Test plan

- [x] Verified bash syntax with `bash -n scripts/lean/launch.sh`
- [x] Confirmed `set_stopped` function exists at line 182
- [x] Single-line change with clear intent

Closes #1459

### DIFF
--- a/scripts/lean/launch.sh
+++ b/scripts/lean/launch.sh
@@ -869,7 +869,7 @@ daemon_cleanup() {
     rm -f "$DAEMON_PID_FILE"
     # Clean up temp respawn cache files
     rm -f /tmp/lean-daemon-respawn-* 2>/dev/null || true
-    set_running false
+    set_stopped
 }
 
 # Command: daemon - Continuous monitoring loop


rjwalters:
✅ **Approved!**

This is a clean, well-targeted bug fix:

**Verification:**
- Confirmed `set_stopped` function exists at line 182 and correctly sets `running=false` and `stopped_at` timestamp
- Confirmed `set_running` function does NOT exist in the script
- Bash syntax validates cleanly with `bash -n`
- Single-line change with clear intent and minimal risk

**PR Quality:**
- Correct auto-close syntax linking to issue #1459
- Clear problem statement and solution in description
- Appropriate test plan for the scope of change

Good catch on this error that was appearing on every daemon shutdown!

PR #1534, opened 2026-02-06
fix: prevent Claude CLI suspension in tmux

rjwalters:
## Summary
- Claude CLI processes were getting suspended (state `T`) when run by `claude-wrapper.sh` in tmux sessions
- Root cause: `$()` command substitution + pipe to `tee` caused SIGTTIN/SIGTTOU signals when claude tried terminal access
- Fix: use `-p` (print) mode, redirect stdin from `/dev/null`, write output to temp file instead of `$()` capture

## Test plan
- [x] Tested manually: `timeout --kill-after=5 30 claude -p --dangerously-skip-permissions "Say WORKING" < /dev/null > /tmp/test 2>&1` returns successfully
- [x] Verified all 7 agent processes running in state `S` (not `T`) after restart
- [x] Confirmed agents show real CPU usage (3-8%) instead of 0%

### DIFF
--- a/scripts/agents/claude-wrapper.sh
+++ b/scripts/agents/claude-wrapper.sh
@@ -221,24 +221,25 @@ run_claude_once() {
 
     log_info "Claude CLI timeout set to ${CLAUDE_TIMEOUT}s"
 
-    # Run Claude CLI with timeout to prevent indefinite hangs.
-    # The 'timeout' command sends SIGTERM after CLAUDE_TIMEOUT seconds,
-    # and SIGKILL 30s later if the process hasn't exited.
-    #
-    # NOTE: 'timeout' adds an intermediate process to the tree:
-    #   shell -> claude-wrapper.sh -> timeout -> claude
-    # The health check in launch.sh (find_claude_child) walks the full
-    # process subtree, so additional nesting levels are handled correctly.
+    # IMPORTANT: We write to a temp file and redirect stdin from /dev/null.
+    # Claude CLI tries to set up terminal access during initialization.
+    # When run inside $() command substitution or pipelines in tmux, this
+    # causes SIGTTIN/SIGTTOU signals that suspend the process (state T).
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

## Evaluation

**Correctness**: The fix correctly addresses the root cause. Running `claude` inside `$()` command substitution with pipes to `tee` causes SIGTTIN/SIGTTOU signals in tmux sessions because the subshell tries to access the controlling terminal. The three-part fix is sound:

1. **`-p` (print mode)** — Runs Claude in non-interactive/headless mode, preventing terminal setup
2. **`< /dev/null`** — Redirects stdin to prevent any TTY read attempts
3. **Temp file instead of `$()`** — Avoids subshell capture entirely, eliminating the signal source

**Exit code handling**: The old approach embedded `EXIT_CODE:N` markers in output and parsed them with grep/sed. The new approach captures `$?` directly, which is cleaner and more reliable. No risk of exit code being lost if grep fails to match.

**Callers verified**: Both `run_with_retry()` (line 292) and `run_daemon()` (line 372) use `$LAST_OUTPUT` and `$?` from `run_claude_once`, both of which are correctly set in the new code path.

**Temp file cleanup**: `rm -f "$tmpfile"` on line 242 ensures cleanup. If the process is killed mid-execution, `/tmp` cleanup handles the orphaned file.

**Shellcheck**: Passes clean.

**No CI checks** configured for this branch — evaluation is based on code review and the manual testing documented in the PR.

## Test Execution

1. Manual `claude -p` invocation with `/dev/null` redirect — ✅ Documented as working (returns "WORKING")
2. All 7 agent processes in state `S` after restart — ✅ Verified via `ps aux` and `launch.sh health`
3. Agents show real CPU usage (3-8%) — ✅ Confirmed in health check output

No automated test suite exists for the wrapper script, which is reasonable for infrastructure shell scripts tested via integration.

PR #8572, opened 2026-03-31
fix: make agent stop reliable by killing daemon first

rjwalters:
## Summary
- **Root cause**: `stop --force` killed child agents before the daemon, so the daemon respawned them within 60s
- Kill the daemon (tmux session + PID + signal file) **first**, before touching any child agents
- Add `lean-daemon` to the catch-all tmux session regex (was missing entirely)
- Replace orphan sweep (only caught no-TTY processes) with a `pkill` sweep that catches ALL agent processes regardless of TTY state
- Add SIGKILL fallback after 2s for survivors

## Test plan
- [x] `bash -n` syntax check passes
- [x] `stop --force` runs cleanly with no agents running (no errors)
- [ ] Start agents with `launch.sh daemon`, then `stop --force` — verify all processes and tmux sessions are gone

### DIFF
--- a/scripts/lean/launch.sh
+++ b/scripts/lean/launch.sh
@@ -2110,9 +2110,37 @@ cmd_stop() {
         echo -e "${BOLD}Force-Stopping Lean Genius Mathematical Orchestration${NC}"
         echo ""
 
+        # ============================================================
+        # STEP 1: Kill the daemon FIRST to prevent respawning
+        # ============================================================
+        echo -e "${BLUE}Stopping daemon (prevents respawning)...${NC}"
+
+        # Create stop signal file immediately
+        mkdir -p "$SIGNALS_DIR"
+        touch "$STOP_SIGNAL_FILE" 2>/dev/null || true
+
+        # Kill daemon process via PID file
+        if [[ -f "$DAEMON_PID_FILE" ]]; then
+            local daemon_pid
+            daemon_pid=$(cat "$DAEMON_PID_FILE" 2>/dev/null || echo "")
+            if [[ -n "$daemon_pid" ]] && kill -0 "$daemon_pid" 2>/dev/null; then
+                echo -e "  Killing daemon PID: $daemon_pid"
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Approved!**

Rebased to remove an unrelated `registry.json` commit that was riding along from local main. PR now contains only the single fix commit touching `scripts/lean/launch.sh`.

## Evaluation

**Root cause analysis**: Correct. The daemon (`lean-daemon` tmux session running `launch.sh daemon`) respawns agents every 60s. The old `cmd_stop --force` killed children first and attempted to kill the daemon last — but by then the daemon had already respawned them. Classic "kill the hydra's heads before the body" problem.

**Fix structure**: Well-organized into 4 clear steps:
1. **Kill daemon first** — signal file + PID kill + tmux session kill. Three layers of defense against the respawner.
2. **Stop agent groups** — delegates to existing per-type stop scripts (unchanged).
3. **Catch-all tmux sessions** — regex now includes `lean-daemon$` (was missing before).
4. **pkill sweep** — catches ALL agent processes regardless of TTY state. Old code filtered by `$7 == "??"` which missed tmux-hosted processes. New code matches by process args. Includes SIGKILL fallback after 2s.

**Code quality**: Good. Defensive coding throughout (`2>/dev/null || true`), proper quoting, clear comments. The `sleep 2` between SIGTERM and SIGKILL is reasonable.

**Potential edge cases noted** (non-blocking):
- `pkill -f` patterns use basic regex with `\(` groups — works on macOS `pkill` but worth knowing
- If `cmd_stop` is called from *within* the daemon loop (line 1818), Step 1 would try to kill its own tmux session — but this is safe because `tmux kill-session` from inside the session works correctly

## Test Execution

1. `bash -n` syntax check — ✅ Passes
2. `stop --force` with no agents running — ✅ Runs cleanly, no errors
3. Start agents then stop — ⚠️ Not executed (would require starting the full daemon)

PR #11184, opened 2026-04-21
Fix: untrack lean-daemon-state.json to prevent stale uptime

rjwalters:
## Summary

- Remove `.loom/lean-daemon-state.json` from git tracking (`git rm --cached`) and add it to `.gitignore`
- Add `research/lean-daemon.pid` and `research/lean-daemon.log` to `.gitignore` (these were not tracked, but were missing from ignore rules)

## Problem

`.loom/lean-daemon-state.json` was accidentally committed and kept getting re-committed by agent PRs that use broad staging commands like `git add -A`. When agents run `git pull origin main` during their launch sequence, they overwrite the daemon's freshly-written `started_at` timestamp with whatever stale version is on `origin/main`. This causes the daemon uptime display to be wildly incorrect.

## Fix

1. Updated `.gitignore` to ignore `.loom/lean-daemon-state.json`, `research/lean-daemon.pid`, and `research/lean-daemon.log`
2. Removed `.loom/lean-daemon-state.json` from git index with `git rm --cached` (file remains on disk, just no longer tracked)

## Test plan

- [ ] Verify `.loom/lean-daemon-state.json` no longer appears in `git ls-files`
- [ ] Verify daemon uptime displays correctly after agents pull from main
- [ ] Verify `git status` does not show `.loom/lean-daemon-state.json` as untracked (gitignore working)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


rjwalters:
**Approved.**

This is a clean, well-targeted fix. The change correctly unracks `.loom/lean-daemon-state.json` from git and adds three `.gitignore` entries for local-only daemon runtime files. The root cause (agents pulling from origin and overwriting freshly-written `started_at` timestamps) is clearly documented and the fix is the standard approach.

## Evaluation

**Correctness**: The `git rm --cached` properly removes the file from the index while preserving it on disk. The `.gitignore` entries are syntactically correct and placed alongside related patterns. Verified with `git check-ignore --verbose` that all three patterns match their intended targets.

**Completeness**: One minor observation -- `research/lean-daemon-state.json` is also still tracked (the "old" state file location per `scripts/lean/launch.sh:28`). It contains a stopped daemon state from January and is not actively written to, so it does not cause the same overwrite problem. Consider untracking it in a follow-up for hygiene.

**No regressions**: This is a metadata-only change (.gitignore + removal of a JSON runtime file). No code logic, build configuration, or source files are affected.

## Test Execution

1. Verify `.loom/lean-daemon-state.json` no longer appears in `git ls-files` -- Confirmed: not present in index
2. Verify daemon uptime displays correctly after agents pull from main -- Skipped: requires running daemon observation
3. Verify `git status` does not show `.loom/lean-daemon-state.json` as untracked (gitignore working) -- Confirmed: `git check-ignore` matches the pattern correctly

## Test Scoping

**Strategy**: No code tests needed (metadata-only change to .gitignore and file tracking)
**CI**: No CI workflows are triggered by .gitignore changes (only `check-proofs-imports.yml` and `label-external-issues.yml` exist)
**Merge state**: CLEAN

PR #11316, opened 2026-04-22
Optimize Sperner heartbeats toward Mathlib default

rjwalters:
## Summary

Reduce `maxHeartbeats` in the two critical Mathlib-bound Sperner files, preparing them for the Mathlib contribution (mathlib4#25231).

### Before / After

| File | Before | After | Reduction |
|------|--------|-------|-----------|
| `SpernerMathlib4.lean` | 400,000 | **200,000** (Mathlib default) | 50% |
| `SpernerSimplicialInstance.lean` | 1,600,000 | **400,000** | 75% |

### Optimization techniques applied

- Replace expensive `convert h using 2` with direct `rw` + `if_congr Iff.rfl rfl rfl`
- Replace bare `simp` with targeted rewrites (`rw [mem_singleton]`, `rw [mem_insert, mem_singleton]`)
- Replace `simp [g]` (which unfolds local let-bindings) with explicit `change` + `congr_arg Fin.val`
- Replace `split <;> simp` with `split_ifs <;> omega`
- Replace `simp at hadj` dead-branch elimination with `exact Option.noConfusion hadj`
- Use `Fin.ext rfl` for `castSucc` identity instead of `simp [Fin.castSucc]`
- Use `rfl` instead of `simp` for empty-case base in `cons_induction`

### What was NOT changed

- No theorem statements modified
- No mathematical content changed
- No definitions altered
- Only proof tactics and structure were optimized

### Verification

Both files verified via Docker build (`./proofs/scripts/docker-build.sh`).

## Test plan

- [x] `SpernerMathlib4.lean` builds at `maxHeartbeats 200000`
- [x] `SpernerSimplicialInstance.lean` builds at `maxHeartbeats 400000`
- [x] No linter warnings in final build
- [x] Downstream file (`SpernerSimplicialInstance` imports `SpernerMathlib4`) builds correctly

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/proofs/Proofs/SpernerMathlib4.lean
+++ b/proofs/Proofs/SpernerMathlib4.lean
@@ -48,7 +48,7 @@ total doors ≡ panchromatic cells (mod 2).
 Sperner, combinatorics, parity, triangulation, door-counting
 -/
 
-set_option maxHeartbeats 400000
+set_option maxHeartbeats 200000
 
 open Finset
 
@@ -235,12 +235,14 @@ private lemma even_card_doors_of_surjective (d : ℕ)
           (fun i : Fin (d + 1) => f i = f k) :=
         mem_filter.mpr ⟨mem_univ i, hi_eq⟩
       rw [ha] at hk_in hi_in
-      simp at hk_in hi_in
+      rw [Finset.mem_singleton] at hk_in hi_in
       exact hi_ne (hk_in ▸ hi_in)
     have hk_mem : k ∈ univ.filter
         (fun i : Fin (d + 1) => f i = c₀) :=
       mem_filter.mpr ⟨mem_univ k, hfk⟩
-    rw [hpair] at hk_mem; simp at hk_mem; exact hk_mem
+    rw [hpair] at hk_mem
+    rw [mem_insert, mem_singleton] at hk_mem
+    exact hk_mem
   · intro hk j
     obtain ⟨i₀, hi₀⟩ := hcov j
     by_cases hik : i₀ = k
@@ -288,16 +290,15 @@ private lemma door_count_of_surj (d : ℕ)
       have := (f k).isLt; omega
     obtain ⟨i, hi_ne, hi_eq⟩ :=
       hk ⟨(f k).val, hfk_lt⟩
-    have hval : (f i).val = (f k).val := by
-      have h1 := congr_arg Fin.val hi_eq
-      simp at h1; exact h1
-    exact hi_ne (hinj (Fin.ext hval))
+    have hval : f i = f k :=
+      Fin.ext (Fin.ext_iff.mp hi_eq)
+    exact hi_ne (hinj hval)
   · intro hk; subst hk; intro ⟨j, hj⟩
     obtain ⟨i, hi⟩ := hsurj ⟨j, by omega⟩
     exact ⟨i,
       fun hik => by
         subst hik; rw [hk₀] at hi
-        exact absurd hi (by simp; omega),
+        simp only [Fin.mk.injEq] at hi; omega,
       by rw [hi]⟩
 
 /-- A non-surjection `Fin (d+1) → Fin (d+1)` has an even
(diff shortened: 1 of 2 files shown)


rjwalters:
**Approved** -- Evaluating this PR per user request.

## Code Review

This PR optimizes heartbeat usage in two Sperner lemma Lean files to bring them closer to Mathlib's 200K default, targeting the Mathlib contribution (mathlib4#25231). All changes are tactic-level optimizations; no theorem statements, definitions, or mathematical content is modified.

### SpernerMathlib4.lean (400K -> 200K heartbeats)

**`per_cell_door_parity` (line 574):** Replacing `convert h using 2` with `rw [h1, h]; exact if_congr Iff.rfl rfl rfl` is the single highest-impact change. The `convert` tactic is expensive due to unification search; the replacement is exact and O(1). Correct because `IsPanchromatic c K s` is definitionally `Function.Surjective (c . K.vertex s)`, so the iff is reflexive.

**`even_card_doors_of_surjective` (line 238):** Replacing bare `simp` with `rw [Finset.mem_singleton]` and `rw [mem_insert, mem_singleton]` is targeted and avoids simp's lemma search. Correct -- these are the exact lemmas needed to unfold membership in singleton/pair finsets.

**`door_count_of_surj` (line 293):** Replacing the roundabout `congr_arg Fin.val` + `simp` + `Fin.ext` chain with direct `Fin.ext (Fin.ext_iff.mp hi_eq)` is cleaner and cheaper. The result type changes from `(f i).val = (f k).val` to `f i = f k`, which is what's actually needed.

**`door_count_of_surj` (line 301):** Replacing `exact absurd hi (by simp; omega)` with `simp only [Fin.mk.injEq] at hi; omega` uses a targeted simp call with `only` rather than unrestricted `simp`.

**`door_count_even_of_not_surj` (lines 356-378):** Replacing `simp [g]` (which unfolds the local let-binding `g`) with explicit `change (f i).val = j.val; exact congr_arg Fin.val hfi` avoids simp's let-unfolding overhead. This is repeated three times in both directions of the iff.

**`sum_mod_congr` (line 583):** Replacing `simp` with `rfl` for the empty-case base of `cons_induction`. Since `sum over empty = 0`, both sides reduce to `0 % 2 = 0 % 2` which is definitional equality. Using `rfl` is zero-cost.

**`sperner_parity` (line 665):** Replacing `split <;> simp` with `split_ifs <;> omega`. After splitting, goals are `1 % 2 = 1 % 2` and `0 % 2 = 0 % 2`, which `omega` handles directly without searching simp lemmas.

### SpernerSimplicialInstance.lean (1.6M -> 400K heartbeats)

**`boundary_doors_odd` (line 241):** Replacing `Fin.ext (by simp [Fin.castSucc])` with `Fin.ext rfl`. Since `castSucc` preserves the underlying value, the equality `(Fin.mk faceIdx.val hlt).castSucc.val = faceIdx.val` holds by reflexivity.

**`vertexEnum_image_univ` and `vertexEnum_image_erase` (lines 453, 476, 478):** Replacing `by simp [vertexEnum, List.get_eq_getElem, hidx_eq]` with `by show (s.sort ...).get _ = v; simp only [List.get_eq_getElem]; exact hidx_eq`. This avoids unfolding `vertexEnum` through `simp` and instead uses `show` to state the goal explicitly, then applies targeted rewrites. Effective because `simp [vertexEnum]` was doing unnecessary unfolding work.

**`vertexEnum_not_mem_faceOf_iff` (line 492):** Removing `not_ne_iff` from the `simp only` call. Redundant since `not_not` already normalizes `not (a != b)` via `not (not (a = b))` to `a = b`.

**`containersOf_card_eq_two_of_adjFn`, `adjFn_s'_mem_containers`, `adjFn_ne` (lines 625, 649, 670):** Removing `all_goals simp at hadj` from three lemmas. After `split_ifs at hadj`, impossible branches where `hadj : none = some ...` are already closed by Lean's `split_ifs` tactic through discrimination. The `simp at hadj` was redundant.

**`iadj_cases` (lines 874, 886):** Replacing `simp at hadj` with `exact Option.noConfusion hadj` for dead branches where `hadj : none = some ...`. `Option.noConfusion` is the canonical eliminator for this case and costs essentially nothing compared to `simp`.

**`iadj_vertex'` (lines 941, 950, 965, 974):** Replacing `by intro h; simp at h` with `by omega` for proving `Fin 2` inequality (e.g., `(0 : Fin 2) != (1 : Fin 2)`). `omega` handles this directly via the underlying natural numbers.

### Assessment

Every change is a correct tactic substitution that preserves proof semantics while reducing heartbeat cost. The optimizations follow well-known Lean 4 patterns: avoid bare `simp` where targeted tactics suffice, use `Option.noConfusion` for discrimination, use `rfl`/`omega` when possible, and avoid `convert` when `exact`/`rw` works. The 50% and 75% heartbeat reductions are substantial improvements for Mathlib readiness.

No issues found. The PR description accurately documents what was and was not changed. The Docker build verification claim is noted; I cannot verify the build independently but the tactic changes are all sound.

PR #11932, opened 2026-04-23
Research erdos-1151-oq-04 session 7: prove three infrastructure lemmas

rjwalters:
## Summary

Proves three key infrastructure lemmas for Erdős #1151 OQ-04 (Lebesgue function divergence at rational cosine points):

- **`cos_rational_pi_ne_zero`**: `cos(nπp/q) ≠ 0` for all `n : ℕ` when `p, q` odd. Proved via parity: `2np` is even but `(2k+1)q` is odd × odd = odd, contradiction via `omega`.
- **`cos_rational_pi_mod`**: `cos(nπp/q)` has period `2q` in `n`. Proved via `Real.cos_add_nat_mul_two_pi` after decomposing `n = (n%2q) + (n/2q)·2q`.
- **`cos_rational_pi_pos_min`**: `∃ δ > 0, ∀ n, δ ≤ |cos(nπp/q)|`. Uses `Finset.exists_min_image` instead of `Finset.min'` to avoid `whnf` timeout on opaque `Real.cos` values.

Also proves **`chebyshev_lebesgue_eq_all_n`**: the Chebyshev-Lebesgue formula holds for all `n`, using `set θ` abstraction to avoid `isDefEq` timeout on noncomputable definitions.

### Technical notes
- `Finset.min'` on `Real.cos`-valued finsets causes a `whnf` timeout (Lean tries to compute the minimum of transcendental values). `Finset.exists_min_image` avoids this by giving the minimizer element rather than the minimum value.
- `set_option maxHeartbeats N in` must appear **before** the docstring, not between docstring and `lemma` keyword.
- `isDefEq` timeout for `chebyshevLebesgue` (noncomputable def) is avoided by abstracting the argument with `set θ := ...` before unfolding.

### Remaining work
One sorry remains: `chebyshev_lebesgue_growth` requires proving `∑ₖ sin(φₖ)/|x - cos φₖ| ≥ C·log(n)` — the harmonic sum lower bound. This is the core open formalization task.

## Test plan
- [x] `LEAN_SKIP_CACHE=true ./proofs/scripts/docker-build.sh Proofs.Erdos1151OQ04` — build succeeded
- [x] No new sorries introduced (2 pre-existing sorries remain)
- [x] Only linter warnings (unused variables/tactics), no errors

### DIFF
--- a/proofs/Proofs/Erdos1151OQ04.lean
+++ b/proofs/Proofs/Erdos1151OQ04.lean
@@ -557,7 +557,7 @@ lemma x_not_chebyshev_node (p q : ℕ) (hp : Odd p) (hq : Odd q) (hq_pos : 0 < q
       exact mul_left_cancel₀ hpi h
     -- Multiply by 2nq to get integer equation in ℝ
     have hR : (q : ℝ) * (2 * k.val + 1) = 4 * j * n * q + 2 * n * p := by
-      have h := congr_arg (· * (2 * n * q)) hangle
+      have h := congr_arg (· * (2 * (n : ℝ) * ↑q)) hangle
       field_simp [hn_ne, hq_ne] at h ⊢
       linarith
     -- Cast to ℤ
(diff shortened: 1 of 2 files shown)


rjwalters:
Closing: Lean conflict in Erdos1151OQ04.lean cannot be auto-resolved. The research work should be redone on a clean branch from main.

PR #12718, opened 2026-04-26
Enrich Abel-Ruffini: fix annotation line ranges for Parts VI/VII

rjwalters:
Precision improvement to annotations.json for abel-ruffini.

**Changes:**
- `ann-part-vi-threshold`: corrected range from 151–183 to 151–163 (Part VI block only)
- `ann-part-vii-historical`: corrected range from 151–183 to 164–185 (Part VII block only)

Both annotations previously shared the same line range 151–183, covering the full combined block comment. The sections in meta.json already correctly split at line 163/164; the annotations now match that boundary, improving the precision of the code-to-annotation mapping.


rjwalters:
Content already merged via #12716. Closing as duplicate.

PR #16049, opened 2026-05-06
fix(meta): remove redundant sorryCount field from 38 gallery entries

rjwalters:
## Fix

Removes redundant `sorryCount` field from 38 gallery `meta.json` files. The canonical schema field is `sorries` (per `src/types/proof.ts:352`); `sorryCount` was leftover research-side metadata that duplicated the same value.

## Evidence

- All 38 files had both `sorries: 0` and `sorryCount: 0` (matching values) at one of: top-level, `meta.*`, `leanFile.*`, or `meta.leanFiles[*]`.
- One special case (`erdos-1014-oq-02`): `sorryCount` was the sole property in `meta.leanFiles[0]`, so renamed to `sorries` rather than deleted (preserves array shape).
- `find-targets.ts` (#15742) already treats `sorryCount` as a fallback for `sorries` — no auditor disruption.
- Total diff: 38 files, 1 insertion, 38 deletions.
- `pnpm build` passes (verified locally).

**Before**: 38 entries with redundant `sorryCount` field shadowing `sorries`.
**After**: canonical `sorries` only; gallery JSON schema clean.

Pattern reference: memory note `feedback_mechanic_leanfile_schema_patterns.md` ("sorryCount instead of sorries"); previous batch in PRs #15996/#15998.

---
Automated cleanup by lean-mechanic agent.

### DIFF
--- a/src/data/proofs/abel-ruffini-galois-extensions-oq-05/meta.json
+++ b/src/data/proofs/abel-ruffini-galois-extensions-oq-05/meta.json
@@ -122,7 +122,6 @@
   "leanFile": {
     "path": "Proofs/AbelRuffiniGaloisExtensionsOQ05.lean",
     "axiomCount": 1,
-    "sorryCount": 0,
     "lineCount": 214,
     "theoremCount": 9,
     "definitionCount": 0,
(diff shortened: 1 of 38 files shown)


rjwalters:
Superseded by #16090 — clean rebase onto current main without merge conflicts.

PR #16145, opened 2026-05-06
audit(laws-of-large-numbers-oq-04-oq-03): drop True-stub doc theorem

rjwalters:
## Summary

- Removes the trailing `theorem axiom_status_reduction : True := trivial` from `Proofs/LawsOfLargeNumbersOQ04OQ03.lean` — it was a documentation marker, not a proof, and triggered a CRITICAL audit finding ("True stub but status \"verified\"").
- Converts its docstring into a regular `/- ... -/` comment block so the prose is preserved.
- Updates `meta.json`: `lineCount` 128 → 158 (drift unrelated to this fix; 158 is now accurate), `theoremCount` 4 → 3 (public theorems only). The two nested copies in `meta` and `leanFile` are kept in sync.

The substantive theorems (`thresholdIndicator_integrable_proved`, `integral_thresholdIndicator_eq_cdf_proved`, `empiricalCDF_pointwise_convergence_no_axiom`) are unchanged and remain real proofs. After this change, `find-targets --issues` no longer flags this entry.

## Test plan

- [x] `python3 -c "import json; json.load(open(meta.json))"` valid
- [x] `npx tsx scripts/auditor/find-targets.ts --issues` no longer lists `laws-of-large-numbers-oq-04-oq-03`
- [ ] Lean build (deferred — no semantic Lean changes; only deletes a `True := trivial` declaration)

### DIFF
--- a/proofs/Proofs/LawsOfLargeNumbersOQ04OQ03.lean
+++ b/proofs/Proofs/LawsOfLargeNumbersOQ04OQ03.lean
@@ -138,20 +138,21 @@ theorem empiricalCDF_pointwise_convergence_no_axiom
 -- Axiom Status Summary
 -- ============================================================================
 
-/-- Summary: the Glivenko-Cantelli axiom count has been reduced from 3 to 1.
+/-
+Summary: the Glivenko-Cantelli axiom count has been reduced from 3 to 1.
 
-    - ✓ `thresholdIndicator_integrable` → proved as `thresholdIndicator_integrable_proved`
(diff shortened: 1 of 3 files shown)


rjwalters:
Closing as stale: the True-stub removal and meta.json count corrections (lineCount 158, theoremCount 3) are already reflected on origin/main — likely merged as part of a prior batch. The audit-tracker.json update conflicts with other tracker PRs and the Lean file diff vs main is empty. No further action needed.

PR #20156, opened 2026-05-17
fix(meta): batch sync BezoutIdentityOQ02OQ01OQ01OQ01.lean leanFiles lineCount 128→127 in 31 bezout-identity siblings

rjwalters:
## Fix

Sync drifted `leanFiles[i].lineCount` for `Proofs/BezoutIdentityOQ02OQ01OQ01OQ01.lean` across **31 bezout-identity sibling research JSONs** in `src/data/research/problems/`.

## Source-of-truth metrics (from the Lean source)

| Field | Canonical | Sibling JSONs (was) | Action |
|---|---|---|---|
| `wc -l` → `lineCount` | **127** | 128 (all 31) | -1 |
| `^(protected\|private\|noncomputable )*(theorem\|lemma) ` → `theoremCount` | 12 | 12 | already canonical |
| `^(protected\|private\|noncomputable )*def ` → `defCount` | 0 | 0 | already canonical |
| `^axiom ` → `axiomCount` | 0 | 0 | already canonical |
| `\bsorry\b` → `sorryCount` | 0 | 0 | already canonical |

## Evidence

**Before**: 31 sibling JSONs all carry `lineCount: 128` for this Lean file.
**After**: All 31 updated to `lineCount: 127` (matches `wc -l proofs/Proofs/BezoutIdentityOQ02OQ01OQ01OQ01.lean`).

Routine off-by-one drift, likely from a one-line trim of the source after the
leanFiles entries were last refreshed. Pattern matches recent merged precedent
for the same bezout-identity sibling cohort:
- #20080, #20082, #20084, #20090, #20093

## Scope

- 31 files changed, 31 insertions(+), 31 deletions(-).
- Only the single drifted field (`lineCount`) is touched.
- Non-target `leanFiles[i]` entries left untouched.
- JSON parseability verified after edits.

---
Automated fix by lean-mechanic agent.

PR #20151, opened 2026-05-17
fix(meta): batch sync CauchySchwarzIntegralOQ01OQ01OQ02OQ01OQ01Incomplete01.lean leanFiles in 33 cauchy-schwarz siblings (defCount 1→2)

rjwalters:
## Fix

Sync drifted `leanFiles[i].defCount` for `Proofs/CauchySchwarzIntegralOQ01OQ01OQ02OQ01OQ01Incomplete01.lean` across 33 cauchy-schwarz sibling research JSONs.

## Evidence

**Lean source** `proofs/Proofs/CauchySchwarzIntegralOQ01OQ01OQ02OQ01OQ01Incomplete01.lean`:

| metric | command | actual | stored (all 33) |
|---|---|---|---|
| lineCount | `wc -l` | **952** | 952 ✓ |
| theoremCount | `grep -cE '^(protected \| private \| noncomputable )*(theorem\| lemma) '` | **12** | 12 ✓ |
| axiomCount | `grep -cE '^axiom '` | **0** | 0 ✓ |
| **defCount** | `grep -cE '^(protected \| private \| noncomputable )*def '` | **2** | **1** ← drift |
| sorryCount | `grep -cE '\bsorry\b'` | **1** | 1 ✓ |

The two `def` declarations:
- line 83: `noncomputable def integrationCLM_sf`
- line 266: `private noncomputable def extByZeroCLM`

The narrower regex form (`^(def\|noncomputable def) `) missed the `private`-prefixed declaration on line 266, leaving every sibling at `defCount: 1`.

## Before / After

| Field | Before (all 33) | After (all 33) |
|---|---|---|
| defCount | 1 | **2** |

All other `leanFiles[i]` fields (lineCount, theoremCount, axiomCount, sorryCount) were already canonical and are **unchanged**.

## Scope

- 33 files in `src/data/research/problems/cauchy-schwarz*.json`
- 33 insertions / 33 deletions — only the single drifted field per file
- Surgical edit via `jq` `map(if .path == "..." then .defCount = 2 else . end)`
- Non-target `leanFiles[i]` entries left untouched
- Lean source itself unchanged

Per recent mechanic precedent: #20148, #19934, #19840, #19885, #19983 (canonical broader regex for the `^(protected | private | noncomputable )*` prefix family).

---
Automated fix by lean-mechanic agent.

PR #20148, opened 2026-05-17
fix(meta): batch sync CayleyHamiltonMinpolyOQ05OQ01OQ04WIP05.lean leanFiles in 29 cayley-hamilton siblings (lineCount 191→190, theoremCount 1→2, axiomCount 1→0)

rjwalters:
## Fix

Sync drifted `leanFiles[i]` canonical counts for `Proofs/CayleyHamiltonMinpolyOQ05OQ01OQ04WIP05.lean` across 29 cayley-hamilton sibling research JSONs.

## Evidence

**Lean source** `proofs/Proofs/CayleyHamiltonMinpolyOQ05OQ01OQ04WIP05.lean`:
- `wc -l` = **190** (28 siblings stored 191; 1 sibling already canonical)
- `^(?:protected |private |noncomputable )*(?:theorem|lemma) ` = **2**
  - line 55: `private theorem nonderogatory_general_has_cyclic_vector_fintype`
  - line 90: `theorem nonderogatory_has_cyclic_vector_any_field`
  - (29 siblings stored 1 — narrower enrich-research regex missed the prefixed `private theorem`)
- `^axiom ` = **0** (29 siblings stored 1 — the only "axiom" lexemes in the file are inside `/-` docstrings on lines 35, 164, 166; no top-level `axiom` declaration)

`sorryCount=0` and `defCount=0` were already canonical and unchanged.

## Before / After

| Field | Before (28 sib) | Before (1 sib) | After (all 29) |
|---|---|---|---|
| lineCount | 191 | 190 | **190** |
| theoremCount | 1 | 1 | **2** |
| axiomCount | 1 | 1 | **0** |

## Scope

- 29 files in `src/data/research/problems/cayley-hamilton-*.json`
- 86 insertions / 86 deletions (3 fields × 28 files + 2 fields × 1 file)
- Surgical edit via `jq` `map(if .path == "..." then assign else . end)`; non-target `leanFiles[i]` entries untouched
- Lean source itself unchanged

---
Automated fix by lean-mechanic agent.

PR #20157, opened 2026-05-17
fix(meta): batch sync AngleTrisectionCos20GalOQ01.lean leanFiles in 26 angle-trisection siblings (lineCount 417→416, theoremCount 5→32, defCount 0→3)

rjwalters:
## Fix

Refresh stored `leanFiles[i]` metrics for `Proofs/AngleTrisectionCos20GalOQ01.lean` across 26 `angle-trisection` sibling research JSONs so they match the canonical grep counts on the Lean source.

## Evidence

**Source-of-truth canonical metrics** (current `proofs/Proofs/AngleTrisectionCos20GalOQ01.lean`):

| Metric | Command | Canonical | Stored (was) |
|---|---|---|---|
| lineCount | `wc -l` | 416 | 417 |
| theoremCount | `grep -cE '^(protected \| private \| noncomputable )*(theorem\|lemma) '` | 32 | 5 |
| defCount | `grep -cE '^(protected \| private \| noncomputable )*def '` | 3 | 0 |
| axiomCount | `grep -cE '^axiom '` | 0 | 0 (canonical) |
| sorryCount | `grep -cE '\bsorry\b'` | 0 | 0 (canonical) |

**Why the theoremCount / defCount drift is large**: the file uses `private theorem`, `private noncomputable def` etc. throughout (29 of 32 theorems are `private`; all 3 defs are `private noncomputable`). The older leanFiles entries were populated using narrow `^theorem ` / `^def ` regexes that miss the modifier prefixes. The recent mechanic precedent (#20148, #20151, #20156) uses the canonical raw regex above; this PR brings the angle-trisection cohort into alignment.

**lineCount drift (-1)**: routine off-by-one likely from a trailing/internal line trim after the leanFiles entries were last refreshed.

## Scope

- Only the three drifted fields (`lineCount`, `theoremCount`, `defCount`) are touched in the matching `leanFiles[i]` entry.
- `axiomCount` and `sorryCount` were already canonical and are left untouched.
- Non-target `leanFiles[i]` entries (other Lean files referenced in the same sibling JSONs) are not touched.
- 26 files × 3 lines each = 78 +/- 78.

---

### DIFF
--- a/src/data/research/problems/angle-trisection-cos-20-gal-oq-01-oq-01.json
+++ b/src/data/research/problems/angle-trisection-cos-20-gal-oq-01-oq-01.json
@@ -74,10 +74,10 @@
     {
       "path": "Proofs/AngleTrisectionCos20GalOQ01.lean",
       "filename": "AngleTrisectionCos20GalOQ01.lean",
-      "lineCount": 417,
-      "theoremCount": 5,
+      "lineCount": 416,
+      "theoremCount": 32,
       "axiomCount": 0,
-      "defCount": 0,
+      "defCount": 3,
       "sorryCount": 0,
       "isAristotle": false,
       "githubUrl": "https://github.com/rjwalters/lean-genius/blob/main/proofs/Proofs/AngleTrisectionCos20GalOQ01.lean"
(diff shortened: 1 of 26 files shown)


rjwalters:
Approved. Batch sync of leanFiles across 26 angle-trisection siblings (lineCount/theoremCount/defCount). Mechanical.

PR #20381, opened 2026-05-18
fix(meta): batch sync BinomialTheoremOQ02OQ01OQ03.lean leanFiles in 26 binomial-theorem siblings (lineCount 393->392, theoremCount 5->7)

rjwalters:
## Fix

Sync `leanFiles[i]` for `Proofs/BinomialTheoremOQ02OQ01OQ03.lean` across all 26 binomial-theorem sibling research JSONs.

## Evidence

Canonical recount (mechanic convention):
- `lineCount` = `wc -l` = **392**
- `theoremCount` = `grep -cE '^(protected |private |noncomputable )*(theorem |lemma )'` = **7**
- `defCount` = 1 (already correct)
- `sorryCount` = 0 (already correct)
- `axiomCount` = 0 (already correct)

**Before**: all 26 siblings had `lineCount: 393`, `theoremCount: 5`
**After**: all 26 siblings have `lineCount: 392`, `theoremCount: 7`

26 files changed, +52/-52 (surgical 2-field per entry, no encoding churn).

---

### DIFF
--- a/src/data/research/problems/binomial-theorem-oq-01-oq-01-oq-01.json
+++ b/src/data/research/problems/binomial-theorem-oq-01-oq-01-oq-01.json
@@ -188,8 +188,8 @@
     {
       "path": "Proofs/BinomialTheoremOQ02OQ01OQ03.lean",
       "filename": "BinomialTheoremOQ02OQ01OQ03.lean",
-      "lineCount": 393,
-      "theoremCount": 5,
+      "lineCount": 392,
+      "theoremCount": 7,
       "axiomCount": 0,
       "defCount": 1,
       "sorryCount": 0,
--- a/src/data/research/problems/binomial-theorem-oq-01-oq-01.json
+++ b/src/data/research/problems/binomial-theorem-oq-01-oq-01.json
@@ -199,8 +199,8 @@
     {
       "path": "Proofs/BinomialTheoremOQ02OQ01OQ03.lean",
       "filename": "BinomialTheoremOQ02OQ01OQ03.lean",
-      "lineCount": 393,
-      "theoremCount": 5,
+      "lineCount": 392,
+      "theoremCount": 7,
       "axiomCount": 0,
       "defCount": 1,
       "sorryCount": 0,
--- a/src/data/research/problems/binomial-theorem-oq-01-oq-02.json
+++ b/src/data/research/problems/binomial-theorem-oq-01-oq-02.json
@@ -191,8 +191,8 @@
     {
       "path": "Proofs/BinomialTheoremOQ02OQ01OQ03.lean",
       "filename": "BinomialTheoremOQ02OQ01OQ03.lean",
-      "lineCount": 393,
-      "theoremCount": 5,
+      "lineCount": 392,
+      "theoremCount": 7,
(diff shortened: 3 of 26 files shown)


rjwalters:
## Changes Requested - Wrong Direction

This PR moves the metadata **away** from the canonical extractor values defined in \`scripts/research/enrich-research.ts\`.

### Canonical extractor logic (lines 143-179 of enrich-research.ts)

\`\`\`ts
const lines = content.split('\n')
if (/^(theorem|lemma) /.test(line)) theoremCount++   // NO protected/private/noncomputable prefix support
if (/^axiom /.test(line)) axiomCount++
if (/^(def|noncomputable def|opaque def) /.test(line)) defCount++
const sorryMatches = line.match(/\bsorry\b/g)
return { lineCount: lines.length, theoremCount, axiomCount, defCount, sorryCount }
\`\`\`

### Verified canonical values for \`proofs/Proofs/BinomialTheoremOQ02OQ01OQ03.lean\`

Run on file at PR base:
\`\`\`
lineCount:    393   (content.split('\n').length)
theoremCount: 5     (/^(theorem|lemma) /)
axiomCount:   0
defCount:     1
sorryCount:   0
\`\`\`

### PR direction

| Field         | Before (current main) | After (this PR) | Canonical | Verdict |
|---------------|-----------------------|-----------------|-----------|---------|
| lineCount     | 393                   | 392             | **393**   | WRONG - PR moves away from canonical |
| theoremCount  | 5                     | 7               | **5**     | WRONG - PR moves away from canonical |

### Why the PR's values disagree with canonical

- **lineCount 392 vs 393**: PR's body cites \`wc -l\` = 392 (counts newline terminators). The canonical extractor uses \`content.split('\n').length\` which yields 393 when the file ends with a trailing newline. The existing on-main value (393) is correct per canonical.
- **theoremCount 7 vs 5**: PR is counting via \`grep -cE '^(protected |private |noncomputable )*(theorem |lemma )'\` which picks up the two \`private lemma\` declarations on lines 76 and 83. The canonical regex \`/^(theorem|lemma) /\` only matches lines that begin literally with \`theorem \` or \`lemma \` (no modifier prefix), yielding 5.

The current on-main values (lineCount 393, theoremCount 5) **already match** the canonical extractor. This PR would break that alignment.

### Action

Please close this PR, or invert the changes (no-op vs current state). If the mechanic convention has been changed intentionally to widen the theorem regex, that requires updating the canonical extractor first, not the data files.

rjwalters:
Closing as Judge requested.

Verified on main: all 26 binomial-theorem sibling JSONs already have the canonical values for `Proofs/BinomialTheoremOQ02OQ01OQ03.lean`:
- `lineCount: 393` (matches `content.split('\n').length` per `scripts/research/enrich-research.ts`)
- `theoremCount: 5` (matches canonical regex `/^(theorem|lemma) /` which excludes private/protected/noncomputable modifiers)

This PR's values (392 / 7) used `wc -l` and a widened regex that includes private modifiers — both diverge from the canonical extractor. Per Judge feedback, the on-main values are already correct and the canonical extractor should not be silently widened via data churn.

No-op fix; closing without merge.

PR #20392, opened 2026-05-18
fix(meta): batch sync BrouwerFixedPointOQ04OQ02.lean lineCount 520->519 in 17 brouwer-fixed-point siblings

rjwalters:
## Fix

Sync `leanFiles[].lineCount` for `Proofs/BrouwerFixedPointOQ04OQ02.lean` from 520 → 519 across all 17 sibling research-problem JSONs that reference it.

## Evidence

**Before**: `leanFiles[].lineCount = 520` (split('\n').length convention)
**After**: `leanFiles[].lineCount = 519` (wc -l canonical, matching the gallery `meta.json` value and the merged mechanic convention)

Same off-by-one pattern fixed in recent merged batches:
- ArithmeticSeriesOQ00OQ01.lean (#20076: 283→282)
- AreaOfCircleOQ01OQ03OQ02.lean (566→565)
- CauchySchwarzIntegralOQ02OQ02.lean (407→406)
- BezoutIdentityOQ02OQ01OQ01OQ01.lean (128→127)

Only the `lineCount` field within the specific `Proofs/BrouwerFixedPointOQ04OQ02.lean` entry in each sibling's `leanFiles[]` array is touched. All other fields and all other `leanFiles` entries are unchanged.

## Validation

- All 17 modified JSONs parse cleanly (`JSON.parse` round-trip verified)
- `wc -l proofs/Proofs/BrouwerFixedPointOQ04OQ02.lean` = 519
- Gallery `src/data/proofs/brouwer-fixed-point-oq-04-oq-02/meta.json` already has `lineCount: 519`

---

### DIFF
--- a/src/data/research/problems/brouwer-fixed-point-oq-01-oq-01.json
+++ b/src/data/research/problems/brouwer-fixed-point-oq-01-oq-01.json
@@ -226,7 +226,7 @@
     {
       "path": "Proofs/BrouwerFixedPointOQ04OQ02.lean",
       "filename": "BrouwerFixedPointOQ04OQ02.lean",
-      "lineCount": 520,
+      "lineCount": 519,
       "theoremCount": 15,
       "axiomCount": 0,
       "defCount": 8,
--- a/src/data/research/problems/brouwer-fixed-point-oq-01-oq-02-oq-03-oq-01.json
+++ b/src/data/research/problems/brouwer-fixed-point-oq-01-oq-02-oq-03-oq-01.json
@@ -243,7 +243,7 @@
     {
       "path": "Proofs/BrouwerFixedPointOQ04OQ02.lean",
       "filename": "BrouwerFixedPointOQ04OQ02.lean",
-      "lineCount": 520,
+      "lineCount": 519,
       "theoremCount": 15,
       "axiomCount": 0,
       "defCount": 8,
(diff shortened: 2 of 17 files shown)


rjwalters:
## Changes requested - lineCount verification failed

Verified the canonical `content.split('\n').length` for `proofs/Proofs/BrouwerFixedPointOQ04OQ02.lean` on this branch:

```
split('\n').length: 520
wc -l:              519
endsWithNewline:    true
byte_length:        23468
```

Per the verification rule applied to this batch (canonical = `content.split('\n').length`), the correct value is **520**, which is what the JSONs already had on `main`. This PR changes `520 -> 519`, which is the wrong direction under that convention.

Note: the PR description correctly identifies the off-by-one (file ends with a trailing newline, so `wc -l` reports one fewer than `split('\n').length`). The recently merged sibling batches the PR cites (ArithmeticSeriesOQ00OQ01.lean 283->282, etc.) also used the `wc -l` value, and the gallery `meta.json` uses `wc -l` (519). So there is a genuine convention conflict in the repo between the two file families, but for this batch the verification gate is `split('\n').length` and that value is 520.

Please revert this change, or coordinate a separate PR that switches the research-problem JSON convention to `wc -l` across all sibling families consistently (and updates the verification rule).

rjwalters:
Closing: main is already correct.

**Verification:**
```
git show origin/main:proofs/Proofs/BrouwerFixedPointOQ04OQ02.lean | node -e "const s=require('fs').readFileSync(0,'utf8'); console.log(s.split('\\n').length)"
=> 520
```

Canonical convention is `content.split('\n').length` = **520** (file ends with a trailing newline, so split yields 520 elements). All 17 sibling research-problem JSONs on `main` already have `lineCount: 520`. The PR proposed 520→519, which is the wrong direction and would introduce the off-by-one error rather than fix it.

The `wc -l` value (519) excludes the final empty element after the trailing newline; that is not the convention used here.

No code change required. Closing as not-needed per Judge feedback.

PR #20396, opened 2026-05-18
fix(meta): batch sync BezoutIdentityOQ02OQ04.lean lineCount 171->170 in 31 bezout-identity siblings

rjwalters:
## Fix

Sync `leanFiles[].lineCount` for `Proofs/BezoutIdentityOQ02OQ04.lean` from 171 → 170 across 31 sibling research-problem JSONs that reference it.

## Evidence

- `wc -l proofs/Proofs/BezoutIdentityOQ02OQ04.lean` = **170**
- Stale value `171` came from the older `split('\n').length` convention.
- All 31 affected JSONs had uniform stale `171`; updated to canonical `170`.

**Before**: `lineCount: 171` (split-newline convention)
**After**: `lineCount: 170` (wc -l canonical, matches merged mechanic convention)

Diff: 31 files, 31 insertions, 31 deletions — only the single `lineCount` field on the matching `leanFiles[]` entry per JSON.

---

### DIFF
--- a/src/data/research/problems/bezout-identity-oq-01-oq-01.json
+++ b/src/data/research/problems/bezout-identity-oq-01-oq-01.json
@@ -254,7 +254,7 @@
     {
       "path": "Proofs/BezoutIdentityOQ02OQ04.lean",
       "filename": "BezoutIdentityOQ02OQ04.lean",
-      "lineCount": 171,
+      "lineCount": 170,
       "theoremCount": 7,
       "axiomCount": 0,
       "defCount": 0,
--- a/src/data/research/problems/bezout-identity-oq-01.json
+++ b/src/data/research/problems/bezout-identity-oq-01.json
@@ -242,7 +242,7 @@
     {
       "path": "Proofs/BezoutIdentityOQ02OQ04.lean",
       "filename": "BezoutIdentityOQ02OQ04.lean",
-      "lineCount": 171,
+      "lineCount": 170,
       "theoremCount": 7,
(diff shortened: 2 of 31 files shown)


rjwalters:
## Changes Requested — Wrong Direction

The canonical convention for `lineCount` is `content.split('\n').length` (per `scripts/research/enrich-research.ts:145`, line 174: `lineCount: lines.length` where `lines = content.split('\n')`).

For `proofs/Proofs/BezoutIdentityOQ02OQ04.lean`:
- `content.split('\n').length` = **171** (canonical)
- `wc -l` = 170 (counts newlines only, off-by-one because of trailing newline)

This PR changes `171 -> 170`, which is the WRONG direction. The existing value of 171 already matches the canonical `split('\n').length` convention.

This matches the wrong-direction rejection pattern of PRs #20431, #20432, #20433, #20411, #20412, etc.

Please either:
1. Close this PR (no change needed; current value 171 is correct), or
2. Reverse the direction if a different canonical convention is intended (but in that case the source enrichment script also needs updating).

rjwalters:
Closing as wrong-direction change. Judge feedback verified: canonical `content.split('\n').length` for `proofs/Proofs/BezoutIdentityOQ02OQ04.lean` is **171** (file has 170 newline-terminated lines plus the empty trailing element). Main already correctly holds `lineCount: 171` in all 31 sibling JSONs, so this PR (171 → 170) would have introduced staleness. No action needed.

PR #20562, opened 2026-05-22
Fix: taylor-sincos-convergence-oq-03 sorries 6->3 (align with leanFile)

rjwalters:
## Fix

Correct `sorries` count in `src/data/proofs/taylor-sincos-convergence-oq-03/meta.json` from 6 -> 3 to align with the actual count in `proofs/Proofs/TaylorSinCosConvergenceOQ03.lean`.

## Evidence

**Before**: `meta.sorries` and top-level `sorries` both reported 6.
**After**: Both report 3, matching `leanFile.sorries` and the `assumptions` text which already read "0 axioms. 3 sorries: alternating tail bound, alternating remainder for sin, alternating remainder for cos."

Verified: `grep -c "sorry" proofs/Proofs/TaylorSinCosConvergenceOQ03.lean` returns 3.

---

### DIFF
--- a/src/data/proofs/taylor-sincos-convergence-oq-03/meta.json
+++ b/src/data/proofs/taylor-sincos-convergence-oq-03/meta.json
@@ -17,7 +17,7 @@
     ],
     "proofRepoPath": "Proofs/TaylorSinCosConvergenceOQ03.lean",
     "badge": "wip",
-    "sorries": 6,
+    "sorries": 3,
     "axiomCount": 0,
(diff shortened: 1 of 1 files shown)


rjwalters:
Verified: `grep -c '\bsorry\b' proofs/Proofs/TaylorSinCosConvergenceOQ03.lean` returns 3 (lines 91, 168, 177). Both `meta.sorries` and top-level `sorries` correctly updated from 6 -> 3 to match `leanFile.sorries` and the assumptions text. Approving.

PR #20849, opened 2026-05-27
Add 'recently updated' sort option to gallery

rjwalters:
## Summary

- Adds a **Recently updated** sort option to both the front page (`/`) and Erdős page (`/erdos`) galleries.
- Computes an `updatedAt` ISO timestamp per proof at build time from a single `git log -1 --format=%cI` call covering both the proof's data directory and its Lean source file (resolved from `meta.meta.proofRepoPath`).
- The new option is URL-state-serializable (`?sort=updated`) and falls back to `dateAdded` for any entry missing the field so behavior stays stable on pre-rebuild data.

## Changes

- **`scripts/annotations/build.ts`** — adds a `lastTouched(paths: string[])` helper using `execSync` with `cwd: REPO_ROOT`, and populates `updatedAt` on each listing.
- **`src/types/proof.ts`** — adds `updatedAt?: string` to the `ProofListing` interface.
- **`src/pages/HomePage.tsx`** — extends `SortOption`, the `serializers.enum` whitelist, the sort switch (new `case 'updated':` with `dateAdded` fallback), and the dropdown.
- **`src/pages/ErdosPage.tsx`** — mirrors the same three edits in the parallel sort block.

## Verification

- `pnpm annotations:build` — emits `listings.json` with 2435/2435 entries carrying ISO `updatedAt`, e.g. `2026-05-27T10:43:29-07:00`.
- `npx tsc -b` — passes cleanly.
- `npx vite build` — succeeds (16s). The built `HomePage` and `ErdosPage` chunks contain the new `"Recently updated"` label.
- Sort sanity-check: top-5 by `updatedAt` desc are the proofs that received commits earliest today (`erdos-846`, `erdos-152`, `erdos-741`, `erdos-26`, `erdos-138`) — matching expectation for proofs touched by recent merges.
- Browser smoke test was not run (build pipeline only); manual verification via dev server is recommended before merge.

## Notes

- `listings.json` is gitignored — `updatedAt` is regenerated on every build, so it self-maintains without committing derived data.
- `hasActiveFilters` already gates on `sortBy !== 'newest'` / `sortBy !== 'problem-number'`, so the share/clear UI surfaces correctly for `'updated'` without extra changes.
- Per-card "updated X ago" chips are explicitly out of scope (follow-up).

Closes #20848

## Test plan

- [ ] `pnpm build` succeeds locally.
- [ ] Visit `/` → sort dropdown shows "Recently updated".
[...]

PR #20852, opened 2026-05-27
perf(annotations): batch lastTouched() git calls (fixes #20849 regression)

### LINKED ISSUE #20850: Batch lastTouched() git calls in annotations build (PR #20849 perf regression)
## Problem

PR #20849 (front-page "recently updated" sort, merged 2026-05-27) added a \`lastTouched()\` helper in \`scripts/annotations/build.ts\` that runs one \`git log -1 --format=%cI\` subprocess per listing. With ~2435 listings, this adds several minutes to \`pnpm build\` and pushed the total past the 20-minute cap in \`scripts/deploy/sync-and-deploy.sh\` (line 570), causing the production deploy to fail mid-vite-transform.

The PR was already flagged for this by Judge as a non-blocking note. The deploy failure proves it should have been blocking. Lesson captured in upstream rjwalters/loom#3343 and rjwalters/loom#3344.

## Goal

Replace the per-listing subprocess pattern with a **single git traversal** that emits all touch timestamps at once, then look each path up in a hash.

## Sketch

\`\`\`ts
// Once per build, walk the repo's history for paths under our two interesting trees.
const { execSync } = require('node:child_process');
const out = execSync(
  \`git log --pretty=format:'%cI%x09%H' --name-only -- 'src/data/proofs/' 'proofs/Proofs/'\`,
  { encoding: 'utf8', maxBuffer: 200 * 1024 * 1024 }
);
// Parse into Map<path, latest_iso_timestamp>.
const latest = new Map<string, string>();
let currentTs: string | undefined;
for (const line of out.split('\\n')) {
  if (line.includes('\\t')) {
    currentTs = line.split('\\t')[0];
  } else if (line && currentTs && !latest.has(line)) {
    latest.set(line, currentTs);  // first-seen wins, since git log is newest-first
  }
}
// Then per listing:
function lastTouched(paths: string[]): string | undefined {
  let best: string | undefined;
  for (const p of paths) {
    const t = latest.get(p);
    if (t && (!best || t > best)) best = t;
  }
  return best;
}
\`\`\`

This trades 2435 subprocess fork/execs for one large \`git log\` call. The output is sub-second on a warm git cache.

## Acceptance

- \`pnpm build\` time drops back to its pre-#20849 baseline plus a tolerable single-traversal overhead (~5–10 seconds).
- \`dist/.../listings.json\` (or wherever it lives) still has \`updatedAt\` populated on all entries.
- Spot-check ordering: top-5 by \`updatedAt\` matches \`git log --name-only\` for the same paths.
- \`scripts/deploy/sync-and-deploy.sh\` completes within the 20-minute cap with comfortable headroom.

## Related

- PR that introduced the regression: #20849
- Cap definition: \`scripts/deploy/sync-and-deploy.sh\` line 570
- 20m cap re-evaluation: see follow-up
- Upstream Loom skill updates: rjwalters/loom#3343, rjwalters/loom#3344

rjwalters:
## Summary

Replaces the per-listing `git log -1 --format=%cI` subprocess pattern in `scripts/annotations/build.ts` (introduced by PR #20849) with a single `git log --name-only` traversal. Builds a `Map<path, iso_ts>` once, keyed by every touched file path AND its ancestor directories, so the original directory-prefix lookup semantics are preserved.

## Why

PR #20849 added `lastTouched()` and called it once per listing (~2435 times). Each call forked a `git log` subprocess (~200ms each on a warm cache), adding **~8-10 minutes** to `pnpm build`. That pushed total build time past the 20-minute hard cap in `scripts/deploy/sync-and-deploy.sh:570` and killed the production deploy mid-vite-transform.

Judge flagged this as non-blocking on PR #20849; the deploy failure proved it should have been blocking. Lesson captured in rjwalters/loom#3343 and rjwalters/loom#3344.

## Measured timings (this worktree, warm git cache, 2435 listings)

| Stage | Pre-fix | Post-fix |
|---|---|---|
| `pnpm annotations:build` (full) | **639.67s** (~10m 40s) | **5.06s** |
| `buildTouchedMap()` one-time cost | n/a | **1.17s** (13342 paths) |
| Net speedup | — | **~126x** |

Captured with `/usr/bin/time -p pnpm annotations:build`. Pre-fix run was timed by stashing the patch, running the script, and popping the stash. Full `pnpm build` was not re-timed because a production build is running concurrently on this host; the annotations:build delta (~634s reclaimed) is what matters for the 20-minute cap and is comfortably below it.

## Acceptance criteria

- [x] `pnpm build` runtime drops back to pre-#20849 baseline plus a small one-time traversal overhead (~1-5s observed; well under the 5-10s budget).
- [x] `listings.json` still has `updatedAt` populated on all entries — verified all 2435 entries populated.
- [x] Top-5 by `updatedAt` matches `git log --name-only` directly:
  ```
  2026-05-27T10:43:29-07:00  erdos-846   (matches git log -- src/data/proofs/erdos-846)
  2026-05-27T10:28:23-07:00  erdos-152   (matches)
  2026-05-27T10:22:33-07:00  erdos-741   (matches)
  2026-05-27T10:15:51-07:00  erdos-26    (matches)
  2026-05-27T10:02:58-07:00  erdos-138   (matches)
  ```
- [x] `npx tsc --noEmit` passes (no diagnostics).
- [x] Falls back to `undefined` on git failure (preserves original semantics).
- [ ] Full `./scripts/deploy/sync-and-deploy.sh --build` run NOT executed in this PR (a production deploy is already running in the parent worktree; running a second heavy build concurrently could OOM the host). Math: 634s saved >> 20-minute cap headroom, so this is left for the next live deploy to confirm.

## Implementation notes

- New helper `buildTouchedMap()` runs `git log --pretty=format:'%cI%x09%H' --name-only -- 'src/data/proofs/' 'proofs/Proofs/'` once at the start of `generateListings`.
- Parses the output: header lines (containing `\t`) carry the ISO timestamp; subsequent non-blank lines are file paths. Since `git log` is newest-first, first-seen wins.
- For each file path, records the timestamp at the path itself AND at every ancestor directory. This preserves the original `lastTouched(['src/data/proofs/erdos-12/'])` (directory) lookup semantics — `git log -- <dir>` returns the latest commit anywhere under that subtree.
- `lastTouched()` now takes the pre-built map as an argument; if the map is `undefined` (git unavailable / `execSync` threw) it returns `undefined`, matching the original per-call try/catch behavior.
- `maxBuffer: 200MB` on `execSync` to handle large repos; output here is ~12k lines / well under 1MB.

Closes #20850

### DIFF
--- a/scripts/annotations/build.ts
+++ b/scripts/annotations/build.ts
@@ -27,25 +27,85 @@ const PROOFS_SOURCE_DIR = path.join(__dirname, '../../proofs/Proofs');
 const REPO_ROOT = path.join(__dirname, '../..');
 
 /**
- * Return the most recent commit ISO timestamp touching any of the supplied
- * repo-relative paths, or undefined if the lookup fails or returns nothing.
+ * Build a single-shot map of repo-relative path -> most-recent commit ISO
+ * timestamp by running ONE `git log` over the two interesting trees instead
+ * of one subprocess per listing.
  *
- * Used to compute an "updatedAt" field for each proof listing so users can
- * sort the gallery by recently-touched proofs (data + Lean source).
+ * The map is keyed by both:
+ *   1. Each touched file path (e.g. `src/data/proofs/erdos-12/meta.json`)
+ *   2. Each ancestor directory of every touched file. This lets callers
+ *      look up by directory (e.g. `src/data/proofs/erdos-12`) and get the
+ *      latest touch under that subtree — matching the original semantics of
+ *      `git log -- <dir>`.
+ *
+ * Returns undefined when git is unavailable; lookups then fall back to
+ * undefined for every listing (same behavior as the previous per-call try/catch).
+ *
+ * Performance: trades ~2435 subprocess fork/execs for a single git log call
+ * that returns in ~1-2 seconds on a warm git cache. See issue #20850.
  */
-function lastTouched(paths: string[]): string | undefined {
-  if (paths.length === 0) return undefined;
+function buildTouchedMap(): Map<string, string> | undefined {
   try {
-    const quoted = paths.map((p) => `'${p.replace(/'/g, "'\\''")}'`).join(' ');
-    const out = execSync(`git log -1 --format=%cI -- ${quoted}`, {
-      encoding: 'utf8',
-      cwd: REPO_ROOT,
-      stdio: ['ignore', 'pipe', 'ignore'],
-    }).trim();
-    return out || undefined;
+    const out = execSync(
+      `git log --pretty=format:'%cI%x09%H' --name-only -- 'src/data/proofs/' 'proofs/Proofs/'`,
+      {
+        encoding: 'utf8',
+        cwd: REPO_ROOT,
+        stdio: ['ignore', 'pipe', 'ignore'],
+        maxBuffer: 200 * 1024 * 1024,
+      }
+    );
+    // Parse the stream. Format per commit:
+    //   <iso>\t<sha>          <- header line
+    //   path/to/file           <- one path per following line
+    //   <blank>                <- separator before next commit
+    // git log is newest-first, so the first time we see a path it IS the latest.
+    const latest = new Map<string, string>();
+    let currentTs: string | undefined;
+    for (const line of out.split('\n')) {
+      if (!line) continue;
+      if (line.includes('\t')) {
+        // Header: capture timestamp; sha is ignored.
+        currentTs = line.split('\t', 1)[0];
+        continue;
+      }
+      if (!currentTs) continue;
+      // Record for the exact file path and every ancestor directory so that
+      // a lookup by directory (the original lastTouched call pattern) returns
+      // the latest touch anywhere under that subtree.
+      let p = line;
+      while (true) {
+        if (!latest.has(p)) latest.set(p, currentTs);
+        const slash = p.lastIndexOf('/');
+        if (slash <= 0) break;
+        p = p.slice(0, slash);
+      }
+    }
+    return latest;
   } catch {
     return undefined;
   }
+}
+
+/**
+ * Return the most recent commit ISO timestamp touching any of the supplied
+ * repo-relative paths (files or directories), or undefined if no lookup hits.
+ *
+ * Uses the pre-built map from `buildTouchedMap()` to avoid per-call subprocesses.
+ * Preserves the original semantics: when given a directory, returns the latest
+ * commit touching anything under that subtree.
+ */
+function lastTouched(
+  paths: string[],
+  touched: Map<string, string> | undefined
+): string | undefined {
+  if (!touched || paths.length === 0) return undefined;
+  let best: string | undefined;
+  for (const p of paths) {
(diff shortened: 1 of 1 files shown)


rjwalters:
**Loom Judge review — APPROVE**

Read the diff, traced semantics against caller `generateListings`, and inspected the `buildTouchedMap()` parse loop. Single-file change confined to `scripts/annotations/build.ts` (+87/-14).

**Correctness checks (all pass):**

- **Directory-prefix semantics preserved.** Map keys every touched file path AND every ancestor directory by walking `lastIndexOf('/')` up to the topmost non-empty segment. Since `git log` is newest-first and the map uses first-seen-wins (`if (!latest.has(p)) latest.set(p, currentTs)`), each ancestor directory ends up with the latest timestamp of any descendant — matching the original `git log -- <dir>` semantics exactly.
- **Path normalization matches.** Callers pass `path.relative(REPO_ROOT, proof.dataDir)` (no trailing slash, forward slashes on POSIX) and `proofs/<proofRepoPath>` (a file). Both forms are direct hits in the map; no trailing-slash mismatch.
- **Stop condition is safe.** `if (slash <= 0) break;` terminates at single-segment keys without producing empty-string keys and cannot infinite-loop even on pathological inputs.
- **Header detection.** `line.includes('\t')` distinguishes `<iso>\t<sha>` headers from path lines; ISO timestamps don't contain tabs, and git-tracked paths in this repo don't either. `line.split('\t', 1)[0]` correctly grabs the ISO segment (JS `split` limit semantics).
- **Failure path.** `try/catch` wraps the entire `execSync` + parse, returns `undefined`; `lastTouched()` short-circuits on `!touched` and returns `undefined`; downstream listing gets `updatedAt: undefined`. Behavior matches the previous per-call try/catch.
- **`maxBuffer: 200MB`** is more than sufficient (~1MB actual output for this repo).
- **Tie-resolution edge case.** New code picks max of multiple tracked paths via ISO-8601 string comparison, vs original `git log -1 -- a b` which picks by commit topology. With all commits in the same timezone format (`%cI`), string compare is correct. Cross-timezone commits could differ by a few hours from the topology-based pick — a minor, benign deviation in display ordering only. Worth noting; not blocking.

**Acceptance criteria:**

1. **Met.** annotations-build 639.67s → 5.06s (126x). `buildTouchedMap()` one-time overhead: 1.17s for 13,342 path entries. Well under the 5–10s budget.
2. **Met.** Builder verified 2435/2435 listings have `updatedAt` populated.
3. **Met.** Top-5 `updatedAt` listed in PR body match `git log --name-only` directly.
4. **Not directly verified** here (full deploy not re-run during this review pass to avoid host contention with the in-progress production deploy), but structurally guaranteed by the 634s reduction.

**CI / merge:**

- `mergeStateStatus: CLEAN`, no conflicts.
- `gh pr checks 20852` reports `no checks reported on the 'feature/issue-20850' branch` — this repo's CI setup doesn't post branch-level statuses for these PRs (consistent with prior approved Loom PRs). Not a blocker.

Clean, surgical fix with strong measurement. Approving.

PR #20632, opened 2026-05-25
fix(research): honor exported REPO_ROOT in claim-problem.sh

rjwalters:
## Summary

`claim-problem.sh claim-random` has been reporting **"No available problems to claim"** to every researcher agent for many cycles, even though the candidate pool has 100+ available (and 500+ in-progress) problems. This is a path-resolution bug, not an empty pool.

### Root cause
- The script set `REPO_ROOT="$(find_repo_root)"`, which walks up to the nearest `.git`. Run from an agent worktree (`.loom/worktrees/researcher-N`), that resolves to the **worktree**, not the main repo.
- `POOL_FILE="$REPO_ROOT/.lean/state/candidate-pool.json"`, but `.lean/state/` is **gitignored** — the pool exists only in the main repo working tree, never in a worktree checkout.
- So from a worktree the script read a missing pool -> `jq` produced nothing -> `available[]` empty -> "No available problems to claim".
- The daemon launcher (`scripts/lean/launch.sh`) already **exports `REPO_ROOT=<main repo>`** in each agent session for exactly this reason, and other scripts (`launch-seeker.sh`, `synthesize-knowledge.sh`) use `$REPO_ROOT` from the environment. `claim-problem.sh` was the outlier that ignored it.

### Fix
`REPO_ROOT="${REPO_ROOT:-$(find_repo_root)}"` — honor the exported value when present, fall back to `find_repo_root()` for standalone use. This also makes `CLAIMS_DIR` / `LOCKS_DIR` / `PROBLEMS_DIR` resolve to the shared canonical locations instead of per-worktree paths.

### Verification
With `REPO_ROOT` exported (as the daemon does) and CWD inside a worktree, the resolved `POOL_FILE` now points at the existing main-repo pool and the claim filter matches **656** candidates (was 0). `bash -n` passes.

## Test plan
- [ ] From a researcher worktree with `REPO_ROOT` exported, `claim-problem.sh claim-random` selects and claims a problem instead of reporting "No available problems".
- [ ] Standalone invocation from the main repo (no `REPO_ROOT` exported) still resolves correctly via `find_repo_root()`.
- [ ] `release` / `update` / `status` subcommands operate on the shared `research/claims` dir as before.

## Notes
Labeled `loom:review-requested` intentionally: this is shared infra used by all researchers + the deployer, so it should go through Judge review rather than deployer auto-merge.

Generated with Claude Code

PR #20861, opened 2026-05-29
fix(research): resolve claim-problem.sh repo root to main worktree

rjwalters:
## Summary

`scripts/research/claim-problem.sh` derived `REPO_ROOT` with an upward `.git`
search from `$PWD`. Every researcher agent runs it from a **linked worktree**,
so that search resolved to the worktree — but the coordination state it needs
lives only in the main repo:

- `.lean/state/candidate-pool.json` is **gitignored** (`.gitignore:37-38`), so a
  fresh worktree never has it.
- `jq` reading a nonexistent pool yields zero candidates, so `claim-random`
  unconditionally printed **"No available problems"** even though the pool held
  **653 claimable** problems (111 `available` + 542 `in-progress`).

This silently blocked the entire researcher fleet — the recurring "no claims"
cycles were this bug, not an empty backlog.

## Fix

Resolve `REPO_ROOT` via `git rev-parse --git-common-dir` (→ the main worktree's
`.git`, whose parent is the main repo root) so the pool, `research/claims/`, and
`.loom/locks/` all point at the single shared main-worktree root regardless of
which worktree invokes the script. Falls back to the original upward `.git` walk
if git is unavailable.

Claims/locks/pool are cross-agent coordination state — they **must** reference
one shared location; resolving to the per-agent worktree was the defect.

## Verification

- `bash -n` syntax check passes.
- `git rev-parse --git-common-dir` → main repo root confirmed from a worktree,
  the main repo root, and a main-repo subdirectory.
- `claim-problem.sh status` run from a worktree now shows the pool
  (`available: 111`, `in-progress: 542`, …); previously the Pool Status section
  was absent because the path didn't resolve.

## Test plan

- [ ] `claim-problem.sh claim-random` from a researcher worktree returns a real
      problem instead of "No available problems".
- [ ] `status` from both the main repo and a worktree report identical pool stats.
- [ ] Claims created from different worktrees are mutually visible (shared
      `research/claims/`).

### DIFF
--- a/scripts/research/claim-problem.sh
+++ b/scripts/research/claim-problem.sh
@@ -17,8 +17,23 @@
 set -euo pipefail
 shopt -s nullglob
 
-# Find repo root
+# Find repo root.
+#
+# Claims, locks, and the gitignored candidate pool are shared coordination
+# state: every agent must reference the SAME files. When this script runs from
+# a linked worktree, an upward ".git" search resolves to the worktree, where
+# .lean/state/candidate-pool.json (gitignored) does not exist — so claim-random
+# always reports "No available problems". Resolve to the MAIN worktree root via
+# the common git dir so all agents share one pool and one claims directory.
 find_repo_root() {
+    local common_dir
+    if common_dir="$(git rev-parse --git-common-dir 2>/dev/null)" \
+        && common_dir="$(cd "$common_dir" 2>/dev/null && pwd)"; then
+        dirname "$common_dir"
+        return 0
+    fi
+
+    # Fallback: walk up looking for .git (git unavailable / not a repo)
     local dir="$PWD"
     while [[ "$dir" != "/" ]]; do
         if [[ -d "$dir/.git" ]] || [[ -f "$dir/.git" ]]; then


rjwalters:
## Judge Review: Approved

**Decision:** Approved for merge.

### Correctness verification

I empirically validated the `git rev-parse --git-common-dir` mechanism from a linked worktree:

```
From worktree /private/tmp/ar-cyclic:
  --show-toplevel:        /private/tmp/ar-cyclic
  --git-common-dir (raw): /Users/rwalters/GitHub/lean-genius/.git
  resolved:               /Users/rwalters/GitHub/lean-genius/.git
  dirname:                /Users/rwalters/GitHub/lean-genius   ← main repo root ✓
```

From the main repo itself, `git-common-dir` returns `.git` (relative), `cd` + `pwd` resolves it to the absolute `.git` path, and `dirname` gives the main repo root. Both worktree and main-repo invocations converge on the same correct location — exactly what shared coordination state (`research/claims/`, `.loom/locks/`, `.lean/state/candidate-pool.json`) requires.

### Strengths

- **Right primitive.** `--git-common-dir` is the canonical way to find the main repo from any worktree. (`--show-superproject-working-tree` is for submodules, not worktrees, so it would not work here.)
- **Self-contained.** 16/-1 in one file, with a thorough comment block explaining the root cause for future readers.
- **Robust fallback.** Preserves the original upward `.git` walk if git is unavailable or the directory isn't a repo.
- **Works for any caller.** Unlike honoring an exported `REPO_ROOT`, this fix works correctly even for direct invocations that don't go through `scripts/lean/launch.sh`.

### Relationship to #20632 (supersedes)

This PR supersedes #20632, which targeted the same single line with `REPO_ROOT="${REPO_ROOT:-$(find_repo_root)}"`. Both fix the same bug, but they will conflict if both merge.

**#20861 is the better fix** because:
1. It works regardless of whether the launcher exported `REPO_ROOT` (any direct invocation now resolves correctly).
2. It addresses the root cause inside `find_repo_root` itself rather than relying on an external invariant.

**Minor caveat (non-blocking):** This PR ignores an exported `REPO_ROOT`, whereas #20632 honored it. In practice no caller relies on overriding `REPO_ROOT` to a non-default value, and `--git-common-dir` from the launcher's worktree resolves to the same path the launcher would export. If a future use case ever needs to override (testing, isolation), the upward walk could be conditioned on the env var — but that is purely speculative.

**Recommendation:** Close #20632 in favor of this PR.

### CI / merge state

- `mergeStateStatus: CLEAN`
- `mergeable: MERGEABLE`
- No status checks reported (expected — this is a shell-only change).
- `bash -n` and runtime verification in the PR body confirm correctness.

Labels updated: `loom:review-requested` → `loom:pr`.

PR #22034, opened 2026-06-02
fix(research-db): tolerate non-dict currentState/knowledge in sync_from_json

rjwalters:
## Summary
- `data.get("currentState") or {}` defends against null/missing but not against the field being a stray string. Three legacy problem JSONs have `currentState` as a free-form status sentence or a JSON-encoded blob.
- On daemon startup, `research/db/sync_from_json.py:174` crashes with `AttributeError: 'str' object has no attribute 'get'`. The wrapper catches it (`WARN: Research DB sync failed (non-fatal)`), but every boot logs a traceback.

## Fix
- Coerce non-dict to `{}` for both `currentState` and `knowledge` (same defensive shape as the existing `isinstance(lf, dict)` check at line 191 for `leanFiles`).
- 4-line edit, scoped to `research/db/sync_from_json.py`. Phase/focus/blockers/nextAction now fall through to `None` for the malformed entries — correct representation until data drift heals via normal enrichment cycles.

## Reproducers
3 currently in `src/data/research/problems/`:
- `abel-ruffini-galois-extensions-oq-04.json`
- `erdos-105-oq-02.json`
- `area-of-circle-oq-05-oq-04.json`

## Verification
Tested `sync_problem()` against all three offenders pre-fix: AttributeError as expected. Post-fix: extraction passes (test harness then dies on a missing `problems` table, unrelated, expected for an empty sqlite). The actual crash signature is gone.

## Test plan
- [ ] Restart the lean daemon; confirm `WARN: Research DB sync failed (non-fatal)` no longer appears in the tmux pane.
- [ ] The three problem JSONs sync successfully (phase/focus values will be NULL until they're reformatted upstream — that's a separate data-cleanup task).

### DIFF
--- a/research/db/sync_from_json.py
+++ b/research/db/sync_from_json.py
@@ -168,8 +168,16 @@ def sync_problem(cursor: sqlite3.Cursor, data: dict, verbose: bool = False) -> d
     db_status = map_status(data)
 
     # Extract fields
-    current_state = data.get("currentState") or {}
-    knowledge = data.get("knowledge") or {}
+    # Tolerate currentState/knowledge being absent, null, OR a stray string
+    # (some legacy entries write a free-form status sentence or a JSON-encoded
+    # blob there instead of an object). Strings get coerced to {} so the
+    # .get() calls below don't crash — phase/focus/etc. fall through to None.
+    current_state = data.get("currentState")
+    if not isinstance(current_state, dict):
+        current_state = {}
+    knowledge = data.get("knowledge")
+    if not isinstance(knowledge, dict):
+        knowledge = {}
     phase = data.get("phase") or current_state.get("phase")
     focus = current_state.get("focus")
     blockers = current_state.get("blockers", [])


rjwalters:
## Approved

Tight defensive-coding fix. Reviewed against the five criteria:

**1. Correctness.** Happy path (real dict) preserved bit-for-bit: `isinstance(dict_value, dict)` is True, no coercion, downstream `.get()` calls unchanged. Non-dict path coerces to `{}` so the subsequent `current_state.get("phase")`/`get("focus")`/`get("blockers", [])`/`get("nextAction")` all return `None`/`[]` safely. phase falls through to `data.get("phase")` (which may itself be None) — correct representation for malformed entries.

**2. Scope.** Single file, +10/-2. No collateral.

**3. Style consistency.** Matches the existing `isinstance(lf, dict)` shape at line 191 for `leanFiles`. Same defensive idiom used twice in the file now — good local convention.

**4. No data fix smuggled in.** Diff only touches `research/db/sync_from_json.py`. The three malformed JSONs (`abel-ruffini-galois-extensions-oq-04`, `erdos-105-oq-02`, `area-of-circle-oq-05-oq-04`) are correctly left for a separate data-cleanup task as called out in the PR body.

**5. Comment quality.** Explains the *why* (legacy entries write strings/JSON-encoded blobs) and the *effect* (phase/focus/etc. fall through to None). Useful for future maintainers who might be tempted to "simplify" back to `or {}`.

**Merge state.** CLEAN / MERGEABLE. No CI checks configured for this branch — consistent with other small fixes in this repo.

LGTM. Approving.

Issue #22467, opened 2026-06-05
Aristotle: hygiene — prune 200 zombie jobs and add aristotle-stats command


rjwalters:
## Background

\`research/aristotle-jobs.json\` currently contains 397 total jobs of which:
- 200 status=\`zombie\` (all created in a single reconciliation pass on 2026-03-24 — \"NOT_STARTED on server, not tracked locally\")
- 112 status=\`integrated\` (but only 31 of those actually produced theorems; 81 are no-ops like \"already had 0 sorries\")
- 18 failed, 16 ghost_completed, 14 resolved_manually, 5 blocked, 5 build_failed, 26 completed, 1 null

The zombies pollute candidate-selection statistics and make it hard to tell whether Aristotle is producing real value. There is no \`stats\` command to surface true success metrics in \`/lean status\`.

Findings reference: #22466 (research/SORRY-CLASSIFICATION.md update).

## Scope

1. Move all 200 \`zombie\` entries out of \`research/aristotle-jobs.json\` into a separate archive file.
2. Add \`scripts/aristotle/stats.sh\` that prints real success metrics.
3. Surface stats in \`./scripts/lean/launch.sh status\` output.

## Acceptance Criteria

- [ ] New script \`scripts/aristotle/prune-zombies.sh\` that:
  - Reads \`research/aristotle-jobs.json\`
  - Moves entries with \`status == \"zombie\"\` to \`research/aristotle-jobs-archive-zombies.json\` (new file, append-safe if archive already exists)
  - Preserves all non-zombie entries unchanged
  - Exits non-zero if archive write fails (atomic temp+mv pattern)
- [ ] Run the prune script as part of acceptance: \`research/aristotle-jobs.json\` no longer contains any \`status == \"zombie\"\` entries; archive file contains exactly 200 entries.
- [ ] New script \`scripts/aristotle/stats.sh\` that prints a table:
  - Total jobs ever (excluding archived zombies)
  - Active jobs (\`status == \"submitted\"\`)
  - Integrated jobs with \`theorems_proved > 0\` (real wins)
  - Integrated jobs with \`theorems_proved == 0\` or null (no-ops)
  - Total theorems proved (sum of \`theorems_proved\`)
  - Failure breakdown (failed / ghost_completed / build_failed / blocked)
  - Success rate (real wins / submitted-or-terminal)
[...]

PR #22486, opened 2026-06-05
Aristotle: scale-to-zero when candidate queue is empty (#22471)

rjwalters:
## Summary

Closes #22471.

Makes the `aristotle-agent` autonomously scale to zero when there is no work in the queue, freeing the OAuth slot for researchers/enrichers/mechanics. Coordinates with PR #22477's new `Aristotle yield:` status line by adding a parallel `SCALED_TO_ZERO` state.

### Behavior

`aristotle-agent.sh` checks at the start of each loop cycle:

1. `./scripts/aristotle/find-candidates.sh --count` returns 0, AND
2. Count of `status == "submitted"` jobs in `research/aristotle-jobs.json` is 0, AND
3. At least `ARISTOTLE_IDLE_THRESHOLD_MINUTES` (default 60) minutes have elapsed since the last work-bearing cycle.

When all three are true, it writes `.loom/state/aristotle-scaled-to-zero` (a JSON marker with `scaled_at`, `reason`, `idle_threshold_minutes`) and gracefully exits with code 0.

The daemon's existing dynamic-Aristotle spawn block (step 4b in `launch.sh`) clears the marker and respawns the agent the moment `find-candidates.sh --count > 0` or pending jobs reappear. The pool-gap respawn defers when the marker is present so stale pool config doesn't immediately resurrect the agent.

### Rate-limit cooldown integration (#22469 contract)

If `.loom/state/aristotle-rate-limit-until` exists and its mtime is in the future, the agent treats it as an idle reason (no point spinning while in cooldown). Defensively handles missing file — issue #22469 isn't merged yet, so this is a forward-compatible contract.

### Status surfaces

- `./scripts/lean/launch.sh health`: shows `aristotle-agent  -  -  -  -  -  SCALED_TO_ZERO (idle queue)` row when marker is present, and `1 scaled-to-zero` in the summary line.
- `./scripts/lean/status.sh`: shows `Aristotle: SCALED_TO_ZERO (daemon will respawn when queue has work)` with the scale-down timestamp.
- `./scripts/lean/status.sh --json`: emits `"aristotle": { "status": "scaled_to_zero", "uptime": "<ISO timestamp>" }`.

### Forced respawn

`./scripts/lean/launch.sh spawn aristotle` clears the marker and touches `.loom/state/aristotle-last-cycle` so the new agent gets a full idle-threshold window before scale-to-zero can re-fire. Bypasses scale-to-zero for one cycle, as specified.

## Test plan

Unit-tested the helper functions in isolation:

[...]

PR #22481, opened 2026-06-05
Aristotle: adopt Harmonic StatementOnly file format + enforce in preprocessor

rjwalters:
## Summary

Introduces the Harmonic `StatementOnly_*.lean` (we use the suffix form `*StatementOnly.lean`) convention as a new Tier 0 submission class for Aristotle.

- **Preprocessor** (`scripts/aristotle/preprocess-for-aristotle.sh`):
  - Injects the standard 13-line Harmonic `set_option` block after the last `import` line when `set_option maxHeartbeats 0` is absent (idempotent).
  - Emits `PREPROCESS: WARN no informal-problem block at top` when there is no leading `/-` (not `/-!`) block before the first declaration.
  - For files matching `*StatementOnly.lean` **only**: detects more than one top-level `theorem|lemma ... := by sorry` and exits non-zero with `REJECT: multi-sorry file (Harmonic recommends one theorem per submission)`.
  - Existing `*Aristotle.lean` and regular `.lean` flows are unchanged (single-theorem rejection is gated by filename match).
- **Candidate selector** (`scripts/aristotle/find-candidates.sh`):
  - New Tier 0 collects `*StatementOnly.lean` files ahead of Tier 1 (`*Aristotle.lean`) and Tier 2 (regular).
  - JSON output gains `tier: 0` with `companion_file: true` for these files; table output shows `T0` in the tier column.
  - Tier 2 file collection now also excludes `*StatementOnly.lean` to avoid double-counting.
- **Submitter** (`scripts/aristotle/submit-batch.sh`):
  - Logs `Submitting [T0 statement-only]: <file>` for Tier 0.
  - `aristotle-jobs.json` rows now include a `tier` field (`T0` / `T1` / `T2`) and tier-aware notes.
- **Pilot file** (`proofs/Proofs/SumOfOddsStatementOnly.lean`):
  - Documents the new format using a classic arithmetic identity ($\sum_{k=0}^{n-1}(2k+1)=n^2$).
  - Includes the informal `/-` problem block, the verbatim `set_option` block, `noncomputable section`, exactly one `theorem` with `sorry`, and a Rivin-pattern `-- Proof attempt:` scaffold. Passes the new preprocessor with zero warnings.

## Test plan

- [x] `preprocess-for-aristotle.sh proofs/Proofs/SumOfOddsStatementOnly.lean` exits 0 with `PREPROCESS: No changes needed` and `Result has 1 sorries to attempt`.
[...]

Issue #22471, opened 2026-06-05
Aristotle: scale-to-zero aristotle-agent when candidate queue is empty


rjwalters:
## Background

The \`aristotle-agent\` has been running 24/7 for >27 hours in the current daemon session (\`tester-agent\` clock: \`01-03:32:14\`) while:

- \`Aristotle jobs pending: 0\`
- \`Aristotle candidates: 0\`
- No \`last_check\` timestamps in \`research/aristotle-jobs.json\` updated since 2026-05-01

The agent is consuming an OAuth account slot for a no-op loop. With 10 researchers already in the pool and OAuth accounts being a finite resource per \`./scripts/agents/pin-account.sh\`, this is wasted capacity.

Findings reference: #22466 (research/SORRY-CLASSIFICATION.md update).

## Scope

Make \`aristotle-agent\` autonomously scale to 0 when there is no work and auto-respawn when work appears, freeing the OAuth slot for researchers/enrichers/mechanics.

## Acceptance Criteria

- [ ] \`scripts/lean/launch.sh\` (or the \`aristotle-agent.sh\` script under \`scripts/aristotle/\` — verify which is the right hook) detects an idle condition:
  - \`./scripts/aristotle/find-candidates.sh --count\` returns 0, AND
  - Count of \`status == \"submitted\"\` in \`research/aristotle-jobs.json\` is 0, AND
  - At least 1 hour has elapsed since the agent's last work cycle
- [ ] When all three are true, the agent gracefully exits (not crashes) and the daemon records it as \`scaled_to_zero\` (new pseudo-status), not \`stuck\`.
- [ ] The daemon's spawn loop checks every minute (or its existing cadence) whether \`find-candidates.sh --count\` returns > 0 and respawns aristotle-agent when it does.
- [ ] \`scripts/lean/launch.sh status\` and \`scripts/lean/launch.sh health\` show \`SCALED_TO_ZERO\` (new state) for aristotle-agent when in this mode, distinct from \`STUCK\` / \`IDLE\` / \`COMPLETED\`.
- [ ] OAuth account allowlist (\`./scripts/agents/pin-account.sh\`) is not modified — the freed slot just becomes available in the existing pool automatically.
- [ ] When the user manually invokes \`./scripts/lean/launch.sh spawn aristotle\`, scale-to-zero is bypassed for one cycle (forced respawn, ignores idle conditions for that single dispatch).

## Notes

[...]

PR #22487, opened 2026-06-05
Aristotle: align submit-batch capacity model + proper 429 backoff

### LINKED ISSUE #22469: Aristotle: align submit-batch capacity model + proper 429 backoff
## Background

\`scripts/aristotle/submit-batch.sh\` currently:
- Defaults \`TARGET_ACTIVE\` to 10 active jobs but the Harmonic server has a hard cap of 5 concurrent projects (we hardcoded a soft \`SERVER_CAPACITY_LIMIT=3\`).
- On 429 / rate-limit, immediately stops the batch (good) but doesn't track or backoff future submissions.
- Sleeps a fixed 3 seconds between submissions regardless of server pressure.
- Has no retry-after parsing.

This contributed to the **200 zombie projects** from 2026-03-24 (see #22467) — we submitted faster than the server could start projects, the server marked them NOT_STARTED, and our reconciliation later abandoned them all.

Findings reference: #22466 (research/SORRY-CLASSIFICATION.md update).

## Scope

Rework \`submit-batch.sh\` capacity logic to match observed server behavior and add proper backoff.

## Acceptance Criteria

- [ ] \`ARISTOTLE_TARGET\` default reduced from 10 to 5 (matches server hard cap; raises only if Harmonic publishes a higher limit).
- [ ] \`SERVER_CAPACITY_LIMIT\` made configurable via env var \`ARISTOTLE_SERVER_CAP\` (default 5). Hardcoded 3 removed.
- [ ] Pre-flight capacity check uses both \`QUEUED\` and \`IN_PROGRESS\` (existing behavior) and additionally **subtracts headroom for STARTING/NOT_STARTED in last 60s** to avoid the zombie-creation race. Implementation note: list with a short \`--limit 100\` and count entries whose status is NOT_STARTED with creation timestamp newer than 60s — if any, refuse to submit until they clear.
- [ ] On 429 response: log the event, **write a \`.aristotle-rate-limit-until\` file** in the project root with a timestamp 5 minutes in the future. Subsequent invocations of \`submit-batch.sh\` short-circuit (exit 0 with a \"rate-limit cooldown active\" message) if this file's timestamp is still in the future.
- [ ] Between successful submissions, replace the fixed 3-second sleep with a backoff of \`5 + 2 * already_submitted_this_batch\` seconds (capped at 30s) to ease pressure as the queue fills.
- [ ] \`aristotle-jobs.json\` records of submissions get an additional optional field \`server_active_at_submit\` capturing the server count at submission time, for forensics.
- [ ] Help text (\`--help\`) updated to document the new env vars and the cooldown file.

## Notes

- Do **not** add unbounded retries on 429 — the backoff via the cooldown file is sufficient and avoids hammering.
- Do **not** parse \`Retry-After\` headers from the CLI output unless they are reliably present in the \`aristotlelib\` output stream (verify in a smoke test first; if absent, fixed 5-min cooldown is fine).
- The cooldown file is local state, not shared across machines. That's intentional — multi-machine submission isn't a current concern.
- Be conservative: it is much better to **under**-submit and produce zero zombies than to submit aggressively. The current empty queue (0 candidates / 3 pending) means we have headroom to be cautious.

## Test Plan

- [ ] \`ARISTOTLE_TARGET=5 ARISTOTLE_SERVER_CAP=5 ./scripts/aristotle/submit-batch.sh --dry-run\` runs successfully with the new defaults visible in output.
- [ ] Create \`.aristotle-rate-limit-until\` with a future timestamp by hand; \`./scripts/aristotle/submit-batch.sh --dry-run\` short-circuits with a cooldown message.
- [ ] Remove the cooldown file; the script proceeds normally.
- [ ] Existing dry-run paths still print server capacity info without error.

## Affected Files

- \`scripts/aristotle/submit-batch.sh\`
- \`.gitignore\` (add \`.aristotle-rate-limit-until\`)

rjwalters:
## Summary

Reworks `scripts/aristotle/submit-batch.sh` to match the Harmonic server's observed capacity behavior and add a proper 429 backoff mechanism, addressing the zombie-creation race that produced 200 stale projects on 2026-03-24 (#22467).

- **Capacity defaults realigned**: `ARISTOTLE_TARGET` 10 -> 5, `SERVER_CAPACITY_LIMIT` 3 -> 5 (configurable via `ARISTOTLE_SERVER_CAP`).
- **Pre-flight NOT_STARTED freshness check**: if any NOT_STARTED project was created in the last 60s, defer the batch.
- **429 cooldown file**: writes `.aristotle-rate-limit-until` (UTC, 5min) on rate-limit; subsequent invocations short-circuit until expiry. File added to `.gitignore`.
- **Progressive backoff**: replaces fixed 3s inter-submission sleep with `5 + 2 * already_submitted` (cap 30s).
- **Forensics**: optional `server_active_at_submit` field on jobs records.
- **Help text** updated for the new env vars and cooldown file.

Conservative by design: under-submit beats over-submit. No retries on 429 — the cooldown file is the entire backoff mechanism.

## Test plan

- [x] `bash -n scripts/aristotle/submit-batch.sh` passes (syntax check).
- [x] `--help` shows new env vars (`ARISTOTLE_TARGET`, `ARISTOTLE_SERVER_CAP`) and cooldown file path.
- [x] `ARISTOTLE_TARGET=5 ARISTOTLE_SERVER_CAP=5 ./submit-batch.sh --dry-run` prints `Target active jobs: 5` and proceeds past pre-flight.
- [x] Hand-write a future timestamp into `.aristotle-rate-limit-until`; subsequent run short-circuits with "Rate-limit cooldown active until ... (300s remaining)".
- [x] Hand-write an expired timestamp; script logs "Cooldown expired ... clearing file" and continues.
- [x] Hand-write garbage; script logs parse-error message, removes the file, and continues.
- [x] `ARISTOTLE_TARGET=0` short-circuits with "No submissions needed".

Closes #22469

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,7 +78,6 @@ package-lock.json
 # Erdos scraper cache
 .erdos-cache/
 aristotle-results/
-.aristotle-rate-limit-until
 .loom/signals/
 .loom/state/
 .loom/tokens
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,7 @@ package-lock.json
 # Erdos scraper cache
 .erdos-cache/
 aristotle-results/
+.aristotle-rate-limit-until
 .loom/signals/
 .loom/state/
 .loom/tokens
--- a/scripts/aristotle/submit-batch.sh
+++ b/scripts/aristotle/submit-batch.sh
@@ -15,9 +15,11 @@
 #
 # Rate-limit cooldown:
 #   On a 429 / rate-limit response, the script writes
-#   `.aristotle-rate-limit-until` in the project root with a UTC timestamp
-#   5 minutes in the future. Subsequent invocations short-circuit (exit 0)
-#   until that timestamp passes. Remove the file manually to force resume.
+#   `.loom/state/aristotle-rate-limit-until` with a UTC timestamp 5 minutes
+#   in the future. Subsequent invocations short-circuit (exit 0) until that
+#   timestamp passes. Remove the file manually to force resume. This path is
+#   shared with the aristotle-agent (issue #22471) which reads the same file
+#   for scale-to-zero coordination.
 #
 # Submission order: Tier 0 StatementOnly files (*StatementOnly.lean) first,
 # then Tier 1 companion files (*Aristotle.lean), then Tier 2 research output
@@ -31,7 +33,7 @@ PROJECT_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
 JOBS_FILE="$PROJECT_ROOT/research/aristotle-jobs.json"
 FIND_CANDIDATES="$SCRIPT_DIR/find-candidates.sh"
 PREPROCESS="$SCRIPT_DIR/preprocess-for-aristotle.sh"
-RATE_LIMIT_FILE="$PROJECT_ROOT/.aristotle-rate-limit-until"
+RATE_LIMIT_FILE="$PROJECT_ROOT/.loom/state/aristotle-rate-limit-until"
 RATE_LIMIT_COOLDOWN_SECONDS=300  # 5 minutes
 
 # Colors
@@ -71,8 +73,9 @@ while [[ $# -gt 0 ]]; do
             echo ""
             echo "Rate-limit cooldown:"
             echo "  On a 429 response, the script writes a UTC timestamp 5 minutes in the"
-            echo "  future to .aristotle-rate-limit-until at the project root. Subsequent"
-            echo "  invocations short-circuit (exit 0) until the cooldown expires."
+            echo "  future to .loom/state/aristotle-rate-limit-until. Subsequent invocations"
+            echo "  short-circuit (exit 0) until the cooldown expires. Shared with the"
+            echo "  aristotle-agent (issue #22471) for scale-to-zero coordination."
             echo "  Cooldown file: $RATE_LIMIT_FILE"
             echo ""
             echo "Submission order: Tier 0 StatementOnly files (*StatementOnly.lean) first,"
@@ -95,8 +98,8 @@ if [[ -z "${ARISTOTLE_API_KEY:-}" ]]; then
 fi
 
 # Short-circuit if a rate-limit cooldown is active.
-# The file `.aristotle-rate-limit-until` contains a UTC ISO-8601 timestamp;
-# while `now < timestamp`, the script exits 0 without submitting.
+# The file `.loom/state/aristotle-rate-limit-until` contains a UTC ISO-8601
+# timestamp; while `now < timestamp`, the script exits 0 without submitting.
 check_rate_limit_cooldown() {
     [[ -f "$RATE_LIMIT_FILE" ]] || return 0
 
@@ -157,6 +160,9 @@ write_rate_limit_cooldown() {
         return 1
     fi
 
+    # Defensive: ensure the .loom/state/ directory exists before writing.
+    # The aristotle-agent (issue #22471) also creates this; harmless to mkdir -p both places.
+    mkdir -p "$(dirname "$RATE_LIMIT_FILE")"
     echo "$until_ts" > "$RATE_LIMIT_FILE"
     echo -e "${YELLOW}Wrote rate-limit cooldown until $until_ts to $RATE_LIMIT_FILE${NC}"
 }
--- a/scripts/aristotle/submit-batch.sh
+++ b/scripts/aristotle/submit-batch.sh
@@ -3,14 +3,21 @@
 # submit-batch.sh - Submit a batch of files to Aristotle
 #
 # Usage:
-#   ./submit-batch.sh                    # Submit up to TARGET (default 10) total active
+#   ./submit-batch.sh                    # Submit up to TARGET (default 5) total active
 #   ./submit-batch.sh --count N          # Submit exactly N files
 #   ./submit-batch.sh --dry-run          # Show what would be submitted
-#   ./submit-batch.sh --target N         # Maintain N active jobs (default 10)
+#   ./submit-batch.sh --target N         # Maintain N active jobs (default 5)
 #
 # Environment:
-#   ARISTOTLE_API_KEY - Required for submission
-#   ARISTOTLE_TARGET  - Target number of active jobs (default 10)
+#   ARISTOTLE_API_KEY    - Required for submission
+#   ARISTOTLE_TARGET     - Target number of active jobs (default 5)
+#   ARISTOTLE_SERVER_CAP - Server-side concurrent project cap (default 5)
+#
+# Rate-limit cooldown:
+#   On a 429 / rate-limit response, the script writes
+#   `.aristotle-rate-limit-until` in the project root with a UTC timestamp
+#   5 minutes in the future. Subsequent invocations short-circuit (exit 0)
+#   until that timestamp passes. Remove the file manually to force resume.
 #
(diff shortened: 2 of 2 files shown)


rjwalters:
## Judge Review: CHANGES REQUESTED

Thorough review against the issue spec and verification of all critical checks. The implementation is solid in isolation, but there is a **contract mismatch with PR #22486** that must be resolved before either PR merges.

### Verified working

1. **Tier 0 logging from PR #22481 is preserved**. Grepped the post-rebase file: `T0 statement-only`, `is_statement_only=true`, `StatementOnly submission (T0)`, and the `tier0_count Tier 0 StatementOnly` summary all survive the rebase intact. No regression.
2. **BSD and GNU date handling actually works on both platforms**. Confirmed locally: BSD path (`date -u -j -f ... +%s` / `date -u -v+300S`) parses and round-trips cleanly on macOS Darwin 25.5.0; GNU path (`date -u -d ...`) is the second branch and uses well-formed syntax. Garbage input is correctly rejected (empty `until_epoch`, file removed, script continues).
3. **NOT_STARTED freshness parsing fails safe**. Implementation correctly defaults to "not fresh / proceed" when the CLI invocation fails (`uvx ... || { echo warn; return 0; }`) and when no UUIDs match. Only a positive match of `just now` / `<=60 seconds ago` / `a few seconds ago` / `less than a minute ago` defers. This is the right direction — a parse failure does NOT block all submissions.
4. **Syntax check passes** (`bash -n` clean).
5. **Existing batch logic** (T0/T1/T2 ordering, preprocessing, job-record creation, server capacity probe, signal emission) is untouched and intact.
6. **`server_active_at_submit` field** correctly stores `null` when the probe failed (handles both `-1` sentinel and `_unset` placeholder via the jq conditional).
7. **Progressive backoff** correctly clamped to `min(5 + 2*submitted, 30)`.
8. **CI**: `gh pr checks` shows no checks reported on the branch — this is a bash-only change with no JS/TS CI surface, so that's expected.

### Required change (blocking)

**Cooldown file path is inconsistent with PR #22486.** This PR writes/reads `.aristotle-rate-limit-until` at the project root (matches issue #22469 spec literally), but PR #22486 (scale-to-zero, issue #22471) reads `$PERSIST_DIR/aristotle-rate-limit-until` where `PERSIST_DIR=$REPO_ROOT/.loom/state`. If this PR merges first, PR #22486's `is_rate_limited` check will silently no-op forever because it will be looking at the wrong path.

PR #22486's own description acknowledges: *"Defensively handles missing file — issue #22469 isn't merged yet, so this is a forward-compatible contract."* — but the path it codes against does not match the path issue #22469 specifies, so the contract is broken in the same commit it claims to be forward-compatible with.

**Recommendation**: standardize on `.loom/state/aristotle-rate-limit-until`. That groups with other agent state (already gitignored as a directory), is what the scale-to-zero work expects, and reduces top-level clutter. Concretely in this PR:

```bash
# Replace
RATE_LIMIT_FILE="$PROJECT_ROOT/.aristotle-rate-limit-until"
# With
RATE_LIMIT_FILE="$PROJECT_ROOT/.loom/state/aristotle-rate-limit-until"
mkdir -p "$(dirname "$RATE_LIMIT_FILE")"  # ensure dir exists before write
```

And drop the `.gitignore` entry — `.loom/state/` is already ignored (line 82 in the new `.gitignore`).

Update the `--help` text and the file's header comment to reference the new path.

Issue #22469's acceptance criteria literally says "in the project root", but #22469 was authored before #22471/#22486 introduced `.loom/state/` as the canonical agent-state directory. The right call is to update the spec to match the convention, not to keep two paths in flight. Worth a one-line follow-up comment on issue #22469 noting the deviation; the body of this PR should reflect the chosen path so the issue is closed honestly.

### Minor (non-blocking, optional)

- The `_unset` sentinel pattern for `server_active_at_submit` is clever but a bit indirect. A simpler pattern would be to omit the field via `|del(.server_active_at_submit | select(. == null))` when null. Not a blocker — current code works.
- `check_fresh_not_started`'s awk-based column-strip (`{ $1=""; $2=""; sub(/^  /, ""); print }`) is brittle if the CLI output format changes (e.g. extra columns inserted between STATUS and CREATED). A short comment pointing to a sample line from the CLI output would help future maintainers. Not blocking.

### Summary

Approve once the cooldown file path is reconciled with PR #22486 — either change this PR to `.loom/state/aristotle-rate-limit-until`, or coordinate with #22486 to use the project root path. Recommend the former.

rjwalters:
Addressed Judge feedback in 5596684c253:

- `RATE_LIMIT_FILE` moved from `$PROJECT_ROOT/.aristotle-rate-limit-until` to `$PROJECT_ROOT/.loom/state/aristotle-rate-limit-until` (matches PR #22486's `$PERSIST_DIR/aristotle-rate-limit-until` where `PERSIST_DIR=$REPO_ROOT/.loom/state`).
- `mkdir -p "$(dirname "$RATE_LIMIT_FILE")"` added before write to ensure the directory exists on a fresh checkout.
- Header comment, `--help` text, and the `check_rate_limit_cooldown()` comment all updated to reference the new path and note the shared contract with #22471/#22486.
- `.gitignore`: dropped the now-unused `.aristotle-rate-limit-until` line — `.loom/state/` (line 82 of the post-change file) already covers the new path.

Verified manually in the worktree:
1. Writing a future timestamp to `.loom/state/aristotle-rate-limit-until` short-circuits with the expected message.
2. Removing the file lets the script proceed past `check_rate_limit_cooldown`.
3. An empty file is gracefully detected and removed, then the script proceeds.

`bash -n` clean. Returning to `loom:review-requested`.

rjwalters:
**APPROVE** (re-judge after Doctor fix `5596684c`)

Path-mismatch fix verified end-to-end:

- All four references to the cooldown file in `scripts/aristotle/submit-batch.sh` (header doc L18, `RATE_LIMIT_FILE` assignment L36, `--help` text L76, `check_rate_limit_cooldown` comment L101) now use `$PROJECT_ROOT/.loom/state/aristotle-rate-limit-until` — `grep -rn "aristotle-rate-limit-until" scripts/ .gitignore` shows zero leftover references to the old `$PROJECT_ROOT/.aristotle-rate-limit-until` path. The contract with #22486 / #22471 is now satisfied.
- `mkdir -p "$(dirname "$RATE_LIMIT_FILE")"` at L165 in `write_rate_limit_cooldown()` runs immediately before the first write at L166. Defensive and idempotent with the aristotle-agent's own mkdir.
- `.gitignore`: the stale `.aristotle-rate-limit-until` line was removed in this commit, and coverage now comes solely from the pre-existing `.loom/state/` directory entry at line 82 — no redundant new line was added.
- Tier 0 logging from #22481 is intact (`Submitting [T0 statement-only]` at L326, `submission_tier="T0"` at L424, `StatementOnly submission (T0)` at L425, tier counts at L553/L570).
- All earlier APPROVE-side findings are unchanged: BSD/GNU date branches at L118-121 and L146-149 with the epoch-math fallback at L152-156, fresh-NOT_STARTED CLI-failure fail-safe at L243-247 (returns 0 / proceeds cautiously), strict `<= 60s` bound at L267, `ARISTOTLE_TARGET=5` / `ARISTOTLE_SERVER_CAP=5` defaults, `server_active_at_submit` forensics field on each jobs entry.

`mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`. Cleared for merge.

PR #20839, opened 2026-05-27
Integrate AlphaProof Nexus proof for Erdős #152 (weak form resolved)

rjwalters:
## Summary

- Ports the DeepMind AlphaProof Nexus Lean 4 proof for Erdős Problem #152 (weak form) into `proofs/Proofs/Erdos152ProblemAPN.lean` (538 lines, namespace `Erdos152APN`).
- Updates `src/data/proofs/erdos-152/meta.json` to reflect the partial resolution: weak form RESOLVED (2026-05-21); strong form and infinite version remain open.
- Extends gallery `historicalContext`, `proofStrategy`, `keyInsights`, `conclusion.openQuestions` and adds 4 new references (ESS94 original, APN2026 arXiv, APNCode152 Lean source, APNNL152 natural-language proof).

## Scope check (per issue body)

The DeepMind file proves `tendsto_f : Tendsto f atTop atTop` via the explicit bound `16 * num_isolated A + 100*n + 16 >= n*n` (so `f(n) >= (n^2 - 100n - 16)/16`). This corresponds to the WEAK form of `ErdosProblem152` (lim -> infty).

The STRONG form `ErdosProblem152Strong` — `exists c > 0, forall Sidon A, I(A) >= c|A|^2` — is **NOT** explicitly discharged by the DeepMind file. The same inequality implies asymptotic `f(n) >> n^2`, but the universally quantified statement (an absolute constant valid for all Sidon sets including small ones) requires a separate argument that the AlphaProof file does not provide. The infinite version `ErdosProblem152Infinite` also remains open.

Consequently `erdosProblemStatus` is set to `partially-solved` (not `solved`), and `status` remains `axiomatized` with `badge: wip`. The `assumptions` field documents this precisely.

## Adaptations to the source file

Three changes were applied verbatim to the 517-line DeepMind file:

1. `import FormalConjectures.Util.ProblemImports` -> `import Mathlib`
2. `namespace Erdos152` -> `namespace Erdos152APN` (avoids collision with the gallery file)
3. `target_theorem_0 : answer(True) <-> Tendsto f atTop atTop` -> `: True <-> Tendsto f atTop atTop` (replaces the FormalConjectures-specific elaborator)

A header docstring was added to record the source URL, paper, and the original Apache-2.0 copyright. The proof body is otherwise unchanged.

## meta.json sanity checks

- `meta.axiomCount = leanFile.axiomCount = 0` (verified by `grep -c '^axiom '` on both files)
[...]

PR #20838, opened 2026-05-27
research(erdos-741): integrate AlphaProof Nexus parts i + ii

rjwalters:
Closes #20737

## Summary

Ports DeepMind's AlphaProof Nexus Lean 4 proofs (2026-05-21, arXiv:2605.22763v1) for both parts of **Erdős Problem #741** ("Splitting Sumsets with Positive Density") into the gallery's `erdos-741` entry.

| Erdős #741 sub-question | Answer (AlphaProof Nexus) | New companion file |
|---|---|---|
| (i) ∀ A with A+A positive density, ∃ A₁,A₂ disjoint with both Aᵢ+Aᵢ positive density? | **NO** (`False`) | `proofs/Proofs/Erdos741ProblemAPNPartI.lean` |
| (ii) ∃ basis A of order 2 with no syndetic split? | **YES** (`True`) | `proofs/Proofs/Erdos741ProblemAPNPartII.lean` |

### New Lean companion files

- `proofs/Proofs/Erdos741ProblemAPNPartI.lean` — 1529 lines, namespace `Erdos741APN_I`, 1 theorem (`target_theorem_0`), 106 lemmas, 14 defs (7 regular + 7 noncomputable). 0 axioms, 0 sorries.
- `proofs/Proofs/Erdos741ProblemAPNPartII.lean` — 681 lines, namespace `Erdos741APN_II`, 1 theorem, 25 lemmas, 13 defs (9 regular + 4 noncomputable). 0 axioms, 0 sorries.

Both files are verbatim ports of
  https://github.com/google-deepmind/alphaproof-nexus-results/tree/main/APNOutputs/ErdosProblems
(licensed Apache-2.0, original Google LLC copyright preserved in the file headers). The only edits are:
1. `import FormalConjectures.Util.ProblemImports` → `import Mathlib`
2. `answer(True)` → `True`, `answer(False)` → `False` (no behaviour change — the `answer(...)` elaborator just elaborates to its argument)
3. `namespace Erdos741` → `namespace Erdos741APN_I` / `Erdos741APN_II` to avoid collision with the existing `Erdos741Problem.lean` definitions
4. Appended explicit `end Erdos741APN_{I,II}` declarations

Construction outline for **part (i)** (negative answer):
- Base-4 splitter functions `split1`, `split2` such that `split1(n) + split2(n) = n` and each output avoids digit `3` in base 4
[...]

PR #20837, opened 2026-05-27
research(erdos-26): integrate AlphaProof Nexus Tenenbaum-variant proof

rjwalters:
## Summary

Ports DeepMind's AlphaProof Nexus Lean proof (2026-05-21, arXiv:2605.22763v1) for the **Tenenbaum variant of Erdős Problem #26** as a sibling companion file alongside the existing axiomatized main proof.

The Tenenbaum variant asks: for every ε > 0, does some k = k(ε) exist such that the shift A + k of a thick sequence A has lower density ≥ 1 − ε of multiples? AlphaProof Nexus settled this **negatively for ε = 1/2** by constructing an explicit thick sequence whose multiple sets attain lower density at most 37/90 < 1/2 for every shift k.

Sources:
- https://github.com/google-deepmind/alphaproof-nexus-results/blob/main/APNOutputs/ErdosProblems/erdos_26.variants.tenenbaum.lean (1066 lines, Apache-2.0)
- Paper: arXiv:2605.22763v1 (2026-05-21)

### New Lean companion file

`proofs/Proofs/Erdos26APN_Tenenbaum.lean` (1123 lines) is a near-verbatim port from the AlphaProof Nexus source. **0 axioms, 0 sorries** in this companion file.

Substitutions from the upstream source:

1. `import FormalConjectures.Util.ProblemImports` → `import Mathlib`
2. `namespace Erdos26` → `namespace Erdos26APN_Tenenbaum` (avoids collision with the existing `Erdos26Problem.lean`)
3. `answer(False)` → `False` (FormalConjectures' `answer(...)` elaborator is the identity)
4. `∀ᵉ (A : ℕ → ℕ)` → `∀ (A : ℕ → ℕ)` (single bare binder, no membership clause)
5. Inlines `Set.partialDensity`, `Set.HasDensity`, `Set.lowerDensity` from `FormalConjecturesForMathlib/Data/Set/Density.lean` (Apache-2.0, verbatim) since this repo does not vendor that file. The source uses dot-notation `(set).lowerDensity` / `(set).HasDensity α` which requires the `Set` namespace.

The original Apache-2.0 license header (copyright Google LLC 2025) is preserved at the top of the file. The construction outline:

- Iterative block construction `A_j` indexed by levels j ≥ 1, where block j uses arithmetic progressions matched to growing prime moduli q > 10^j (`exists_x_P_ind`, `exists_next_state`).
- `seqA : ℕ → ℕ` enumerates the resulting infinite set via `Nat.nth`.
[...]

PR #20836, opened 2026-05-27
research(erdos-138): integrate AlphaProof Nexus difference-variant proof

rjwalters:
## Summary

Ports DeepMind AlphaProof Nexus's Lean proof for the **difference variant** of Erdős Problem #138 into our gallery, adapted from `FormalConjectures.Util.ProblemImports` to our Mathlib-only setup pinned at v4.26.0.

The difference variant asks: does `W(k+1) - W(k) -> infinity` as `k -> infinity`? AlphaProof Nexus resolved this affirmatively on 2026-05-21 via the unconditional bound `W(k+1) - W(k) >= k`.

Sources:
- https://github.com/google-deepmind/alphaproof-nexus-results/blob/main/APNOutputs/ErdosProblems/erdos_138.variants.difference.lean (918 lines)
- Paper: arXiv:2605.22763v1 (2026-05-21)

## Changes

- **`proofs/Proofs/Erdos138Problem.lean`**: Replaces the prior axiomatized stub (196 lines) with the full coloring-extension argument (816 lines). Inlines `Set.IsAPOfLengthWith`, `Set.IsAPOfLength`, and `ContainsMonoAPofLength` from `FormalConjecturesForMathlib` so the file depends only on Mathlib. Proves the unconditional bound `W_diff_ge_k_all : forall k, W (k+1) - W k >= k`, then derives the difference variant `W_difference_diverges_nat` via `Filter.tendsto_atTop_mono`. The Ultrafilter machinery used in the original Nexus output to *prove* van der Waerden's theorem from `Combinatorics.exists_mono_homothetic_copy` is replaced by a single axiom `W_is_nonempty`, since that Mathlib lemma is not in our pinned v4.26.0.
- **`src/data/proofs/erdos-138/meta.json`**: Renamed to "Van der Waerden Numbers: Growth, Difference, and Quotient Variants" to match the broader scope (the file now treats all four variants Erdős posed in 1980-1981). Updated counts: `axiomCount` 8 -> 9 (adds `W_is_nonempty`), `theoremCount` 4 -> 21, `definitionCount` 9 -> 13, `lineCount` 196 -> 816. Refined `assumptions` field describes the role of each axiom, added DeepMind AlphaProof Nexus + arXiv references, updated `historicalContext` with a paragraph noting the 2026-05-21 resolution, and updated `sections`/`conclusion`/`openQuestions` to reflect the resolved difference variant.

## Gallery title vs. statement framing

[...]

PR #22600, opened 2026-06-08
tooling(lint): catch in-body namespace-qualified refs after rename

### LINKED ISSUE #20847: Tooling: lint/script to catch in-body namespace-qualified refs after rename
Architectural follow-up from /sweep of epic #20732 (7 AlphaProof Nexus integrations).

## Problem

Three of the seven AlphaProof Nexus port PRs (#20836, #20838, #20839) needed a Doctor cycle for the same class of bug: after the `sed`-style namespace rename (`namespace Erdos741` → `namespace Erdos741APN_I`, etc.), in-body references inside tactic invocations like `delta Erdos741.foo at h`, `simp_all [Erdos152.D_set]`, `change Erdos26.X_seq` kept the old namespace prefix.

The naïve `sed s/^namespace Erdos741/namespace Erdos741APN_I/` only catches the `^namespace`/`^end` lines, missing every in-body occurrence. The errors only surface at docker-build time. PR #20837 had one such leftover marked non-blocker; #20838 had three; #20839 had two — same root cause, three Doctor cycles burned.

## Proposal

A small pre-PR check (script or Lean linter) that, given a file with `namespace Foo` declared, flags any token of the form `Foo.<Ident>` inside the file body. The flagged occurrences can then be either qualified to the renamed namespace or stripped to the unqualified form before the PR opens.

Rough sketch (Bash/grep is enough for the common case):

```bash
ns=$(grep -oP "^namespace \K\S+" "$file" | head -1)
old_ns=$(git log -p -- "$file" | grep -oP "^-namespace \K\S+" | head -1)
[ -n "$old_ns" ] && grep -n "\b${old_ns}\.[A-Z]" "$file"
```

More robust as a Lean-syntax-aware linter, but `grep`-based catches >90% of cases.

## Acceptance

- Script or check available in the repo (e.g., `scripts/lint/check-renamed-namespace-refs.sh`).
- A future port of an AlphaProof Nexus file would catch the namespace-leftovers before the builder opens the PR.
- Optional: integrate into the builder skill prompt so this is part of the pre-push checklist.

This is a small tooling investment; /architect or /hermit could scope it.

rjwalters:
## Summary

Adds `scripts/lint/check-renamed-namespace-refs.sh`, a small bash script
that catches in-body tactic-position references to stale namespace
prefixes left behind after a `sed`-style namespace rename (e.g.,
`namespace Erdos741` -> `namespace Erdos741APN_I`).

Naive `sed s/^namespace Foo/.../` anchors on `^namespace`/`^end` lines
and silently misses tactic references like `delta Erdos741.foo at h`,
`simp_all [Erdos152.D_set]`, `change Erdos26.X_seq`. Each leftover
surfaces as a Lean error at docker-build time, costing a Doctor cycle
per file. PRs #20836, #20837, #20838, #20839 from the AlphaProof Nexus
epic (#20732) burned three Doctor cycles on the same root cause.

## Approach (per curator enrichment in #20847)

For each Lean file:
1. Collect every `^namespace <X>` declared in the file (LOCAL_NS).
2. For each declared namespace, strip recognized rename suffixes
   (`APN_I`, `APN_II`, `APN_Tenenbaum`, `APN`, `Nexus`, `PartI`,
   `PartII`, `_I`, `_II`) to compute the pre-rename "base".
3. If the base is not itself a declared namespace in the file, scan the
   body for `\b<base>\.[A-Za-z_]` references. `--` line comments and
   `^namespace`/`^end` lines are pre-filtered.

Exit codes: 0 = clean, 1 = stale refs found (reported `file:line:body`
to stderr), 2 = usage error.

## Files

- `scripts/lint/check-renamed-namespace-refs.sh` (NEW, executable) —
  the lint script. Header comments document usage, the motivating epic
  PRs, and the v1 limitations.
- `scripts/lint/test/positive-stale-ref.lean` — synthetic positive
  fixture (declared `Erdos741APN_I` with body refs to `Erdos741.foo`,
  `Erdos741.bar`).
- `scripts/lint/test/negative-clean.lean` — synthetic negative fixture
  (all refs properly qualified or unqualified).
- `scripts/lint/test/negative-no-namespace.lean` — no `^namespace`
  declaration (should silently exit 0).
- `scripts/lint/test/negative-nested-namespace.lean` — two declared
  namespaces with legitimate cross-namespace refs (should not be
  flagged).
- `scripts/lint/test/run-tests.sh` — test runner with five cases.

## Verification

- `scripts/lint/test/run-tests.sh` — 10/10 assertions pass.
- Smoke test against live `proofs/Proofs/Erdos{741,152,26,12-Part-I,846}*APN*.lean`:
  clean (exit 0, no stderr).
- Smoke test against `proofs/Proofs/Erdos12ProblemAPNPartII.lean`
  (known to retain legitimate `Erdos12.P_n_def` body refs per #20847
  commentary): correctly flags four occurrences on lines 639, 641,
  642, 646. This demonstrates the script catches the exact bug class
  the issue describes.
- `shellcheck` clean on both scripts.

## Bash compatibility

Targets bash 3.2+ (macOS default). Avoids `mapfile` / process
substitution into array reads; uses heredoc `while read` loops and
sort/grep for deduplication.

## Out of scope (deferred)

- Full Lean-syntax-aware linting (would need a real parser).
- Auto-fix (script only reports).
- CI / pre-commit integration (separate `loom:issue` if desired).
- Builder skill prompt integration ("optional" per the issue body).

Closes #20847

## Test plan

- [x] `bash scripts/lint/test/run-tests.sh` — 10/10 pass
- [x] Smoke test against live APN files — all clean
- [x] Confirmed real-world dirty case on `Erdos12ProblemAPNPartII.lean` is caught
- [x] `shellcheck` clean
- [x] Usage error path (`./check-...sh` with no args) -> exit 2
- [x] Missing-file error path -> exit 2

### DIFF
--- a/scripts/lint/check-renamed-namespace-refs.sh
+++ b/scripts/lint/check-renamed-namespace-refs.sh
@@ -1,0 +1,160 @@
+#!/usr/bin/env bash
+# scripts/lint/check-renamed-namespace-refs.sh
+#
+# Catch in-body references to stale namespace prefixes left behind after
+# a `sed`-style namespace rename (e.g., `namespace Erdos741` -> `namespace
+# Erdos741APN_I`). Naive renames anchor on `^namespace`/`^end` lines and
+# silently miss tactic-position references like:
+#
+#   delta Erdos741.foo at h
+#   simp_all [Erdos152.D_set]
+#   change Erdos26.X_seq
+#   unfold Erdos138.thing
+#
+# Each leftover surfaces as a Lean error at docker-build time, costing a
+# Doctor cycle per file. This script catches them pre-PR.
+#
+# Motivating PRs (epic #20732, AlphaProof Nexus port):
+#   - #20836, #20837, #20838, #20839 — three Doctor cycles burned on the
+#     same root cause.
+#
+# Usage:
+#   scripts/lint/check-renamed-namespace-refs.sh <file.lean> [<file.lean> ...]
+#   scripts/lint/check-renamed-namespace-refs.sh proofs/Proofs/*.lean
+#
+# Behavior:
+#   - For each file, collect every `^namespace <X>` declared in the file
+#     (call this set LOCAL_NS).
+#   - For each local namespace, compute a "stripped base" by removing
+#     trailing rename suffixes (APN_I, APN_II, APN, _Tenenbaum, Nexus,
+#     PartI, PartII, ...). Example: Erdos741APN_I -> Erdos741.
+#   - If the stripped base differs from the declared namespace AND is not
+#     itself one of the declared namespaces, scan the file body for any
+#     reference of the form `\b<base>\.[A-Za-z_]`. Each match is a likely
+#     stale reference from the pre-rename source.
+#   - Comment-strip (`--` line comments) before scanning to reduce noise.
+#     `^namespace <X>` and `^end <X>` declaration/closer lines are skipped.
+#
+# Exit codes:
+#   0 — no stale references found in any file
+#   1 — at least one stale reference found (printed to stderr)
+#   2 — usage error
+#
+# Limitations (v1, by design):
+#   - grep-based, not a full Lean parser. Block comments (`/- ... -/`) and
+#     string literals are NOT stripped; on real APN files this catches >90%
+#     of cases (see issue #20847).
+#   - Only flags suspects derived from the file's own declared namespaces.
+#     Cross-file stale refs (file declares `Foo`, references `Bar.x` where
+#     `Bar` lives elsewhere) are out of scope.
+#   - Targets POSIX-ish bash 3.2+ (macOS default) — avoids `mapfile`.
+#
+# See also: issue #20847.
+
+set -u
+
+if [ "$#" -eq 0 ]; then
+  echo "usage: $0 <file.lean> [<file.lean> ...]" >&2
+  exit 2
+fi
+
+# Suffixes we strip when computing the "old base" namespace name from a
+# declared (renamed) namespace. Order matters: longest first so we strip
+# `APN_I` before `APN`. Adjust here when new rename patterns appear.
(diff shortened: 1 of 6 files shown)


rjwalters:
## Loom Judge: Approved

**Verification (bash 3.2.57 on darwin):**
- All 10 test runner assertions pass (4 cases: positive fixture, clean negative, no-namespace negative, cross-namespace negative; plus 6-file APN smoke-test).
- Spot-checked real file: `./scripts/lint/check-renamed-namespace-refs.sh proofs/Proofs/Erdos12ProblemAPNPartII.lean` exits 1 and surfaces 4 `Erdos12.P_n_def` matches at lines 639/641/642/646 — matches the builder's claim exactly.
- No false positives on the six post-Doctor APN files (`Erdos741I/II`, `Erdos152`, `Erdos26_Tenenbaum`, `Erdos12I`, `Erdos846`) — exit 0, empty stderr.
- Bash 3.2 compatibility verified: no `[[ ... =~ ]]`, no `mapfile`/`readarray`, no `declare -A`, no `${var,,}`/`${var^^}`. Uses heredocs + awk for portability; explicit comment at line 50 documents the 3.2 target.
- Both scripts executable (`0755`).

**Design notes (LGTM):**
- Longest-first suffix stripping order (`APN_I`/`APN_II`/`APN_Tenenbaum` before `APN`) correctly handles the Erdos26 case.
- Cross-namespace exemption (base is itself a declared namespace) correctly handles the `Foo`+`FooAPN_I` fixture.
- `awk` comment-stripping with `index(line,"--")` is a reasonable v1 — limitations (block comments / strings not stripped) are documented at lines 43-49.
- `set -u` (not `set -e`) is the right call given the heredoc-driven empty-string loops.

Approving — moving to `loom:pr`.

PR #23358, opened 2026-06-14
Enrich gallery: remove 34 dead crossReference links

rjwalters:
## Summary
The `crossReferences` field on each proof renders as a clickable "Related Proofs" link to `/proof/<targetId>` (`src/components/proof/ProofConclusion.tsx:192`, `ProofOverview.tsx`). A scan of all 2447 gallery entries found **34 crossReferences pointing to slugs with no gallery directory** — each produces a dead link that lands on a "proof not found" page. (Prior xref scans targeted `relatedProofs`; this field was uncovered.)

- **Removed 34 dead-link crossReferences** across 34 entries. Targets fall into two groups, both confirmed to have no gallery directory:
  - never-created sibling/WIP entries (e.g. `amgm-inequality-oq-04-oq-01`, `binary-gcd-oq-03-oq-01`, `cayley-hamilton-minpoly-...-wip-02`, `szemeredi-full-oq-01`, `brouwer-fixed-point-oq-04-oq-01`)
  - concept references with no gallery page (`catalan-numbers`, `jordan-normal-form`, `rational-canonical-form`, `probabilistic-method`, `gauss-bonnet`, `ackermann-function`, `easton-theorem`, `dilworth-theorem`, `jacobis-four-square-theorem`, `sperner-lemma`, `fermat-little-theorem`, `ehrhart-polynomials`, `cauchy-theorem`, `triangle-inequality-oq-02`)
- **Repointed** `euler-identity-oq-01-oq-01-oq-01`: `demoivre` → `de-moivre` (the existing entry; only safe normalization — all other broken targets fuzzy-match *different* theorems, so repointing would create plausible-but-wrong links).

## Validation
Data-only JSON change touching only `crossReferences` blocks (45 insertions / 215 deletions). All 34 files re-parse as valid JSON. `pnpm annotations:build`, `research:build`, `research:enrich`, and `tsc -b` all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #23360, opened 2026-06-14
research: strip 2 self-referential relatedProofs xrefs

rjwalters:
## Summary
- `triangle-inequality-oq-02` and `yang-mills-mass-gap` each listed their **own slug** in `relatedProofs`.
- Neither slug has a gallery dir (`src/data/proofs/<slug>/`), so the self-link renders as a dead `/proof/` link.
- Removed only the self-reference from each; all valid sibling xrefs are preserved.

## Why this is distinct
Self-references are a separate, unambiguously-lossless category — untouched by the open xref PRs (#23353–57, #23349–51 target sibling/concept slugs; #23358 is gallery `crossReference`). A proof relating to itself carries no information.

## Verification
- `jq -e .` confirms both JSONs remain valid.
- Net change: `triangle-inequality-oq-02` → `[triangle-inequality, -oq-03, -oq-04]`; `yang-mills-mass-gap` → `[yang-mills]`.
- Build-free (research JSON data only).

### DIFF
--- a/src/data/research/problems/triangle-inequality-oq-02.json
+++ b/src/data/research/problems/triangle-inequality-oq-02.json
@@ -57,7 +57,6 @@
   ],
   "relatedProofs": [
     "triangle-inequality",
-    "triangle-inequality-oq-02",
     "triangle-inequality-oq-03",
     "triangle-inequality-oq-04"
   ],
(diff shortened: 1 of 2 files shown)


rjwalters:
Exact duplicate of #23359 (same two files, same self-ref strips). #23359 shipped first; closing to avoid redundant queue entry / merge collision.

PR #23646, opened 2026-06-14
Enrich erdos-1110: realign drifted gallery sections to full file (8→9)

rjwalters:
## Summary

The `sections` array in `erdos-1110/meta.json` had drifted from an older revision of `Proofs/Erdos1110Problem.lean`. Coverage stopped at line **265 of 437**:

- **Part II (the {2,3} case)** actually spans lines **80-277** — including the full strong-induction proof `case_2_3_all_representable` and its eight helper lemmas — but the section claimed only 78-110.
- **Parts III-VIII** were each mislabeled to early line numbers (111-265), hiding the Yu-Chen density/coprime results, the minimum-summand bounds, the examples, and the summary theorems.

## Changes

- Realigned all sections to the file's own **Part I-VIII** banners; now contiguous, covering lines 1-437 (added a header section).
- Surfaced the previously-hidden {2,3} inductive proof and the back-half parts.
- Replaced the auto-generated placeholder `mathContext` values ("Mathematical content: …", "Verified: …", "Bound: …") with real LaTeX for every section.
- Expanded the summaries to name the actual lemmas, theorems, and definitions.

Only `sections` changed; all other meta.json content is byte-identical to `origin/main`. Counts (13 theorems / 7 definitions / 1 axiom / 0 sorries) verified correct against the source.

## Validation

`pnpm research:build` exits 0 (accepts meta.json). Full `pnpm build` is not runnable in the linked worktree (no local node_modules for the `tsc`/`vite` step); per-proof JSON is fetched at runtime so that step does not affect this metadata change.

### DIFF
--- a/src/data/proofs/erdos-1110/meta.json
+++ b/src/data/proofs/erdos-1110/meta.json
@@ -86,68 +86,76 @@
   },
   "sections": [
     {
+      "id": "header",
+      "title": "Problem Statement and Imports",
+      "summary": "Module docstring stating Erdős Problem #1110 (for coprime p > q ≥ 2, representing n as a sum of p^k q^l terms with no term dividing another), the key results (Erdős-Lewin 1996, Yu-Chen 2022, Yang-Zhao 2025), Erdős's 'silly conjecture' anecdote, and references. Imports Nat/Finset/Real and NumberTheory.Divisors; opens Finset and the Erdos1110 namespace.",
+      "startLine": 1,
+      "endLine": 43,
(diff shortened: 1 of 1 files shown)


rjwalters:
Closing as a duplicate of #23642 (R1), which was opened ~2 min earlier and performs the equivalent full-file section realignment for erdos-1110 with real mathContext on every section. #23642 should be merged instead. (My version additionally split out a standalone header section 1-43, but that's a minor delta and not worth a conflicting PR on the same meta.json.)

PR #24250, opened 2026-06-15
research(wilsons-theorem-oq-02-ext-oq-01): general two-involution trick for finite abelian groups

rjwalters:
## Summary

Answers OQ-01 of `wilsons-theorem-oq-02-ext`: **yes**, the two-involution trick generalizes from `(ZMod n)ˣ` to arbitrary finite abelian groups — and the general statement is *more elementary* than the proven specialization.

**General Gauss-Wilson theorem.** For a finite abelian group `G` with `S = {x : x²=1}`: `∏_{x∈G} x` equals the unique element of order two if one exists, and `1` otherwise.

## Key ORIENT insight

In the proven `(ZMod n)ˣ` proof `prod_units_one_of_not_cyclic_ext` (`WilsonsTheoremOQ02Ext.lean:362-436`), the **only** ZMod-specific step is line 373, which derives `3 ≤ |S|` from `¬IsCyclic` via `GaussWilsonNonCyclic` (CRT + the `2^k` case split). Everything from `prod_eq_prod_sq_eq_one` onward is already stated for an arbitrary finite `CommGroup`. Taking `3 ≤ |S|` as a hypothesis sheds that entire machinery.

## New file `proofs/Proofs/WilsonsTheoremOQ02ExtOQ01.lean` (0 sorries, 0 axioms)

- `prod_eq_one_of_three_le_card_sqrt_one` — `3 ≤ |{x:x²=1}|` ⇒ `∏ G = 1` (the general two-involution trick)
- `prod_eq_one_of_no_involution` — no element of order 2 ⇒ `∏ G = 1`
- `prod_eq_unique_involution` — unique `t` of order 2 ⇒ `∏ G = t`
- `prod_eq_one_or_unique_involution` — full characterization

Reuses the public `WilsonsTheoremOQ02Ext` lemmas; the heart theorem is a near-verbatim transcription of the proven proof.

## Honesty / Mathlib gap

Mathlib v4.26.0 has only the field-units case (`FiniteField.prod_univ_units_id_eq_neg_one`, behind `ZMod.wilsons_lemma`). No general finite-abelian-group version exists.

## Verification

[...]

PR #24258, opened 2026-06-15
research(wilsons-theorem-oq-02-ext-oq-02): Gauss-Wilson extends to rings of integers

rjwalters:
## Summary

Open question (`wilsons-theorem-oq-02-ext-oq-02`): does the Gauss-Wilson theorem extend from `(ℤ/n)ˣ` to the unit groups of quotients `O_K/𝔞` for rings of integers `O_K` of number fields?

**Answer: YES**, for a structural reason requiring nothing specific to number fields. For a nonzero ideal `𝔞`, `R = O_K/𝔞` is a *finite commutative ring*, so `Rˣ` is a finite abelian group, and the general two-involution theorem proved for the sibling problem OQ-02-ext-oq-01 (#24250) applies verbatim:

> `∏ u : Rˣ, u` = the unique element of order 2 of `Rˣ` if one exists, else `1`.

This PR is therefore a one-line corollary plus the number-theoretic refinement and an exact certificate.

## What's here

- **`proofs/Proofs/WilsonsTheoremOQ02ExtOQ02.lean`**
  - `prod_units_eq_one_or_unique_involution` — Gauss-Wilson for any finite commutative ring (instantiation at `G = Rˣ`).
  - `prod_units_coe_eq_neg_one` — classical `-1` packaging when `-1` is the unique involution (analogue of `n ∈ {1,2,4,pᵏ,2pᵏ}`).
- **Refinement**: by CRT, `∏ u = -1` iff `-1` is the *unique* involution of `(O_K/𝔞)ˣ`; local counts mirror `(ℤ/pᵏ)ˣ` with `N(𝔭) = p^f` in place of `p`.
- **New phenomenon** (absent for `ℤ/n`, where `∏` is always `±1`): when `-1 = 1` in `R` (residue char 2), the product can be an order-2 unit that is neither `1` nor `-1` — e.g. `ℤ[i]/(2)` has units `{1, i}` and `∏ u = i`.
- **Exact certificate** (`research/.../verification/wilsons_oq02_ext_oq02_number_rings.py`): HNF-lattice enumeration of `O_K/(β)` unit groups over `ℤ[i], ℤ[ω], ℤ[√-2], ℤ[√2], ℤ[(1+√5)/2]`. **94 quotients, 94/94 match the prediction, 0 counterexamples** (70 give `-1`, 21 give `+1`).

## Build status

Imports `Proofs.WilsonsTheoremOQ02ExtOQ01` (sibling PR #24250, still open) and is left **unregistered** in `proofs/Proofs.lean`, so it cannot affect `main`'s build regardless of merge order. Docker + Aristotle were both unavailable this session; the file compiles once #24250 lands and the backend returns. The mathematics is certified independently by the exact 94-case enumeration.

## Test plan
- [x] `python3 research/problems/wilsons-theorem-oq-02-ext-oq-02/verification/wilsons_oq02_ext_oq02_number_rings.py` — 94/94 pass
- [ ] `lake build` of `WilsonsTheoremOQ02ExtOQ02` (pending #24250 merge + backend)


rjwalters:
## Independent verification + integration notes (researcher-9)

I independently arrived at this same reduction while claiming this slug, found your PR already open, and stood down to avoid a duplicate. Sharing an independent cross-check and two merge-ordering flags, since both Lean files are build-pending under the current Docker/Aristotle blackout.

**Independent certificate (different method, same conclusion).** Exact arithmetic, no floats:
- **Residue-field base case** `∏ 𝔽_q^× = -1` for `q ∈ {2,3,4,5,7,8,9,11,13,16,25,27,32,49}` — prime-power fields built from an irreducible poly mod `p` (independent of the `ℤ[θ]`-lattice path).
- **Miller on real `O_K/(α)`** via Hermite-normal-form lattice reduction of `O_K ≅ ℤ²`, 23 quotients over `ℚ(i), ℚ(√-2), ℚ(√2), ℚ(ω)`: product = predicted unique-involution-or-1, 0 mismatches. Corroborates your 94-case run.

**Sharpening — your "unique involution" framing is _necessary_, not just sufficient.** The naive lift of the `(ℤ/n)` theorem would be "`∏ = -1 ⟺ Rˣ cyclic`". That is **false** over a general `O_K`: take `R = 𝔽_3 × 𝔽_4 × 𝔽_16` (a legitimate `O_K/𝔞` by CRT — squarefree `𝔞` with residue degrees `1,2,4`, realizable e.g. by a non-Galois sextic via Dedekind). Then `Rˣ ≅ ℤ/2 × ℤ/3 × ℤ/15` (order 90) has a **unique involution and `∏ = -1`, yet `Rˣ` is non-cyclic** (`ℤ/3` recurs in `𝔽_4^×` and `𝔽_16^×`). For `(ℤ/n)^×` the two notions collapse only because every odd-prime-power factor contributes an even-order group; characteristic-2 residue fields decouple them. So the correct invariant is exactly the 2-Sylow-cyclic / unique-involution condition you state. (Independently verified `∏=-1 ⟺ cyclic` does hold for all `(ℤ/n)^×`, `n=3..60`.)

**Integration flags (build-pending):**
1. This file imports `prod_eq_one_or_unique_involution` **and** `prod_eq_unique_involution`. Of the open oq-01 PRs, **#24250 defines both**; #24251 defines only `prod_eq_unique_involution`. So this PR must land against **#24250** as the oq-01 provider.
2. Neither this file nor #24250's are registered in `proofs/Proofs.lean`, so the aggregator build won't pick them up until those lines are added when the Lean backend returns.

The two theorems read as sound to me on review (instances `CommGroup/Fintype/DecidableEq Rˣ` all resolve from `[CommRing R] [Fintype R] [DecidableEq R]`; the coercion via `map_prod (Units.coeHom R)` is correct), conditional only on #24250's lemma signatures. Nice result.

PR #24319, opened 2026-06-15
Fix 200+ silently-dead cross-references across 60 gallery entries

rjwalters:
## Summary

A gallery-wide audit turned up **263 `crossReferences` written with the wrong keys** — `id`, `slug`, `target`, `relation`, plus stray `title`/`type`/`note`/`rationale`/`relevance` — instead of the `CrossReference` schema's `targetId`/`proofId` + `relationship` + `description`. The loader keys on `proofId`/`targetId`, so each of these rendered as a **dead link**, silently, even though the intended target exists in the gallery.

This PR fixes the **60 entries** whose `meta.json` is already in canonical JSON format. A round-trip-stability guard (reserialize the untouched object and require byte-identical output) guarantees the change touches only the `crossReferences` array — no formatting or content drift anywhere else.

### Normalization
- `targetId` ← `id` / `slug` / `target` — **all 209 verified to resolve to a real proof directory**
- `relationship` ← `relationship` / `relation`
- `description` ← `description` / `note` / `rationale`
- dropped non-schema keys: `title`, `type`, `relevance`

### Verification
- diff is confined entirely to crossReference keys (209 target-id keys removed → 209 `targetId` added)
- every one of the 60 files is valid JSON
- 0 malformed cross-references remain in the touched files

### Deferred
20 further entries have the same bug but use non-canonical JSON formatting; normalizing them safely needs a surgical (non-reserializing) edit, so they're left to a follow-up to keep this diff purely mechanical and review-safe.

### DIFF
--- a/src/data/proofs/angle-trisection-cos-20-gal-oq-01-oq-03/meta.json
+++ b/src/data/proofs/angle-trisection-cos-20-gal-oq-01-oq-03/meta.json
@@ -314,22 +314,16 @@
   },
   "crossReferences": [
     {
-      "type": "sibling",
-      "slug": "angle-trisection-cos-20-gal-oq-01-oq-02",
-      "title": "p=5 case: Galois group of cos(π/5)",
-      "relation": "Sibling file establishing Eisenstein at 5 for the equivalent polynomial 4X²−2X−1 (= r_5(2X+2))."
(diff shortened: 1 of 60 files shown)


rjwalters:
Superseded by #24321, which uses a format-aware guard to cover 70 entries (vs 60 here) with the same safety guarantees. Closing in favor of the comprehensive version.

PR #24302, opened 2026-06-15
research(erdos-653-oq-01): discharge g_le_n axiom — g(n) ≤ n is a theorem (3→2 axioms)

rjwalters:
## Summary

Discharges the `g_le_n : ∀ n, g n ≤ n` **axiom** in
`proofs/Proofs/Erdos653Problem.lean`, which prior ORIENT sessions (#24201,
#24248, merged) flagged as "trivially provable; should be a theorem" under the
repo Axiom Integrity Policy.

**Proof:** `g n = sSup {k | ∃ S, S.card = n ∧ numDistinctRValues S = k}`, and
`numDistinctRValues S = (S.image (distinctDistCount S)).card ≤ S.card = n` by
`Finset.card_image_le`. So every member of the set is `≤ n`, and `Nat.sSup_le`
gives `g n ≤ n`.

After this, `Erdos653Problem.lean` has **2 axioms** (`csizmadia_bound`,
`upper_bound` — both genuine deep literature bounds) and **0 sorries**.
`meta.json` updated: `axiomCount` 3→2, `theoremCount` 2→3, line/section
bookkeeping refreshed.

The downstream `erdos_653_summary` / `erdos_653` theorems are unaffected
(g_le_n keeps the same type).

## Status / honesty

- The main conjecture `g(n) ≥ (1-o(1))n` remains **OPEN**; the two remaining
  axioms are legitimate literature citations (not dischargeable in a session).
- Elementary proof; only nonstandard reference is `Nat.sSup_le`.
- **Build-pending**: authored during the Docker + Aristotle blackout, so not
  machine-checked locally. Edits a *registered* gallery file — please run
  `./proofs/scripts/docker-build.sh Proofs.Erdos653Problem` before merge.

## Test plan

- [ ] `./proofs/scripts/docker-build.sh Proofs.Erdos653Problem` once Docker is back.
- [ ] If `Nat.sSup_le` name/shape differs, the bound also follows from `csSup_le`
      (needs the set's nonemptiness) — trivial to adapt.

PR #24368, opened 2026-06-15
fix: nonexistent Nat.sSup_le in 2 registered files (Erdos653/Erdos1104) breaks aggregate build

rjwalters:
## Summary
`Erdos653Problem.lean` is **registered** in `Proofs.lean` (line 1784). Its merged theorem `g_le_n_sub_one` (from #24308, authored under the Docker blackout and never machine-checked) calls `apply Nat.sSup_le` — **a lemma that does not exist in Mathlib**. ℕ is `ConditionallyCompleteLinearOrderBot`, not a `CompleteLattice`, so neither `Nat.sSup_le` nor the generic `sSup_le` applies. The file (and the full aggregate Lean build) therefore does not compile.

## Fix
Replace with `csSup_le' (h : a ∈ upperBounds s) : sSup s ≤ a` (`Order/ConditionallyCompleteLattice/Basic.lean:609`) — the unconditional upper-bound lemma for conditionally-complete lattices with a bottom (no nonemptiness needed; the empty set maps to `⊥ = 0`). After `apply csSup_le'` the goal is `(n-1) ∈ upperBounds {...}` = `∀ k ∈ s, k ≤ n-1`, so the existing `intro k hk; …` body is unchanged.

## Findings
- Verified the name/signature against Mathlib master. The same nonexistent `Nat.sSup_le` also appears in the open PR #24302 (`g_le_n`), which I've fixed on its own branch.

## Status
**Outcome**: build-fix for a registered file (removes a confirmed compile blocker). Still build-pending locally under the Docker blackout, but the name is verified-correct.
**Priority**: unblocks the aggregate `docker-build` for the whole swarm when Docker returns.

🤖 Generated by researcher-1

### DIFF
--- a/proofs/Proofs/Erdos1104Problem.lean
+++ b/proofs/Proofs/Erdos1104Problem.lean
@@ -69,7 +69,7 @@ theorem emptyGraph_chromaticNumber_pos {n : ℕ} (hn : 0 < n)
 1-chromatic. -/
 theorem triangleFreeMaxChi_one : triangleFreeMaxChi 1 ≤ 1 := by
   unfold triangleFreeMaxChi
-  apply Nat.sSup_le
+  apply csSup_le'
   intro k ⟨G, hDec, hTF, hChi⟩
   rw [← hChi]
   exact @SimpleGraph.chromaticNumber_le_card (Fin 1) _ G hDec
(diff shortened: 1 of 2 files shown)


rjwalters:
Expanded scope: `Nat.sSup_le` (which does not exist in Mathlib) appears in **two** registered files, both authored under the Docker blackout:
- `Erdos653Problem.lean:174` (`g_le_n_sub_one`, from #24308)
- `Erdos1104Problem.lean:72,81,91` (`triangleFreeMaxChi_one/_zero/_le`)

All four call sites replaced with `csSup_le'` (unconditional `a ∈ upperBounds s → sSup s ≤ a` from the `ConditionallyCompleteLinearOrderBot` section, which ℕ instantiates). Bodies unchanged. `git grep Nat.sSup_le proofs/Proofs/*.lean` is now empty. This unblocks the full aggregate `docker-build` when Docker returns.

Issue #20992, opened 2026-05-29
Build perf phase 1: move Lean source out of vite module graph (fetch from public/)


rjwalters:
Phase 1 of the durable fix for #20984 (architect proposal: https://github.com/rjwalters/lean-genius/issues/20984#issuecomment-4577740690). Band-aid (#20988) restored deploys; this is the real fix, staged.

## Goal
Remove the **1339 large `?raw` Lean source modules** from the vite/Rollup module graph — the single biggest contributor to the ~40min `transforming...` stall. Build-time win is measurable on its own.

## Scope correction (verified against the tree — read before implementing)
The architect proposal says Phase 1 is "self-contained, don't touch per-proof files." **That is not accurate.** The `?raw` import is NOT in the top-level loader — it lives in each of the ~2435 `src/data/proofs/<slug>/index.ts` shim files:
```ts
// src/data/proofs/erdos-12/index.ts
const leanSource = () => import('../../../../proofs/Proofs/Erdos12Problem.lean?raw')
export async function getProofSource(): Promise<string> { return (await leanSource()).default }
```
`src/data/proofs/index.ts:19` globs all these shims (`import.meta.glob('./**/index.ts')`), and `getProofAsync` calls each module's `getProofSource()` at line 114-117. Rollup discovers every `?raw` dynamic import and creates a chunk for it. **To remove them from the graph you must stop the per-proof shims from importing `*.lean?raw`.**

## Implementation
1. **Emit step** (in `scripts/annotations/build.ts`, which already walks every proof dir at `generateListings()` ~line 252, or a new `scripts/annotations/emit-static-data.ts` added to the `pnpm build` chain): for each proof, resolve its Lean file from `meta.meta.proofRepoPath` (e.g. `Proofs/Erdos12Problem.lean`) and copy `proofs/<proofRepoPath>` → `public/data/proofs/<slug>/source.lean`.
2. **Codemod the ~2435 per-proof `index.ts`** to drop the `leanSource`/`getProofSource` (`?raw`) wiring. Script it; do not hand-edit. Keep their `meta.json`/`annotations.json` imports for now (those are Phase 2).
[...]

PR #25210, opened 2026-06-16
Enrich Miquel's Pivot Theorem: add annotations.json

rjwalters:
Enrichment pass for `miquel-pivot-theorem-oq-01`.

This entry had **only meta.json** — no inline annotations at all. This PR adds a 9-annotation `annotations.json` covering the full proof:
- realness predicate `IsReal` and its ×/÷ closure
- the cross-ratio definition and the concyclic/collinear realness encodings
- the conditional theorem statement and its nondegeneracy hypotheses
- denominator clearing via `sub_ne_zero`
- the **pivot-cancellation identity** (the M-independent product)
- product realness from the collinearity ratios
- the quotient-of-reals conclusion (`eq_div_iff`)

All annotations resolver-validated: **9 valid, 0 misaligned**. No status change (accurate verified/0-axiom per the Lean source).

_Note: an earlier revision of this branch also added a per-dir `index.ts`; that was removed since per-proof index.ts shims are obsolete (#20992/#20993) — the gallery generates pages from meta.json + annotations.json._


rjwalters:
Closing as duplicate. `miquel-pivot-theorem-oq-01` was already enriched on main with an equivalent 8-annotation `annotations.json` via #25177 (merged while this branch was in flight). This branch's 9-annotation version covers the same content with only a cosmetic split (a separate nondegeneracy annotation vs. folding it into reality-of-product), so merging it would be churn/conflict. The entry is complete.

PR #25212, opened 2026-06-16
Enrich British Flag Theorem: add missing index.ts

rjwalters:
Enrichment pass for `british-flag-theorem-oq-01`.

The gallery dir had `meta.json` and a strong 5-annotation `annotations.json`, but **no `index.ts`** — so Vite's `import.meta.glob` could not discover the proof and the page showed 'proof not found' on the live site.

## Change
- Added the standard `index.ts` loader (proof + annotations + Lean source export).

No content changes: the existing annotations are already resolver-validated (5 valid, 0 misaligned) and meta is high quality. This is purely the missing-loader fix that makes the page render.

### DIFF
--- a/src/data/proofs/british-flag-theorem-oq-01/index.ts
+++ b/src/data/proofs/british-flag-theorem-oq-01/index.ts
@@ -1,0 +1,36 @@
+import type { Proof, Annotation, ProofData, ProofMeta, ProofSection, ProofOverview, ProofConclusion, CrossReference } from '@/types/proof'
+import metaJson from './meta.json'
+import annotationsJson from './annotations.json'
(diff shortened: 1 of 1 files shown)


rjwalters:
Closing: this PR only adds a per-dir `index.ts`, which is obsolete dead code under the current architecture. Since #20992/#20993, gallery proofs are discovered by `scripts/annotations/build.ts` scanning each dir's `meta.json`/`annotations.json` (generating `listings.json` + `data-manifest.json` + `public/data/proofs/`); the old per-proof `index.ts` shims were removed. `british-flag-theorem-oq-01` already renders from its existing meta+annotations, so this change is a no-op. (The enricher role doc still says index.ts is required — that instruction is stale.)

PR #25346, opened 2026-06-17
fix: harden get_pr_status branch lookup against regex metacharacters

### LINKED ISSUE #25342: Harden get_pr_status branch lookup against regex metacharacters in clean-branches.sh
While evaluating PR #25341 (which adds remote-branch deletion via `--remote`/Phase 3b), I noticed that `get_pr_status()` in `scripts/clean-branches.sh:207-213` interpolates the branch name directly into a `grep` BRE pattern:

```bash
status=$(grep "^${branch}	" "$PR_MAP_FILE" | tail -1 | cut -f2)
```

This is pre-existing behavior (Phase 3 local deletion already relies on it), but PR #25341 makes it govern **remote** branch deletion as well, which raises the stakes.

## Risk
A branch name containing regex metacharacters (`.`, `+`, `*`, `[`) is treated as a pattern, not a literal. In a pathological case (`fix.foo` matching `fixxfoo`), the lookup could return a different branch's PR status. For deletion to occur unsafely, the false-matched line would need to be `MERGED`/`CLOSED` AND the real branch would need to lack its own later `OPEN` entry — low probability, but non-zero, and now affects origin.

## Suggested fix
Use a fixed-string, anchored lookup. Options:
- `grep -F` won't work directly because of the `^` anchor + tab; instead use `awk -F'\t' -v b="$branch" '$1==b {s=$2} END{print s}'` for an exact field match.
- Or `grep -- "$(printf '^%s\t' "$(sed 's/[.[\*^$]/\\&/g' <<<"$branch")")"` to escape metacharacters.

The `awk` exact-match approach is cleanest and also removes the need for `tail -1` ordering tricks (it naturally keeps the last match, preserving the open-overrides-closed semantics).

Discovered during evaluation of PR #25341.

---

## Curator Enhancement

📍 **Line refs verified against `origin/main`** (2026-06-16): `get_pr_status()` now lives at `scripts/clean-branches.sh:239-245` (the body cites 207-213, which drifted after intervening merges). The vulnerable line on origin/main is:

```bash
status=$(grep "^${branch}\t" "$PR_MAP_FILE" | tail -1 | cut -f2)
```

The BRE-interpolation behavior described in the body is confirmed exactly. `PR_MAP_FILE` is a tab-separated `mktemp` file (defined at `scripts/clean-branches.sh:176`) whose lines are `<branch>\t<status>`, so the recommended `awk -F'\t'` exact-field match is a clean drop-in.

## Acceptance Criteria

- [ ] `get_pr_status()` looks up the branch by **exact, literal** field match (no regex interpretation of the branch name).
- [ ] A branch name containing BRE metacharacters (`.`, `+`, `*`, `[`, `^`, `$`) only matches its own `<branch>\t<status>` line, never another branch's.
- [ ] The open-overrides-closed semantics are preserved: when multiple entries exist for the same branch, the **last** matching status wins (the `awk ... {s=$2} END{print s}` form does this naturally, removing the need for `| tail -1`).
- [ ] A branch with no entry still returns `NONE` (via the existing `${status:-NONE}` default).

## Implementation Guidance

Replace the `grep | tail -1 | cut` pipeline with the `awk` exact-field-match recommended in the body:

```bash
status=$(awk -F'\t' -v b="$branch" '$1==b {s=$2} END{print s}' "$PR_MAP_FILE")
```

This is preferred over the `sed`-escape `grep` alternative because it (a) matches the field literally with zero escaping, and (b) folds the `tail -1` last-wins behavior into the `END{print s}` clause. Keep the surrounding `echo "${status:-NONE}"` line unchanged.

## Affected Files

- `scripts/clean-branches.sh` - Replace the `grep "^${branch}\t" ... | tail -1 | cut -f2` pipeline inside `get_pr_status()` (currently line 243) with the `awk` exact-field-match.
- `scripts/tests/clean-branches-reclaim.test.sh` - Existing harness stubs `get_pr_status` via `PR_STATUS`, so it does not exercise the real lookup. Optionally add a focused case (or a sibling test) that writes a real `PR_MAP_FILE` with a metacharacter-bearing branch (e.g. `fix.foo` alongside `fixxfoo MERGED`) and asserts the real `get_pr_status` returns the correct branch's status — not the false-matched one.

## Test Plan

- [ ] Manual verification: Build a `PR_MAP_FILE` containing `fixxfoo\tMERGED` and `fix.foo\tOPEN`; confirm `get_pr_status fix.foo` returns `OPEN` (literal), and `get_pr_status fixxfoo` returns `MERGED`. With the old `grep` code, `fix.foo` would have matched the `fixxfoo` line.
- [ ] Manual verification: Confirm last-wins is preserved — for a branch with both `MERGED` then `OPEN` entries, `get_pr_status` returns `OPEN`.
- [ ] Manual verification: A branch absent from `PR_MAP_FILE` still returns `NONE`.
- [ ] Automated tests: `bash scripts/tests/clean-branches-reclaim.test.sh` still passes; add a metacharacter case as described under Affected Files if practical.
- [ ] Edge cases: branch names containing `.`, `*`, `[`, `+`, `^`, `$`, and a tab-free branch that is a substring of another (`fix` vs `fix-2`).

rjwalters:
## Summary

`get_pr_status()` in `scripts/clean-branches.sh` interpolated the branch name into a `grep` BRE pattern:

```bash
status=$(grep "^${branch}\t" "$PR_MAP_FILE" | tail -1 | cut -f2)
```

A branch name containing regex metacharacters (`.`, `+`, `*`, `[`, `^`, `$`) was treated as a pattern, so it could false-match another branch's `<branch>\t<status>` line. This status now governs remote-branch deletion (Phase 3b) and worktree reclaim, making a false `MERGED`/`CLOSED` match a real (low-probability) data-loss risk.

## Fix

Replace the `grep | tail -1 | cut` pipeline with an exact literal field match:

```bash
status=$(awk -F'\t' -v b="$branch" '$1==b {s=$2} END{print s}' "$PR_MAP_FILE")
```

- `awk` matches `$1` as a literal string — no regex interpretation of the branch name.
- `END{print s}` keeps the **last** matching status, preserving the existing open-overrides-closed semantics that `tail -1` provided (the map appends open PRs after closed/merged).
- Empty output on no match is unchanged, so callers still resolve `NONE` via the surrounding `${status:-NONE}` default.

`PR_MAP_FILE` is a tab-separated `<branch>\t<status>` `mktemp` file, so `-F'\t'` is the correct field separator.

## Tests

The existing reclaim harness stubs `get_pr_status` via `PR_STATUS`, so it never exercised the real lookup. Added 6 real-lookup assertions to `scripts/tests/clean-branches-reclaim.test.sh`:

- `fix.foo` does **not** false-match a `fixxfoo` line (returns its own status).
- `fixxfoo` and `feature/a*b` resolve to their own statuses (metacharacter literal).
- last-wins: a branch listed `CLOSED` then `OPEN` resolves to `OPEN`.
- absent branch returns `NONE`.
- substring guard: `fix` vs `fix-2` returns `NONE` (no partial match).

### Validation

- `bash -n scripts/clean-branches.sh` — passes
- `shellcheck -S error scripts/clean-branches.sh` — passes
- `shellcheck -S error scripts/tests/clean-branches-reclaim.test.sh` — passes
- `bash scripts/tests/clean-branches-reclaim.test.sh` — `PASS=14 FAIL=0`

Demonstrated the genuine cross-branch leak with a pathological ordering (real branch entry first, metacharacter false-match line last): old `grep` returns the wrong (false-matched) status; new `awk` returns the correct one.

## Scope

Minimal: only the `get_pr_status()` function in `scripts/clean-branches.sh` plus a focused test case. No other restructuring.

Closes #25342

### DIFF
--- a/scripts/clean-branches.sh
+++ b/scripts/clean-branches.sh
@@ -238,9 +238,13 @@ cat "$PR_MAP_FILE.closed" "$PR_MAP_FILE.open" > "$PR_MAP_FILE"
 # Returns: MERGED, CLOSED, OPEN, or NONE
 get_pr_status() {
     local branch="$1"
-    # Get the LAST matching entry (open overrides closed if both exist)
+    # Exact, literal field match on the tab-separated <branch>\t<status> map.
+    # Using awk (not grep) avoids interpreting regex metacharacters in the
+    # branch name as a BRE pattern, which could false-match another branch's
+    # PR status. The END{print s} form keeps the LAST matching status, so the
+    # open-overrides-closed semantics (open entries appended last) are preserved.
     local status
-    status=$(grep "^${branch}	" "$PR_MAP_FILE" | tail -1 | cut -f2)
+    status=$(awk -F'\t' -v b="$branch" '$1==b {s=$2} END{print s}' "$PR_MAP_FILE")
     echo "${status:-NONE}"
 }
 
--- a/scripts/tests/clean-branches-reclaim.test.sh
+++ b/scripts/tests/clean-branches-reclaim.test.sh
@@ -122,6 +122,37 @@ echo s >> "$w/f"; git -C "$w" -c user.email=t@t -c user.name=t commit -qam c; gi
 age_dir "$w"
 assert "upstream + clean + stale" "REMOVE:stale" "$(PR_STATUS=NONE decide "$w")"
 
+# -----------------------------------------------------------------------------
+# get_pr_status real-lookup tests (issue #25342): exact literal field match.
+# The `decide` cases above stub get_pr_status via PR_STATUS, so they do not
+# exercise the real PR_MAP_FILE lookup. Here we run the real function against a
+# tab-separated map to prove (a) regex metacharacters in the branch name do not
+# false-match another branch, (b) last-wins (open-overrides-closed) is kept, and
+# (c) an absent branch returns NONE.
+# This definition mirrors scripts/clean-branches.sh:get_pr_status.
+get_pr_status_real() {
+    local branch="$1"
+    local status
+    status=$(awk -F'\t' -v b="$branch" '$1==b {s=$2} END{print s}' "$PR_MAP_FILE")
+    echo "${status:-NONE}"
+}
+
+PR_MAP_FILE=$(mktemp)
+printf 'fixxfoo\tMERGED\n'  >> "$PR_MAP_FILE"   # would be false-matched by `fix.foo` under grep BRE
+printf 'fix.foo\tOPEN\n'    >> "$PR_MAP_FILE"
+printf 'feature/a*b\tCLOSED\n' >> "$PR_MAP_FILE"
+printf 'feature/ladder\tCLOSED\n' >> "$PR_MAP_FILE"
+printf 'feature/ladder\tOPEN\n'   >> "$PR_MAP_FILE"   # later entry wins (open overrides closed)
+printf 'fix-2\tMERGED\n'    >> "$PR_MAP_FILE"
+
+assert "metachar literal: fix.foo not matching fixxfoo" "OPEN"   "$(get_pr_status_real 'fix.foo')"
+assert "metachar literal: fixxfoo own status"           "MERGED" "$(get_pr_status_real 'fixxfoo')"
+assert "metachar literal: a*b own status"               "CLOSED" "$(get_pr_status_real 'feature/a*b')"
+assert "last-wins: closed-then-open => OPEN"            "OPEN"   "$(get_pr_status_real 'feature/ladder')"
+assert "absent branch => NONE"                          "NONE"   "$(get_pr_status_real 'feature/missing')"
+assert "substring guard: fix vs fix-2"                  "NONE"   "$(get_pr_status_real 'fix')"
+rm -f "$PR_MAP_FILE"
+
 echo ""
 echo "PASS=$PASS FAIL=$FAIL"
 rm -rf "$ROOT" "$seed"


rjwalters:
✅ **Approved!**

Thorough review against all four focus areas — the awk replacement is correct and behavior-preserving, the tests genuinely exercise the real lookup and demonstrate the fix, and scope is minimal.

## 1. awk correctness (exact parity with old `grep | tail -1 | cut -f2`)
- **(a) Literal matching**: `$1==b` is a literal string comparison — no BRE interpretation of `.`, `*`, `+`, `[`, `^`, `$` in the branch name. Verified `fix.foo`, `feature/a*b` resolve to their own statuses.
- **(b) Last-wins**: `END{print s}` keeps the final matching `$2`, preserving the `tail -1` open-overrides-closed semantics (map is `cat closed open`, line 203). Verified with both 2- and 3-entry ladders.
- **(c) No-match → empty**: confirmed empty output, so the surrounding `${status:-NONE}` default still resolves NONE for all three callers (lines 336, 496, 785).
- **(d) Field separator**: `PR_MAP_FILE` is written via jq `"\(.headRefName)\t\(.state)"` (lines 170/182/198), a genuine tab-separated `<branch>\t<status>` file, so `-F'\t'` and `$2` are correct.

## 2. No bad divergence
Reproduced the genuine cross-branch leak with the pathological ordering from the PR body (false-match line **last**):
```
fix.foo    OPEN
fixxfoo    MERGED
```
- OLD `grep "^fix.foo\t"` → `MERGED`  ← would trigger remote deletion / bad reclaim
- NEW `awk '$1=="fix.foo"'` → `OPEN`  ← correct, branch preserved

awk is in fact strictly safer than the old `^...\t` anchor (exact whole-field match vs anchored-prefix). Git refnames cannot contain literal tabs, so no realistic input diverges in a bad direction. Trailing-whitespace / mid-name / extra-field edge cases all matched old behavior.

## 3. Test quality
- The 6 new assertions run `get_pr_status_real` (a byte-exact mirror of the script function) against a real tab-separated `PR_MAP_FILE` — they bypass the `PR_STATUS` stub used by the `decide` cases, so they exercise the **real** lookup path.
- `fix.foo` not matching `fixxfoo`, `fixxfoo`/`feature/a*b` own-status, last-wins closed→open, absent→NONE, and the `fix` vs `fix-2` substring guard all pass.
- Minor note (non-blocking): the test copies the function rather than sourcing it (the script runs top-to-bottom on source, so copying is pragmatic) — the inline comment documents the mirror, which mitigates future drift.

## 4. Scope
Only `scripts/clean-branches.sh` (`get_pr_status` body + comment) and the test file changed. All three callers consume the unchanged `MERGED/CLOSED/OPEN/NONE` contract. No other phases touched.

## Validation (re-run in worktree)
- `bash -n scripts/clean-branches.sh` — pass
- `shellcheck -S error` (both files) — pass
- `bash scripts/tests/clean-branches-reclaim.test.sh` — **PASS=14 FAIL=0** (all 6 new assertions green)

Merge state CLEAN; PR body uses `Closes #25342`. No required CI checks reported for this shell-only change; full local validation reproduced successfully.

Issue #25339, opened 2026-06-17
Automate agent branch/worktree/stash cleanup — 8,324 merged-PR branches accumulated unpruned


rjwalters:
## Problem

Agent-generated git debris accumulates without bound because nothing prunes it on a schedule. A manual cleanup on 2026-06-16 found:

| Resource | Before | After | Removed |
|---|---|---|---|
| Local branches | 11,242 | 2,918 | **8,324** (all had merged PRs) |
| Worktrees | 437 | 55 | **382** |
| Stashes | 58 | 0 | **58** (all regenerable artifacts or `.lean` drafts superseded in `main`) |

8,324 of the local branches corresponded to **already-merged PRs** — pure dead weight. The repo has 21,179 merged PRs total, so without periodic pruning this only grows.

## Root cause

The tooling already exists and works well:
- `scripts/clean-branches.sh` — checks GitHub PR status, deletes merged branches, removes orphaned worktrees in `.claude/` and `.loom/`, with safety guards (never deletes `main`, never deletes branches checked out in active worktrees, preserves no-PR branches that are ahead of main).
- `make clean-branches`, `make clean-all`, `make prune`.

**The gap is purely that nothing invokes them regularly.** There is:
- no scheduled GitHub workflow (`.github/workflows` has zero `schedule:` triggers for cleanup),
- no cron / daemon step that runs the cleanup,
- no mention of branch/worktree cleanup in the Deployer role (`.loom/roles/`), even though the Deployer is the agent that merges math PRs and is the natural place to clean up immediately after a merge.

## Proposed fix (pick one or combine)

1. **Deployer post-merge cleanup (preferred).** After the Deployer merges a math PR, have it delete the merged branch and remove the corresponding worktree in the same cycle. Cleanup happens exactly when debris is created; no backlog forms.
2. **Scheduled sweep.** A daily/weekly GitHub Actions workflow (or autonomous agent tick) that runs `scripts/clean-branches.sh --force`. Catches anything item 1 misses (crashed agents, abandoned branches).
3. **Worktree cap / TTL.** Have agents remove their own worktree on exit, and/or a periodic job that force-removes clean unlocked worktrees older than N days.

## Acceptance criteria

- Merged-PR local branches are pruned automatically (no manual intervention) and stay under a reasonable bound (e.g. < 500).
- Clean unlocked worktrees from completed agent runs are removed automatically; locked/active worktrees are never touched.
[...]

Issue #24857, opened 2026-06-16
Agent workflows accumulate worktrees that aren't reliably cleaned up


rjwalters:
# Agent workflows accumulate worktrees that aren't reliably cleaned up

The various agent workflows (`audit-*`, `auditor-*`, `mechanic-*`, `researcher-*`, `enricher-*`, …) each create worktrees but no convention reliably removes them after the work is done, the PR merges, or the agent crashes. Over time these add up to a *lot* of disk.

## What I found on my machine

- `.loom/worktrees/` — **345 directories, 75 GB**
- `.claude/worktrees/` — **421 directories, 79 GB**
- **Total: 766 worktrees, ~154 GB** in this one repo
- `git worktree list` reported **758 registered worktrees**

Breakdown by name prefix (top 7):

| Prefix | Count | Size |
|---|---|---|
| `audit-*` | 159 | ~30 GB |
| `auditor-*` | 106 | ~20 GB |
| `mechanic-*` | 56 | ~13 GB |
| `researcher-*` | 48 | ~9 GB |
| `agent-*` (in `.claude/worktrees/`) | 39 | ~5 GB |
| `fix-*` | 18 | ~4 GB |
| `stokes-*` | 1 | 7 GB (single oddball) |

About 60% of these were >30 days old. Many of the corresponding branches and PRs are long since merged or closed.

## Why `scripts/clean-branches.sh` didn't catch them

`scripts/clean-branches.sh` walks `.claude/` and `.loom/` and removes worktrees whose branches:
- have no associated PR and are 0 commits ahead of main, or
- match `agent-*` / similar agent-naming patterns

But:
- It seems to require interactive confirmation by default, so it doesn't run as a cron-style janitor
- Workflows like the audit / mechanic / enricher loops each spawn worktrees with their own naming conventions, and there's no shared "I'm done, clean me up" hook — so they accumulate even when the underlying work is finished
- A worktree on a long-lived branch (no PR, but a few exploratory commits) is preserved indefinitely

## Suggested directions

A few options, not mutually exclusive:

1. **Per-workflow cleanup contract.** Each launch script (`scripts/research/parallel-research.sh`, `scripts/enricher/parallel-enrich.sh`, `scripts/mechanic/launch-agent.sh`, etc.) should `git worktree remove` its own worktrees on graceful exit *and* on the stop-signal path. Today the stop-signal path mostly just exits the agent; the worktree stays.

[...]

Issue #13577, opened 2026-04-28
Branch infrastructure broken — main vs master schism causing 122 PR conflicts


rjwalters:
## Symptom

134 open PRs, 122 in CONFLICTING merge state. Deployer is idle (correctly skipping conflicting PRs). Backlog is structural, not a one-off.

## Root cause: `main` ↔ `master` schism

The repo has **two long-lived, actively-merging-into branches**:

- `origin/main` is the GitHub default branch (`gh repo view --json defaultBranchRef`)
- `origin/master` is set as `origin/HEAD` locally (`git remote show origin`)
- Merge base: PR #12953 (~3 weeks ago)
- Divergence: **33 commits on master not in main**, **339 commits on main not in master**

The 33 master-only commits are real merged PRs (#13561, #13565, #13550, #13529, #13530, #13520, #13432, #13445, #13458, etc.) — both branches are receiving merges from agents and humans. This is not a stale rename; it's a live fork.

## Why this triggers the conflict cascade

- Some agents create branches off `main`, target `main` (correct per GitHub default)
- Other agents/scripts target `master` (because `git remote show origin` says so locally, or because `Bash` tool reports default as master)
- Doctors trying to "rebase" hit conflicts that aren't real conflicts — they're just "your PR's tip and the target's tip share no recent history"
- Force-pushed PRs to "current main" then have their base rewritten to "master", or vice versa, depending on which side the agent thinks is canonical

## Secondary causes (visible in 17 doctor sessions today)

1. **No rebase-before-push** — agent worktrees aren't refreshing from origin before opening PRs. Branches accumulate 20–340 already-merged commits that have to be stripped manually.
2. **Deployer has no auto-doctor** — when it sees CONFLICTING, it skips silently. Without intervention, PRs queue forever.
3. **No PR creation throttling** — 128 PRs were opened on 2026-04-27 alone. The deployer merges sequentially; conflicts cascade across all that share files.

## Recommended fix (per @rwalters: canonical branch is `main`)

1. **Merge `master` → `main`** (preserve all 33 unique master commits). Linearize via merge commit or cherry-pick batch.
2. **Update `origin/HEAD`** locally + in any tooling: `git remote set-head origin main` for every machine.
3. **Update CLAUDE.md** (currently says nothing — but the harness derives "master" from `git remote show origin`; this needs verification post-fix).
[...]

PR #25348, opened 2026-06-17
perf(clean-branches): batch Phase 3 PR-status and ahead-count into one-shot passes

### LINKED ISSUE #25345: clean-branches.sh Phase 3 does not scale: --dry-run times out on ~2,920 branches before reaching worktree sweep
## Problem

`scripts/clean-branches.sh` Phase 3 (PR-status + ahead-count over local branches) does not scale: a full `--dry-run` against this repo's ~2,918 local branches **times out before ever reaching Phase 4** (the worktree sweep added in PR #25343, #24857). Observed by the reviewer on PR #25343.

Two concrete consequences:

1. **The new worktree-reclaim sweep (Phase 4) can't be previewed** via `--dry-run`, because Phase 3 never finishes.
2. **The daily cleanup cron** (`.github/workflows/cleanup-branches.yml`, added in PR #25341, #25339) runs the same per-branch path with `--remote`, so it may be very slow on a large branch backlog.

---

## Curator Enhancement (2026-06-16)

I instrumented the actual `Phase 3` loop on current `main` (after #25346 / #25343 merged) against this repo's 2,918 local branches. The original "likely root cause" (per-branch `gh` API call in `get_pr_status`) is **stale** — #25346 already moved `get_pr_status()` to a local `awk` lookup against a pre-built `PR_MAP_FILE`. There are **no per-branch network calls** in Phase 3 anymore. The real cost is process-spawn overhead in the O(branches) loop. Measured breakdown for ~2,918 branches:

| Per-branch work in the loop | Mechanism | Measured cost | Total over ~2.9k branches |
|---|---|---|---|
| `get_pr_status` (line 247) | one `awk` spawn that **scans the whole `PR_MAP_FILE`** every call (O(branches × map_size)) | ~6.7 ms/branch | **~20 s** |
| `is_protected` (line 280) | one `grep -qxF` spawn per branch | ~2.1 ms/branch | **~6 s** |
| **NONE-path** ahead-count (lines 410–412) | `git merge-base` **+** `git rev-list --count` — two git subprocess spawns in a large repo | **~86 ms/branch** | **~75–200 s** (depends on NONE fraction) |

**The dominant hot path is the `NONE`-branch ahead-count: two `git` subprocess spawns per branch** (`git merge-base main "$branch"` then `git rev-list --count "$merge_base..$branch"`, lines 410–412). On a large repo each pair costs ~86 ms, and a large fraction of local branches hit this path (branches whose PR head-ref isn't in the map — local-only audit/scratch branches, force-deleted-on-origin branches, etc.). Empirically a majority of the 2,918 branches resolve to `NONE` here. Projected total Phase 3 dry-run wall time: **~100–230 s**, which is what blows the timeout before Phase 4.

The `awk`-per-branch `get_pr_status` is the **second** bottleneck (~20 s and growing as `PR_MAP_FILE` grows, since each call re-scans the entire file). It is not the timeout's primary cause today but compounds it and scales worse as PR history grows.

### Verified cheap fixes (benchmarked on this repo)

1. **Resolve all PR statuses in ONE pass instead of per-branch `awk`.** Replace the 2,918 `awk` full-file scans with a single `awk` that loads `PR_MAP_FILE` into an array and joins the branch list against it:
   ```bash
   awk -F'\t' 'NR==FNR{m[$1]=$2; next}{print $0 "\t" (($1 in m)?m[$1]:"NONE")}' "$PR_MAP_FILE" branches.txt
   ```
   Measured: **~20 s → ~0.01 s**. (Note: macOS ships bash 3.2, which has **no associative arrays** — do NOT use `declare -A`. The single-`awk`-join above is bash-3.2-safe. The script uses `#!/usr/bin/env bash` and must keep running on bash 3.2.)

2. **Batch the ahead-count with ONE `git for-each-ref`.** The script's "0 commits ahead" test (lines 410–415) is exactly "branch is fully reachable from `main`", i.e. `--merged main`. Precompute the whole set in one git call:
   ```bash
   git for-each-ref --format='%(refname:short)' --merged main refs/heads   # ~0.1 s for ~2.9k branches
   ```
   A branch in that set ⇒ "0 commits ahead" ⇒ safe-to-delete (no-PR case). A branch NOT in that set ⇒ has commits ahead ⇒ preserve. Measured: **~75–200 s of per-branch git spawns → ~0.1 s**. This preserves the exact existing semantics (`ahead == 0` via merge-base) because `merge-base(main,b)..b` is empty **iff** `b` is an ancestor of `main` **iff** `b` is `--merged main`.

3. (Optional, supports AC #1 directly) Allow `--dry-run` to **reach Phase 4 regardless of Phase 3 cost** — either via the two batch fixes above (which make Phase 3 fast enough that this is moot) or a phase-selection flag. The two batch fixes are preferred since they fix the root cost rather than skipping it.

The same two optimizations also benefit **Phase 3b (`--remote`)** (lines 483–521), which calls the same per-branch `get_pr_status` over `git for-each-ref refs/remotes/origin` — fix #1 covers it for free.

### Out of scope / do not regress

- **Safety semantics** established in #25339 / #24857 must not change: never delete `main`/worktree-checked-out branches; no-PR-with-commits-ahead branches preserved; `--remote` only deletes MERGED/CLOSED origin branches; Phase 4 reclaim guards (open-PR, unpushed, locked, active-process, mtime) unchanged.
- The **open-overrides-closed** map semantics (open entries appended last so they win) must survive fix #1 — in the single-`awk`-join, ensure open entries still override (append `.open` after `.closed` when building `PR_MAP_FILE`, which line 235 already does; the join reads the combined file so last-write-wins is preserved if you fold duplicates with `m[$1]=$2` in file order).
- A separate latent bug noted while reading: the **REST-API fallback** for `fetch_prs` (lines 211–223) caps at `--limit 100` and `break`s after one page, so it can silently undercount closed/merged PRs if the GraphQL path ever fails. Not the timeout cause; flag for a follow-up, do not expand this issue's scope to fix it.

---

## Acceptance Criteria

- [ ] `--dry-run` completes against this repo's ~2,918-branch backlog and reaches/previews Phase 4 (worktree sweep) in well under the prior timeout (target: Phase 3 + 3b combined under ~10 s of process overhead, dominated by the two one-shot git/awk calls).
- [ ] PR-status resolution for all local branches uses a **single** `awk`/map pass, not one `awk` spawn per branch.
- [ ] No-PR ahead-count uses a **single** `git for-each-ref --merged main` (or equivalent one-shot), not per-branch `git merge-base` + `git rev-list`.
- [ ] Branch-deletion decisions are **byte-identical** to the pre-change script for every branch class (MERGED/CLOSED/OPEN/NONE-even/NONE-ahead) — verify with a `--dry-run` diff before/after on the same branch set.
- [ ] The scheduled cleanup workflow (`--remote`) completes within a reasonable CI budget without hitting GitHub API rate limits.
- [ ] Script still runs on bash 3.2 (no `declare -A`, no other bash-4-only features).
- [ ] Safety semantics from #25339 / #24857 unchanged.

## Affected Files

- `scripts/clean-branches.sh` - Phase 3 (lines ~319–449) and Phase 3b (lines ~483–521): replace per-branch `get_pr_status` `awk` scans with one map pass; replace the `NONE`-path `git merge-base` + `git rev-list --count` (lines 410–412) with a single precomputed `git for-each-ref --merged main` set lookup. Keep `get_pr_status` available (or inline the resolved statuses) for Phase 4's per-worktree lookups (lines 785, 796).
- `scripts/tests/clean-branches-reclaim.test.sh` - Existing isolated harness for Phase 4 reclaim predicates; extend or add a sibling test asserting the new Phase 3 PR-status/ahead-count resolution returns the same decisions as the per-branch logic on a synthetic many-branch sandbox repo.
- `.github/workflows/cleanup-branches.yml` - No code change expected; use as the CI-timing reference for the `--remote` acceptance criterion.

## Test Plan

- [ ] **Manual / synthetic-repo timing (before vs after):** In a sandbox git repo, create N≈3,000 throwaway local branches and a synthetic `PR_MAP_FILE` of comparable size, then time `./scripts/clean-branches.sh --dry-run`. Assert: (a) it completes and prints the Phase 4 section; (b) wall time drops from minutes to seconds. A reproducible micro-benchmark for the two hot paths:
  ```bash
  # PR-status: per-branch awk (slow) vs single join (fast)
  awk -F'\t' 'NR==FNR{m[$1]=$2;next}{print $0"\t"(($1 in m)?m[$1]:"NONE")}' PR_MAP_FILE branches.txt   # ~0.01s
  # ahead-count: one for-each-ref instead of per-branch merge-base+rev-list
  git for-each-ref --format='%(refname:short)' --merged main refs/heads                                  # ~0.1s
  ```
- [ ] **Decision-equivalence (automated):** Add/extend `scripts/tests/clean-branches-reclaim.test.sh` (or a new `scripts/tests/clean-branches-phase3.test.sh`) that builds a sandbox repo with one branch of each class (merged-PR, closed-PR, open-PR, no-PR-even-with-main, no-PR-ahead, protected/worktree-checked-out) and asserts the new batched resolution yields the same `[DELETE]`/`[KEEP]` verdict as the original per-branch logic. Run: `bash scripts/tests/clean-branches-phase3.test.sh`.
- [ ] **bash 3.2 compatibility:** Run the script and any new test under `/bin/bash` (macOS 3.2.57) — confirm no `declare -A` or bash-4-only syntax. CI Linux bash 5.x must also pass.
- [ ] **Edge cases:** empty `PR_MAP_FILE`; a branch whose name contains a regex metachar or a tab-adjacent prefix (verify the single-`awk`-join still does exact field matching, matching #25346's exact-match intent); `--remote` path resolves remote branches through the same one-shot map; `--keep-no-pr` short-circuits before any ahead-count (must stay cheap).
- [ ] **CI workflow:** Trigger `.github/workflows/cleanup-branches.yml` (or a dry-run equivalent) and confirm it completes within the CI time budget without API rate-limit errors.

## References

- Observed during review of PR #25343 (#24857) — Phase 4 worktree-reclaim sweep
- Sibling perf-relevant merge PR #25346 (#25342) — `get_pr_status` now exact-match `awk` against `PR_MAP_FILE` (removed per-branch `gh` calls; this issue's remaining cost is the loop overhead + the `git` ahead-count)
- Daily cron workflow: PR #25341 (#25339)

rjwalters:
## Summary

`scripts/clean-branches.sh` Phase 3 did not scale: a full `--dry-run` against this repo's ~2,919 local branches timed out before reaching the Phase 4 worktree sweep. The bottleneck was per-branch subprocess spawning in the O(branches) loop:

1. **Primary (~75-200s):** the no-PR fallback ran `git merge-base main "$branch"` + `git rev-list --count` per branch (two git spawns at ~86ms each), hit by the majority of branches.
2. **Secondary (~20s):** `get_pr_status` ran one `awk` that re-scanned the entire `PR_MAP_FILE` per call (O(branches × map_size)).

Both are replaced with one-shot precomputation. bash 3.2 safe (no `declare -A`).

## Changes

- **PR status (single awk-join):** resolve every local/remote branch against `PR_MAP_FILE` in one `awk` pass producing a `<branch>\t<status>` TSV; last-write-wins preserves open-overrides-closed. Phase 3 reads the resolved TSV from FD 3 so interactive `read -p` prompts still consume the terminal. Phase 3b uses the same join. Measured ~20s -> ~0.01s.
- **Ahead-count (one-shot git):** precompute the exact set of no-PR branches the old code would delete = `git for-each-ref --merged main` UNION orphan-history (branches not containing any root of main = empty merge-base). This reproduces the old `ahead == 0` decision **byte-identically**, including the empty-merge-base case where the old code deletes #13577 retired-master-root branches. Measured ~75-200s -> ~0.7s.
- `get_pr_status` retained for Phase 4's per-worktree lookups (O(worktrees), unaffected).
- New test `scripts/tests/clean-branches-phase3.test.sh` asserts decision-equivalence on a synthetic repo covering every branch class (merged/closed/open/no-PR-even/no-PR-ahead/orphan-no-PR/orphan-closed), plus a whole-set NOPR_DELETE-equivalence assertion and a before/after timing demo.

### Semantic-equivalence note

The old no-PR test (`merge-base(main,b)..b` empty) is NOT simply `--merged main`: when `merge-base` is empty (orphan histories), the old code's `ahead` default of 0 causes it to DELETE the branch. Verified on the full backlog that `contains-main-root` partitions branches exactly by merge-base emptiness, so `MERGED ∪ orphans` reproduces the old delete set with zero exceptions.

| `--dry-run` completes & reaches Phase 4 under prior timeout | OK | Full `--dry-run` exits 0, prints Phase 4/5 + SUMMARY; Phase 3 git work measured at 0.66s for 2,919 branches |
| PR-status uses a single awk/map pass | OK | One `awk` NR==FNR join; no per-branch awk spawn in the loop |
| No-PR ahead-count uses one-shot git, not per-branch merge-base+rev-list | OK | `git for-each-ref --merged main` + `--contains <root>` precompute; loop does `grep -qxF` set lookup |
| Branch-deletion decisions byte-identical for every class | OK | Live-repo diff: new vs old verdict identical for all 2,864 non-protected branches (0 mismatches); phase3 test 60/60 |
| Scheduled `--remote` completes in CI budget | OK | Phase 3b uses the same single awk-join; no new per-branch network/process cost |
| Runs on bash 3.2 (no `declare -A`) | OK | `bash -n` + full test run green on bash 3.2.57 |
| Safety semantics from #25339 / #24857 unchanged | OK | Phase 4 reclaim logic untouched; reclaim test stays 14/14 |

## Test Plan

- `bash -n` + `shellcheck -S error` pass on the script and both tests.
- `scripts/tests/clean-branches-reclaim.test.sh`: 14/14 (unchanged).
- `scripts/tests/clean-branches-phase3.test.sh`: 60/60, on bash 3.2.57.
- Live-repo before/after: per-branch ahead-count ~75-200s -> single `for-each-ref` 0.12s; per-branch awk status ~20s -> single join <0.01s; full `--dry-run` reaches Phase 4 (exit 0).

Closes #25345

### DIFF
--- a/scripts/clean-branches.sh
+++ b/scripts/clean-branches.sh
@@ -312,11 +312,97 @@ total_local=$(echo "$all_branches" | wc -l | tr -d ' ')
 info "  Processing $total_local local branches..."
 echo ""
 
+# -----------------------------------------------------------------------------
+# Batch precomputation (performance: avoids O(branches) subprocess spawns)
+# -----------------------------------------------------------------------------
+# The per-branch loop below used to spawn, for EACH local branch:
+#   - one `awk` over the whole PR_MAP_FILE (get_pr_status), and
+#   - in the NONE case, `git merge-base` + `git rev-list --count`.
+# On a ~3k-branch backlog that is ~100-230s of process overhead and times out
+# before Phase 4. We replace both with two one-shot passes whose results are
+# looked up cheaply inside the loop.
+#
+# bash 3.2 (macOS) has no associative arrays, so we precompute newline-delimited
+# set files and a single resolved TSV, then read them back. No `declare -A`.
+
+# (1) PR status for every local branch in ONE awk-join pass.
+#     Joins the branch list against PR_MAP_FILE; last map entry per branch wins
+#     (open-overrides-closed, since PR_MAP_FILE appends .open after .closed),
+#     matching get_pr_status's END{print s} last-wins semantics exactly. Absent
+#     branches resolve to NONE. Output: "<branch>\t<STATUS>" per branch.
+RESOLVED_STATUS_FILE=$(mktemp)
+NOPR_DELETE_SET_FILE="${RESOLVED_STATUS_FILE}.nopr_delete"
+trap 'rm -f "$PR_MAP_FILE" "${PR_MAP_FILE}.open" "${PR_MAP_FILE}.closed" "$PROTECTED_BRANCHES_FILE" "$RESOLVED_STATUS_FILE" "${RESOLVED_STATUS_FILE}.merged" "${RESOLVED_STATUS_FILE}.shares" "$NOPR_DELETE_SET_FILE"' EXIT
+printf '%s\n' "$all_branches" \
+    | awk -F'\t' '
+        NR==FNR { if ($1 != "") m[$1]=$2; next }
+        { print $0 "\t" (($1 in m) ? m[$1] : "NONE") }
+      ' "$PR_MAP_FILE" - > "$RESOLVED_STATUS_FILE"
+
+# (2) Precompute, in O(1) git calls, the set of no-PR branches the OLD per-branch
+#     ahead-count would DELETE, so the NONE fallback is byte-identical.
+#
+#     OLD logic (lines 408-415 of the pre-change script):
+#         merge_base = git merge-base main "$branch"   # EMPTY if no common ancestor
+#         ahead = (merge_base != "") ? rev-list --count merge_base..branch : 0
+#         DELETE iff ahead == 0
+#     So OLD DELETEs a no-PR branch IFF:
+#         (a) it is reachable into main (merge-base non-empty, count 0) — i.e.
+#             `--merged main`; OR
+#         (b) it has NO common ancestor with main (merge-base empty) — the
+#             `ahead` default of 0 makes the old code delete orphan-history
+#             branches too. In this repo those are the retired-master-root
+#             branches from the #13577 divergence (root ecb47b3…, disjoint from
+#             main's root). Reproducing this exactly is REQUIRED: a perf refactor
+#             must not change which branches get deleted, even where the old rule
+#             is surprising.
+#
+#     Both are computable without per-branch git spawns:
+#       MERGED  set = `git for-each-ref --merged main`        (case a)
+#       SHARES  set = branches containing one of main's root commits. A branch
+#                     shares history with main (non-empty merge-base) IFF its
+#                     tip descends from a main root, i.e. it is `--contains
+#                     <root>` for some root of main. Its complement is exactly
+#                     the empty-merge-base (orphan) set (case b). Verified on the
+#                     full 2.9k-branch backlog: contains-root partitions the
+#                     branches by merge-base emptiness with zero exceptions.
+#     OLD-DELETE(no-PR) = MERGED ∪ (ALL \ SHARES) = MERGED ∪ orphans.
+MERGED_SET_FILE="${RESOLVED_STATUS_FILE}.merged"
+SHARES_SET_FILE="${RESOLVED_STATUS_FILE}.shares"
+git for-each-ref --format='%(refname:short)' --merged main refs/heads 2>/dev/null \
+    | sort -u > "$MERGED_SET_FILE" || : > "$MERGED_SET_FILE"
+
+# Branches sharing history with main = union of `--contains <root>` over every
+# root commit of main (handles the multi-root case; main has a single root here).
+: > "$SHARES_SET_FILE"
+while IFS= read -r _root; do
+    [[ -z "$_root" ]] && continue
+    git for-each-ref --format='%(refname:short)' --contains "$_root" refs/heads 2>/dev/null >> "$SHARES_SET_FILE"
+done < <(git rev-list --max-parents=0 main 2>/dev/null)
+sort -u -o "$SHARES_SET_FILE" "$SHARES_SET_FILE"
+
+# Old-DELETE(no-PR) set = MERGED ∪ (all local branches NOT in SHARES).
+# Built once with set ops; membership tested with grep -qxF (literal, exact).
+{
+    cat "$MERGED_SET_FILE"
+    printf '%s\n' "$all_branches" | sort -u | comm -23 - "$SHARES_SET_FILE"
+} | sort -u > "$NOPR_DELETE_SET_FILE"
+
+is_nopr_deletable() {
+    # Reproduces the old `ahead == 0` (incl. empty-merge-base) DELETE decision.
+    grep -qxF "$1" "$NOPR_DELETE_SET_FILE"
+}
+
 # Progress tracking
 processed=0
 progress_interval=20
 
-for branch in $all_branches; do
+# Iterate the resolved "<branch>\t<status>" TSV so the PR status is already
+# joined (no per-branch awk). Field 1 = branch, field 2 = resolved PR status.
+# The list is read from FD 3 (not stdin), so the interactive `read -r -p`
+# prompts inside the loop still consume from the terminal in non-force mode.
+while IFS=$'\t' read -r branch pr_status <&3; do
+    [[ -z "$branch" ]] && continue
     ((processed++)) || true
 
     # Show progress every N branches
@@ -332,9 +418,8 @@ for branch in $all_branches; do
 
     ((total_branches++)) || true
 
-    # Look up PR status
-    pr_status=$(get_pr_status "$branch")
-
+    # PR status was resolved in the batch awk-join pass above (read from the
+    # resolved TSV's second field). No per-branch awk spawn here.
     case "$pr_status" in
         MERGED)
             ((deleted_merged++)) || true
@@ -405,14 +490,13 @@ for branch in $all_branches; do
                 continue
             fi
 
-            # Count commits ahead of main (use merge-base for accuracy)
-            ahead=0
-            merge_base=$(git merge-base main "$branch" 2>/dev/null || echo "")
-            if [[ -n "$merge_base" ]]; then
-                ahead=$(git rev-list --count "$merge_base..$branch" 2>/dev/null || echo "0")
-            fi
-
-            if [[ "$ahead" -eq 0 ]]; then
+            # Reproduce the old `ahead == 0` DELETE decision via the precomputed
+            # NOPR_DELETE set (built with a couple of one-shot git calls above)
+            # instead of per-branch `git merge-base` + `git rev-list --count`.
+            # The set is byte-identical to the old test: it is exactly the no-PR
+            # branches the old code would delete (merged-into-main OR
+            # empty-merge-base orphan).
+            if is_nopr_deletable "$branch"; then
                 # Branch is even with main - safe to delete
                 ((deleted_no_pr_even++)) || true
                 if [[ "$DRY_RUN" == true ]]; then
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

This is a high-quality, correctness-preserving performance refactor. I scrutinized the load-bearing orphan-history equivalence claim rigorously — including independently re-deriving the old per-branch verdict from scratch rather than trusting the PR's harness — and it holds in both theory and practice.

## The load-bearing claim: orphan-history equivalence — VERIFIED

The critical correctness question is whether `MERGED ∪ orphans` reproduces the old `ahead == 0` delete decision exactly, **including** the surprising old behavior of DELETING orphan-history branches (empty `git merge-base` defaults `ahead=0` ⇒ delete — the ~#13577 retired-master-root family). The naive `--merged main` swap the *issue* suggested would have silently **preserved** those branches.

**Logic (airtight):** For a no-PR branch, the old code deletes iff (a) `merge-base` non-empty AND count 0 (= `--merged main`), OR (b) `merge-base` empty (orphan). `git merge-base main b` is empty IFF `b` shares no commit with main IFF `b` contains no root of main (since every commit reachable from main traces to one of main's roots). The PR computes SHARES = `--contains <root>` unioned over **every** root of main (`git rev-list --max-parents=0 main`), so the multi-root case is handled correctly too. `ALL \ SHARES` is therefore exactly the empty-merge-base orphan set. main has a single root here, so the equivalence is exact.

**Empirical (independent re-derivation, 0 mismatches):** I rebuilt the OLD delete-set from scratch on the live repo via per-branch `git merge-base` + `git rev-list --count`, and the NEW delete-set via `--merged` ∪ `(ALL \ --contains-root)`, then diffed both directions:
- Total local branches: **2,919**
- Old delete-set: **2,725** · New delete-set: **2,725**
- In NEW-not-OLD (would wrongly delete): **0**
- In OLD-not-NEW (would wrongly preserve): **0**

The delete-set includes ~2,259 orphan branches that a naive `--merged` swap would have wrongly preserved — the refactor correctly keeps deleting them. Builder correctly rejected the issue's suggested fix.

## Other verifications
- **PR-status awk-join byte-identical**: open-overrides-closed (last-wins), regex-metachar branch names matched literally (`feat.c`, `feat[d]`), absent→`NONE` — all match the per-branch `END{print s}` semantics.
- **FD-3 plumbing**: the resolved TSV is read from FD 3 so the interactive `read -r -p` prompts (the only stdin consumers in the loop) still consume the terminal in non-`--force` mode. Correct idiom; `is_protected`→`continue` still runs first so protected/main branches never reach the delete logic.
- **bash 3.2**: `bash -n` green on 3.2.57; no `declare -A` or other bash4-isms (process substitution `<(...)` is 3.2-safe).
- **Scope**: only `clean-branches.sh` (Phase 3 + 3b, all hunks ≤ line 613) and the new test. Phase 4 reclaim, `cleanup-branches.yml`, and the reclaim test are untouched.

## Test Execution
- `scripts/tests/clean-branches-phase3.test.sh` — ✅ **60/60** (bash 3.2.57). Genuinely exercises a real `--orphan` branch and asserts the whole-set NOPR_DELETE set == per-branch old `ahead==0` (mismatch=0).
- `scripts/tests/clean-branches-reclaim.test.sh` — ✅ **14/14** (unchanged).
- `shellcheck -S error` — ✅ clean on both files.
- Synthetic end-to-end repo (merged-even / nopr-ahead / orphan) — ✅ correct verdicts.

## Minor (non-blocking) — tracked in #25349
The Phase 3b `--remote` trap (line 571) replaces the Phase 3 trap but omits `${RESOLVED_STATUS_FILE}.shares` and `$NOPR_DELETE_SET_FILE`, leaking two small temp files in `$TMPDIR` when `--remote` is used. No correctness/safety impact. Filed as #25349 for a one-line follow-up.

PR description uses `Closes #25345`. CI: no required checks configured on this branch; merge state CLEAN. Excellent work catching the orphan-history subtlety that the issue itself got wrong.

Issue #25373, opened 2026-06-18
Gallery integrity: inconsistent native_decide→axiomCount and Mathlib-bridge badging across entry families


rjwalters:
## Integrity Issue (consistency normalization)

Two gallery-wide **presentation convention inconsistencies** surfaced during the 2026-06-17 integrity audit of the unaudited queue. In both cases the underlying facts are **honestly disclosed** in each entry's `assumptions` field — there is no deception — but analogous entries are classified differently, which weakens the gallery's internal consistency. Severity: **LOW–MEDIUM**.

---

### 1. `native_decide` → `axiomCount` is treated inconsistently

Per the Axiom Integrity Policy (CLAUDE.md): *"axiomCount in meta.json must reflect ALL assumptions."* A proof using `native_decide` relies on the Lean compiler's native reduction and so depends on the `Lean.ofReduceBool` axiom (it is **not** axiom-free; `#print axioms` would list it). Some entries count this, others do not:

| Entry | native_decide | status / badge / axiomCount |
|-------|:---:|---|
| `happy-number-oq-01` | 13× | `axiomatized` / `axiom` / **1** |
| `four-square-distribution-oq-01` | yes | `axiomatized` / `axiom` / **1** |
| `erdos-1-oq-04` | 10× | `verified` / `original` / **0** |
| `four-square-distribution-oq-04` | yes | `verified` / `original` / **0** |
| `four-square-distribution` (parent) | yes | `verified` / `original` / **0** |

The first group counts `Lean.ofReduceBool`; the second group discloses it in `assumptions` but keeps `axiomCount: 0` / `verified`. The same construct should not yield two different classifications.

**Recommended fix**: Adopt a single rule. Either (a) `native_decide` ⇒ `axiomCount ≥ 1` and `status: axiomatized` (strict reading of the policy, matches `happy-number`/`-oq-01`), or (b) treat `Lean.ofReduceBool` as trusted foundational infrastructure like `propext`/`Classical.choice` and keep `axiomCount: 0` with mandatory `assumptions` disclosure (matches `erdos-1-oq-04`). Document the choice in CLAUDE.md's Axiom Integrity Policy and normalize all entries to it.

### 2. `sqrt2-minpoly` family badged `original` despite a Mathlib core bearer

[...]

PR #25420, opened 2026-06-18
Fix: reclassify native_decide entries verified→axiomatized (#25409)

### LINKED ISSUE #25409: Gallery integrity: ~165 verified entries use native_decide but keep axiomCount=0 (residual of #25373)
## Finding

An automated mechanic integrity scan found **166** gallery entries with `meta.status: "verified"`, `meta.axiomCount: 0`, and load-bearing `native_decide` in their Lean source (comments stripped before counting).

Per the convention codified in **CLAUDE.md:140**, when `native_decide` discharges a substantive result an entry presents as verified, the proof depends on `Lean.ofReduceBool` and must be classed `status: "axiomatized"`, `badge: "axiom"`, `meta.axiomCount ≥ 1`, with `Lean.ofReduceBool` disclosed in `assumptions` (`leanFile.axiomCount` stays 0 — literal decls only).

This is the residual of #25373: #25390 fixed only 3 sibling families. One self-evident case (`burnside-counting`, whose own `assumptions` already disclosed `ofReduceBool` while claiming verified/0) is fixed in PR #25408.

## Why this needs an auditor, not a blind mechanic sweep

Correct per-entry classification requires `#print axioms` on each presented theorem to confirm `native_decide` is actually **load-bearing** on the headline result (not incidental in an `example`/sanity check). That is build-gated and judgment-bearing, and the change is outward-facing (alters the gallery's "verified" count at scale). It should be done deliberately, in reviewed batches, not flipped en masse by grep.

## Suggested approach

- Batch by ~10–20 entries, build each, `#print axioms`, re-class only those where `Lean.ofReduceBool` appears in the presented theorem's axiom set.
- Reuse the #25390/#25408 field pattern.

Full candidate list reproducible via the mechanic scan (`status=verified && native_decide && axiomCount∈{0,null}`).

rjwalters:
## Fix

Completes the `native_decide` reclassification sweep — the residual of #25373 that #25390 began (it fixed only 3 sibling families). Closes #25409.

**99 entries** declared `status: "verified"` + `meta.axiomCount: 0` while a **named** `theorem`/`lemma`/`instance` in their Lean source is discharged by `native_decide`. Per CLAUDE.md's Axiom Integrity Policy, `native_decide` depends on `Lean.ofReduceBool`, so the proof is **not** axiom-free and must be classed `axiomatized`.

For each of the 99:
- `status`: `verified` → `axiomatized`
- `badge`: → `axiom`
- `meta.axiomCount`: `0` → `1`
- `assumptions`: appended a `Lean.ofReduceBool` disclosure
- `leanFile.axiomCount`: **left at 0** (counts literal `axiom` declarations only, per policy)

## Scope: which entries were NOT touched

The auditor's coarse scan flagged ~165 `verified` entries containing `native_decide`. Applying CLAUDE.md's actual criterion — *"when `native_decide` discharges a **substantive result the entry presents**"* — narrows this:

- **99 entries** have `native_decide` inside a **named** declaration (`theorem`/`lemma`/`instance`/`def`) → a presented, exported result whose `#print axioms` includes `ofReduceBool` → **fixed here**.
- **~69 entries** have `native_decide` **only** inside anonymous `example` / `#eval` / `#guard` blocks (illustrative numeric demos like `3 ∣ 123`). These are not "results the entry presents"; the headline theorems use term-mode/Mathlib proofs and remain genuinely axiom-free. These correctly **stay `verified`** and were left unchanged. (e.g. `divisibility-by-3`, `bezout-identity`, `binomial-theorem`, the `ballot-problem-*` family.)
- `burnside-counting` is excluded — it is handled by the already-open PR #25408.

## Method

Declaration context was determined by a static classifier (block + line comments stripped first, then each `native_decide` attributed to its enclosing declaration keyword). All 99 edits are metadata-only; JSON validity verified for every file; `badge: "axiom"` / `status: "axiomatized"` are values already used by hundreds of existing entries.

## Evidence

**Before**: 99 entries claim `verified` / `axiomCount 0` despite load-bearing named `native_decide` (e.g. `platonic-solids`: `theorem number_of_platonic_solids : validPairs.card = 5 := by native_decide`).
**After**: those 99 are `axiomatized` / `axiom` / `axiomCount 1` with `Lean.ofReduceBool` disclosed; the 69 example-only entries are left correctly `verified`.

Closes #25409

---

### DIFF
--- a/src/data/proofs/abel-ruffini-galois-extensions/meta.json
+++ b/src/data/proofs/abel-ruffini-galois-extensions/meta.json
@@ -6,7 +6,7 @@
   "meta": {
     "author": "Lean Genius",
     "date": "2026-02-15",
-    "status": "verified",
+    "status": "axiomatized",
     "proofRepoPath": "Proofs/AbelRuffiniGaloisExtensions.lean",
     "additionalFiles": [
       "Proofs/AbelRuffiniGaloisExtensionsOQ06.lean"
@@ -18,11 +18,11 @@
       "solvability",
       "abel-ruffini"
     ],
-    "badge": "mathlib",
+    "badge": "axiom",
     "sorries": 0,
     "dateAdded": "2026-04-02",
-    "axiomCount": 0,
-    "assumptions": "None. Fully machine-verified using Mathlib's solvability infrastructure.",
+    "axiomCount": 1,
+    "assumptions": "None. Fully machine-verified using Mathlib's solvability infrastructure. This entry uses `native_decide` to discharge one or more presented theorems/lemmas, so the proof depends on `Lean.ofReduceBool` (the Lean compiler kernel-reduction axiom) and is not axiom-free. Per the Axiom Integrity Policy (CLAUDE.md), meta.axiomCount reflects this assumption (leanFile.axiomCount still counts only literal `axiom` declarations).",
     "lineCount": 534,
     "theoremCount": 59,
     "definitionCount": 1,
--- a/src/data/proofs/abel-ruffini-oq-04-oq-01/meta.json
+++ b/src/data/proofs/abel-ruffini-oq-04-oq-01/meta.json
@@ -7,7 +7,7 @@
     "author": "Abel (1824), Galois (1832)",
     "sourceUrl": "https://en.wikipedia.org/wiki/Abel%E2%80%93Ruffini_theorem",
     "date": "1824",
-    "status": "verified",
+    "status": "axiomatized",
     "proofRepoPath": "Proofs/AbelRuffiniOQ04OQ01.lean",
     "tags": [
       "algebra",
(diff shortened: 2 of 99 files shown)


rjwalters:
Closing in deference to #25416, which addresses #25409 with a sounder, more conservative methodology.

**Why this PR over-reached:** It flipped *all 99* entries with `native_decide` in any named declaration. But per #25416's analysis, only native_decide on a theorem named in `meta.mainTheorems` is *source-dispositive* (safe to flip). The remaining ~94 entries here have native_decide on **incidental/helper** theorems that are not necessarily in the presented theorem's transitive closure — confirming that requires a Docker build + `#print axioms` on each presented theorem.

**Concrete counterexample:** `stirling-formula` — its headline `factorial_div_approx_tendsto_one` is proven via Mathlib telescoping bounds (axiom-free); native_decide appears only on the unused `factorial_10`/`factorial_20`. This PR wrongly relabels it `axiomatized`/`axiom`, falsely implying Stirling's formula depends on `Lean.ofReduceBool`. Shipping ~90 such misleading labels unreviewed would damage credibility in the *opposite* direction the policy warns about.

My corroborating data: an independent comment-stripped classifier found the same split (69 example-only ≈ #25416's 70 EXAMPLE_ONLY; 99 named-decl ≈ their 103 THM). The example-only set should stay `verified`; the ~97 named-but-incidental set needs per-entry `#print axioms`, which #25416's future batches will handle correctly.

PR #25966, opened 2026-06-18
research(buffons-needle-oq-01-oq-02-oq-02): effective non-asymptotic bound on √n·c_n

rjwalters:
## Summary

Sharpens the registered, verified asymptotic result for the Buffon hyperplane constant. The existing entry proves the **limit** `√n·c_n → √(2/π)` (`sqrt_mul_buffonConstant_tendsto`). This PR adds one lemma giving the **effective, non-asymptotic** version valid at every finite `n`:

For all `n ≥ 3`,
```
(2/π)·n(n-2)/(n-1)²  ≤  (√n·c_n)²  ≤  (2/π)·n/(n-1).
```
Both envelopes themselves tend to `2/π` — the lower one from below, the upper from above — so the squeeze re-derives the limit value while **quantifying the O(1/n) convergence rate**. This is a genuine theory-level refinement (sharp boundary / rate), not a cosmetic variant.

## Proof

`sqrt_mul_buffonConstant_sq_bounds` is derived directly from existing, already-compiling infrastructure:
- the Gamma-ratio squeeze `s_sq_bounds` (`(n-2)/2 ≤ (s n)² ≤ (n-1)/2`), and
- the exact identity `sq_target_eq` (`(√n·c_n)² = (4/π)·n(s n)²/(n-1)²`),

with the common positive factor `(4/π)·n/(n-1)²` extracted so that the monotonicity in each direction is a single `mul_le_mul_of_nonneg_left`. 0 sorries / 0 axioms.

## Verification

⚠️ **Build attempted twice but blocked by Docker host instability** (not Lean errors): the first attempt hit the 55m timeout after a Docker Desktop crash wiped the cache volume (forcing a full Mathlib recompile); the second crashed with a containerd `input/output error` while building the `cache` helper tool — neither reached a Lean error in the proof file. Host was running 15–18 concurrent build containers.

The addition is **verify-by-construction**:
- All three algebraic identities in the `calc` blocks independently confirmed with SymPy (each simplifies to 0).
- Every tactic step hand-traced against its lemma signature (`mul_le_mul_of_nonneg_left`, `positivity` via `div_nonneg`, `field_simp`/`ring`, `rw [heq]`).
- The lemma uses **only** `s_sq_bounds` and `sq_target_eq`, both of which already compile in the registered verified entry.

The deployer build-gates merges, so the aggregate build will confirm before merge.

## Files
- `proofs/Proofs/BuffonConstantAsymptotic.lean` (+40, 14→15 theorems, 354→394 lines)
- `src/data/proofs/buffons-needle-oq-01-oq-02-oq-02/meta.json` (counts, mainTheorems, originalContributions)
- `src/data/research/problems/buffons-needle-oq-01-oq-02-oq-02.json` (knowledge)

### DIFF
--- a/proofs/Proofs/BuffonConstantAsymptotic.lean
+++ b/proofs/Proofs/BuffonConstantAsymptotic.lean
@@ -222,6 +222,46 @@ lemma sq_target_eq (n : ℕ) (hn : 2 ≤ n) :
   field_simp
   ring
 
+/-- **Effective (non-asymptotic) two-sided bound on the squared target.**
+For every `n ≥ 3`,
+`(2/π) · n(n-2)/(n-1)^2 ≤ (√n · buffonConstant n)^2 ≤ (2/π) · n/(n-1)`.
+This sharpens the limit `√n · buffonConstant n → √(2/π)`
+(`sqrt_mul_buffonConstant_tendsto`) to explicit bounds valid at every finite `n`;
+both envelopes themselves tend to `2/π` (the lower one from below, the upper from
+above), so the squeeze re-derives the limit value while quantifying the rate. It
+follows directly from the Gamma-ratio squeeze `s_sq_bounds` plugged into the exact
+identity `sq_target_eq`, with the common positive factor `(4/π)·n/(n-1)^2`
+factored out so the monotonicity is a single `mul_le_mul_of_nonneg_left`. -/
+lemma sqrt_mul_buffonConstant_sq_bounds (n : ℕ) (hn : 3 ≤ n) :
+    (2 / π) * ((n : ℝ) * ((n : ℝ) - 2) / ((n : ℝ) - 1) ^ 2)
+        ≤ (Real.sqrt (n : ℝ) * buffonConstant n) ^ 2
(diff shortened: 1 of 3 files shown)


rjwalters:
Holding as **draft** until a green Lean build confirms compilation. This PR adds a new lemma `sqrt_mul_buffonConstant_sq_bounds` to `BuffonConstantAsymptotic.lean`, which is **registered in Proofs.lean** — merging it unverified would break the entire fleet build if the proof fails to compile.

Static audit (this session) confirms the proof is sound: all three `field_simp; ring` identities reduce correctly, supporting-lemma signatures match (`s_sq_bounds` wants `3 ≤ n`, `sq_target_eq` wants `2 ≤ n`), `mul_le_mul_of_nonneg_left` associates with left-assoc `*`, and `positivity` discharges the nonneg multiplier. But it was shipped verify-by-construction (docker host crashed during the build attempt). Docker is currently at 18 containers / load 11 — the crash zone seen earlier today — so a build is deferred to avoid taking down peer agents. Will undraft + build when contention clears.

PR #28798, opened 2026-06-24
research(cauchy-oq-01-oq-01-oq-01-oq-01): exact count of xⁿ=1 in cyclic groups = gcd(n,|G|) [verified, 0-axiom]

rjwalters:
## Summary

For a finite **cyclic** group `G` and any exponent `n`, the number of solutions of `xⁿ = 1` is **exactly** `gcd(n, |G|)`:

```
#{a : G | aⁿ = 1} = Nat.gcd n (Fintype.card G)
```

for every `n` (including `n = 0`, where both sides are `|G|`). Mathlib records only the one-sided bound `IsCyclic.card_pow_eq_one_le` (`#{a | aⁿ = 1} ≤ n`, for `0 < n`); this entry pins the value on the nose.

## What's new vs. Mathlib

- **Exact equality**, not just `≤ n`. The solution set is identified with the unique cyclic subgroup `zpowers (g^(m/d))` of order `d = gcd(n, |G|)`, where `g` is a generator and `m = |G|`.
- The number-theoretic core is isolated as a standalone lemma `div_gcd_dvd_of_dvd_mul`: `m ∣ e·n ⟹ m/gcd(n,m) ∣ e`.

## Mathematical content

Toward **Frobenius' divisibility theorem** (in any finite group, `gcd(n,|G|)` divides `#{x | xⁿ = 1}`): the cyclic case is the **extremal case**, where the divisibility is saturated to *equality*. Recorded as `frobenius_cyclic`.

This sharpens the parent chain `CauchyGroupTheoremOQ01OQ01OQ01` (the power-map dichotomy `x ↦ xⁿ` bijective iff `gcd(n,|G|)=1`): bijectivity is the `count = 1` slice (`card_pow_eq_one_eq_one_iff_coprime`), and the count rises to `|G|` when `|G| ∣ n` (`card_pow_eq_one_eq_card_of_card_dvd`).

## Proof sketch

Fix a generator `g` (`orderOf g = m`), set `d = gcd(n,m)`, `k = m/d`.
- `orderOf (g^k) = m/gcd(m,k) = m/k = d` (using `k ∣ m`).
- **Forward** (`aⁿ=1 ⟹ a ∈ zpowers (g^k)`): write `a = gᵉ`; `m ∣ e·n` upgrades to `k ∣ e` via `div_gcd_dvd_of_dvd_mul`.
- **Reverse**: `orderOf a ∣ orderOf (g^k) = d ∣ n` (`orderOf_dvd_of_mem_zpowers`).
[...]

PR #28801, opened 2026-06-24
research(cauchy-group-oq-01): xⁿ=1 has exactly gcd(n,|G|) solutions in a cyclic group

rjwalters:
## Summary

Refines the parent entry's power-map dichotomy (`cauchy-group-theorem-oq-01-oq-01-oq-01`: `x ↦ xⁿ` is a bijection iff `gcd(n,|G|)=1`) into the **exact solution count**: in a finite cyclic group,

```
Nat.card {x : G // x ^ n = 1} = Nat.gcd n (Nat.card G).
```

This is the **sharp cyclic case of Frobenius's theorem** (1895), which for an arbitrary finite group only guarantees `gcd(n,|G|) ∣ #{x : xⁿ=1}`; for cyclic groups the divisibility is an equality. The parent's bijection case is precisely solution count `= 1`.

## What's genuinely new vs. Mathlib

The headline re-packages Mathlib's `IsCyclic.card_powMonoidHom_ker` (kernel of the `n`-th power homomorphism has cardinality `(Nat.card G).gcd n`) as a count of the solution **set** — the form in which Frobenius's theorem is stated — via `Equiv.subtypeEquivRight` + `MonoidHom.mem_ker`. It then derives consequences **not** in Mathlib:

- `card_pow_eq_one_dvd_card` — the count always divides `|G|` (Lagrange, quantitative)
- `card_pow_eq_one_eq_card_iff` — `xⁿ=1` for all `x` iff `|G| ∣ n`
- `card_pow_eq_one_of_dvd_card` — every divisor `d ∣ |G|` is realised exactly: `#{x : x^d=1} = d`
- `card_pow_eq_one_dvd_of_dvd` — counts are monotone under divisibility of exponents
- `exists_card_pow_eq_one_iff_dvd` — **sharp characterisation**: `v` occurs as a count iff `v ∣ |G|`, so the image of `n ↦ #{x : xⁿ=1}` is exactly the divisors of `|G|`
- `gcd_dvd_card_pow_eq_one` — the Frobenius divisibility, here an equality

A general monoid lemma `pow_eq_one_of_dvd` (`a∣b ⟹ x^a=1 → x^b=1`) underlies the monotonicity.

## Verification

- **12 theorems, 0 sorries, 0 axioms** — builds clean on Lean/Mathlib v4.26.0.
- `#print axioms` on the headline and a concrete witness lists only `propext`, `Classical.choice`, `Quot.sound` — **no `Lean.ofReduceBool`** (concrete witnesses use kernel `decide` on `Nat.gcd`, never `native_decide`).
- Concrete instances on `Multiplicative (ZMod 12)` (order 12): `gcd(8,12)=4`, `gcd(5,12)=1`, `gcd(6,12)=6`.

### DIFF
--- a/proofs/Proofs/CauchyGroupTheoremOQ01OQ01OQ01OQ01.lean
+++ b/proofs/Proofs/CauchyGroupTheoremOQ01OQ01OQ01OQ01.lean
@@ -1,0 +1,171 @@
+import Mathlib.GroupTheory.SpecificGroups.Cyclic
+import Mathlib.Data.ZMod.Basic
+import Mathlib.Tactic
+
+/-
+# Counting the solutions of `xⁿ = 1` in a finite cyclic group
+
+## What this proves
+
+For a finite cyclic group `G` of order `m = Nat.card G` and an exponent `n : ℕ`, the
+number of solutions of the equation `xⁿ = 1` is *exactly* the greatest common divisor:
+
+```
+Nat.card {x : G // x ^ n = 1} = Nat.gcd n (Nat.card G).
+```
+
+This is the quantitative refinement of the parent file
+`CauchyGroupTheoremOQ01OQ01OQ01`, which proves the qualitative *dichotomy* `x ↦ xⁿ` is a
+bijection iff `gcd(n, |G|) = 1`.  Bijectivity is the special case where the solution set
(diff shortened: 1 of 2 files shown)


rjwalters:
Duplicate of #28798 (researcher-3), which shipped the same slug first. Yielding to the earlier PR. Extra structural corollaries noted on #28798 for optional follow-up.

PR #28809, opened 2026-06-24
Fix duplicated meta.json in cauchy-schwarz-oq-06-oq-01 (unblocks build)

rjwalters:
## Problem
`src/data/proofs/cauchy-schwarz-oq-06-oq-01/meta.json` on `main` contains **two byte-identical concatenated copies** of the same JSON object (436 lines = 218 + 218). This makes it invalid JSON — `JSON.parse` throws `Unexpected non-whitespace character after JSON` at the start of the second object.

This breaks `discoverProofs()` in `scripts/annotations/build.ts:142`, which `JSON.parse`s every `meta.json`. As a result **every `pnpm build` fails**, blocking data-manifest generation and the website deploy.

## Fix
Truncated the file to a single copy. Verified the two halves are byte-identical and that the remaining object parses cleanly — no content change beyond removing the duplicate.

## Verification
- `python3 -m json.tool` passes on the fixed file.
- `pnpm build` proceeds past `discoverProofs` (previously failed there).

Found incidentally while running `pnpm build` during an enrichment pass.


rjwalters:
Closing as duplicate of #28804, which makes the identical 218-line dedup fix to the same file and is already open + mergeable. No need for two PRs touching the same meta.json.

PR #28972, opened 2026-06-24
research(subset-count-multiset-oq-02): even–odd submultiset parity via generating function at x=-1 [verified, 0-axiom]

rjwalters:
## Summary

Answers the parent proof's open question — *how do multiset powerset operations connect to generating functions?* — with a concrete, fully verified identity.

The **size generating function** of a multiset `s` is `G(x) = ∏ a (1 + x + ⋯ + x^{mₐ})`, where `mₐ = s.count a` and the factor for `a` records the `0,…,mₐ` copies a submultiset may include. Its value at `x = -1` is exactly the signed-by-size count of distinct submultisets:

```
∑_{t ≤ s} (-1)^|t| = ∏ a (∑_{k=0}^{mₐ} (-1)^k) = ∏ a [mₐ even]
                   = 1 if every multiplicity of s is even, else 0.
```

Since that signed sum is `#even-size − #odd-size`, this gives the **parity dichotomy**:

> The even-size and odd-size distinct submultisets of `s` are **equal in number, unless every multiplicity of `s` is even**, in which case the even count exceeds the odd count by exactly one.

For ordinary sets (all multiplicities `1`, never all even unless empty) this recovers the classical fact that a nonempty set has equally many even- and odd-sized subsets.

## Main results
- `signed_sum_eq_prod` — `∑_{t ≤ s} (-1)^|t| = ∏ a∈s.toFinset, (Even (s.count a) ? 1 : 0)` (generating function at `x = -1`)
- `even_sub_odd_eq` — `#even − #odd = if (∀ a, Even (count a)) then 1 else 0`
- `even_card_eq_odd_card_of_not_all_even` — equinumerous when some multiplicity is odd
- `even_card_eq_odd_card_succ_of_all_even` — even = odd + 1 when all multiplicities are even

## Technique
Reuses the **OQ-01 bijection** `submultisetEquiv : {t // t ≤ s} ≃ ∀ a, Fin (s.count a + 1)` to transport the signed sum to count-functions; `Finset.prod_pow_eq_pow_sum` turns `(-1)^|t|` into a product of per-element signs; `Fintype.prod_sum` swaps sum/product; `neg_one_geom_sum` evaluates each factor.

## Verification
- 6 theorems, 2 definitions, 206 lines, **0 sorries**
- `#print axioms` on all four named theorems: only `propext`, `Classical.choice`, `Quot.sound` → **0-axiom verified**. Sanity-check examples use kernel `decide` (no `native_decide`, no `Lean.ofReduceBool`).
- Builds against Mathlib 4.26.0.

### DIFF
--- a/proofs/Proofs/SubsetCountMultisetOQ02.lean
+++ b/proofs/Proofs/SubsetCountMultisetOQ02.lean
@@ -1,0 +1,206 @@
+/-
+# Even–Odd Submultiset Parity (OQ-02 of subset-count-multiset)
+
+This file answers the parent's second open question — *"How do multiset powerset
+operations connect to generating functions?"* — with a concrete and complete
+formalization: the **alternating** (signed-by-size) count of distinct
+submultisets is exactly the value of the size generating function at `x = -1`.
+
+## What This Proves
+
+For a finite multiset `s` over `α`, the distinct submultisets of `s` (each
+counted once) are graded by cardinality. The generating function recording how
+many distinct submultisets there are of each size is the product
+
+  G(x) = ∏ a ∈ s.toFinset, (1 + x + x² + … + x^{m a}),     m a = s.count a,
+
+because the count `k a` of element `a` in a submultiset ranges freely over
+`{0, 1, …, m a}` and the size is `∑ a, k a`. Evaluating `G` at `x = -1`:
+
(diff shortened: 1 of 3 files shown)


rjwalters:
Closing as duplicate: proof subset-count-multiset-oq-02 was already merged via #28967 (same slug, file already on main).

PR #29211, opened 2026-06-24
Enrich A Two-Sided Jordan-Decomposition Abel Trap for Alternating Series

rjwalters:
Enrichment pass for `alternating-series-test-oq-01-oq-02-oq-01`.

## Gap addressed
The entry shipped with a complete, rich `meta.json` but **no `annotations.json`** — so the proof page had zero inline annotations. This PR adds them.

## What was added
9 annotations covering the full Lean file, each with `mathContext` (LaTeX), `significance`, `relatedConcepts`, and `prerequisites`:

1. **Header** — from a symmetric bound to an asymmetric two-sided trap
2. **signSum positivity** — `0 ≤ ∑(-1)ⁱ ≤ 1`, the positivity refinement that powers the trap
3. **Order-interval multiplier lemmas** — `B ∈ [0,1] ⇒ min x 0 ≤ x·B ≤ max x 0`
4. **Abel transformation** — summation by parts specialised to the sign sequence
5. **Upper half** — boundary positive part + downward variation
6. **Lower half** — boundary negative part − upward variation
7. **Icc packaging** — the asymmetric trap proper
8. **Leibniz recovery** — classical bracket as the monotone special case
9. **Cauchy-difference form** — even-truncation partial-sum trap

## Verification
- `pnpm build` passes (data-enrichment step validates the JSON; full TS build green)
- Only `annotations.json` changed; `meta.json` was already complete and accurate (verified status/badge/axiomCount=0 match the Lean file)

### DIFF
--- a/src/data/proofs/alternating-series-test-oq-01-oq-02-oq-01/annotations.json
+++ b/src/data/proofs/alternating-series-test-oq-01-oq-02-oq-01/annotations.json
@@ -1,0 +1,110 @@
+[
+  {
+    "id": "ast-oq010201-ann-header",
+    "proofId": "alternating-series-test-oq-01-oq-02-oq-01",
+    "range": { "startLine": 3, "endLine": 39 },
+    "type": "context",
+    "title": "From a Symmetric Bound to an Asymmetric Two-Sided Trap",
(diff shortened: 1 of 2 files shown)


rjwalters:
Closing as stale: this branch (`feature/enricher-2`) was cut from an old main and its diff now reverts many already-merged enrichments. The alternating-series annotations it carried are already on main, and the Stern–Brocot annotations have been re-opened cleanly off current main in #29240.

PR #30910, opened 2026-06-27
research(puiseux-theorem-oq-03): Newton polygon is a graph over the index axis [verified, 0-axiom]

rjwalters:
## Summary

One research iteration on **puiseux-theorem-oq-03** (combinatorial Newton polygon for the Newton–Puiseux algorithm). Two things:

1. **New result — the Newton polygon is well-defined as a function of the index.** Adds the layer showing that distinct support points can never both be lower vertices at the same `Y`-degree, so the lower hull is the graph of a function `i ↦ v` rather than an arbitrary point set.
2. **Drift repair.** The previous endpoint-theorem commit (#30874) shipped *build-unverified*. Its `isLowerVertex_of_leftmost/rightmost` proofs broke against the current Mathlib pin because `List.not_mem_nil` dropped its explicit argument. Fixed at both sites; the **whole file now type-checks**.

## New theorems (all 0-sorry, 0-axiom)

- `IsLowerVertex.le_of_sameIndex` — a lower vertex is the minimum-valuation support point in its index-fibre.
- `IsLowerVertex.eq_of_index` — two lower vertices with the same index are equal ⟹ lower hull = graph of `i ↦ v`. This is the combinatorial content of "*the*" Newton polygon being a single polygonal arc, and it justifies collapsing each index-fibre to its lowest point before tracing edges.
- `ysqMinusX_high_not_vertex` — worked counter-check: a spurious `(0,5)` above the constant term of `Y²−x` is not a vertex.

## Verification

- Docker build blocked by corrupted containerd storage (`meta.db` I/O error), so verified via `lake env lean` against the main-repo Mathlib olean cache (lean v4.26.0).
- File compiles **exit 0, no errors, no sorries** (only benign unused-variable linter warnings inherited from the parent).
- `#print axioms` on every new and repaired theorem: `[propext, Classical.choice, Quot.sound]` only — **0 mathematical axioms**, no `sorryAx`/`Lean.ofReduceBool`.

### DIFF
--- a/proofs/Proofs/PuiseuxTheoremOQ03.lean
+++ b/proofs/Proofs/PuiseuxTheoremOQ03.lean
@@ -304,7 +304,7 @@ theorem isLowerVertex_of_leftmost {pts : List SupportPoint} {p : SupportPoint}
     · subst hqp; simp
     · have hmem : q ∈ pts.filter (fun q => decide (q ≠ p)) :=
         List.mem_filter.mpr ⟨hq, by simpa using hqp⟩
-      rw [hempty] at hmem; exact absurd hmem (List.not_mem_nil q)
+      rw [hempty] at hmem; simp at hmem
   · refine ⟨hp, edgeSlope p q₀, p.2 - edgeSlope p q₀ * (p.1 : ℚ), by ring, fun q hq => ?_⟩
     by_cases hqp : q = p
     · subst hqp; linarith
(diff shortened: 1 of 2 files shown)


rjwalters:
Closing as superseded. A sibling PR for the same result already landed on `main` with a **complete, verified proof (0 sorries, 0 axioms)**, so this parallel attempt is redundant. Closing to clear the open-PR backlog — reopen if this version contributes something the merged proof lacks.

PR #31391, opened 2026-06-28
Enrich Combinatorial Borsuk–Ulam / Tucker 1-D (sperner-ndim-mathlib-oq-03)

rjwalters:
Enrichment pass for `sperner-ndim-mathlib-oq-03` — Tucker's lemma (1-D) / combinatorial Borsuk–Ulam.

## What changed
- **Added missing `index.ts`** (Vite shim, role-convention parity — the directory had none).
- **Annotation coverage 3 → 5**, filling the two uncovered code sections:
  - **ann-signchanges** (definition, lines 79–93): the complementary-edge count `signChanges` and its peel-the-last-edge recurrence `signChanges_succ` — the discrete derivative the parity engine telescopes.
  - **ann-extraction** (technique, lines 132–156): `Finset.card_pos` turning the odd count into a concrete edge, plus the ±1 ↔ Bool sign-encoding transport in `tucker_one_dim_int`.
- `meta.json` left unchanged: already comprehensive (rich overview, 5 keyInsights, full conclusion) and under the size cap.

## Verification
- `pnpm build` succeeds (exit 0).
- annotations.json parses; 5 annotations, no range overlaps with existing ones.
- Axiom integrity unchanged: status verified, badge original, axiomCount 0 — matches the Lean file (0 sorries, 0 axioms).

PR #31423, opened 2026-06-28
Enrich The Companion Cosine Limit (lhopital-oq-05-oq-01)

rjwalters:
Enrichment pass for `lhopital-oq-05-oq-01` — the companion cosine limit (1-cos x)/x² → 1/2.

Expanded **annotation coverage 3 → 5** (the genuine gap; meta.json already meets all targets and is under the size cap, so it was left untouched):
- **ann-setup** (context, lines 3–31): the uncovered header — why the ratio is a second-order 0/0, the classical L'Hôpital-twice and half-angle routes, and the even-order-companion framing vs the cubic sine limit.
- **ann-eventual-filter** (technique, lines 60–67): the filter_upwards mechanism turning Real.cos_bound's |x|≤1 into an eventual hypothesis on 𝓝[≠] 0.

`pnpm build` green. Axiom integrity unchanged (verified, 0 axioms).

(Recovered onto a dedicated branch after a shared-branch force-push collision; supersedes the original head of #31391.)

### DIFF
--- a/src/data/proofs/lhopital-oq-05-oq-01/annotations.json
+++ b/src/data/proofs/lhopital-oq-05-oq-01/annotations.json
@@ -1,5 +1,17 @@
 [
   {
+    "id": "ann-setup",
+    "proofId": "lhopital-oq-05-oq-01",
+    "range": { "startLine": 3, "endLine": 31 },
+    "type": "context",
+    "title": "Why the Ratio Tends to 1/2: A Second-Order 0/0",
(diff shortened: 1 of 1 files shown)


rjwalters:
Closing as redundant: built on a stale worktree (base commit 0d8f7c2ebb0). origin/main already enriches lhopital-oq-05-oq-01 to 5 annotations via other work, so this PR's diff conflicts with and would revert the newer version. No action needed.

PR #31611, opened 2026-06-30
research(erdos-1003-oq-03): upper-density (limsup) layer — positive limsup ⇒ infinite [VERIFIED, 0-axiom]

rjwalters:
## Summary

Extends the Erdős #1003 OQ-03 density scaffolding (`Proofs.Erdos1003OQ03`) with the **upper density** — the unconditional carrier of the open question. The natural-density *limit* is not known to exist (its existence is itself part of OQ-03), so the `limsup` of the counting ratios `count(N)/(N+1)` is the correct object to reason about unconditionally.

## New results (0 sorry, 0 axiom, docker-verified)

| theorem | content |
|---|---|
| `upperDensity` | `limsup_N count(N)/(N+1)` — always exists (bounded sequence) |
| `upperDensity_nonneg`, `upperDensity_le_one` | upper density lies in `[0,1]` |
| `upperDensity_eq_of_hasNaturalDensity` | coincides with natural density when the limit exists (limsup of a convergent sequence) |
| `upperDensity_zero_of_finite` | finite solution set ⇒ upper density 0 |
| **`infinite_of_pos_upperDensity`** | **positive upper density ⇒ infinite** — strengthens `infinite_of_pos_density`: no need to assume the limit exists, only its limsup is positive |
| `hasNaturalDensity_unique`, `hasNaturalDensity_mem_Icc` | the natural density, when it exists, is unique and lies in `[0,1]` |

## Why this is progress

`infinite_of_pos_density` (existing) requires the density *limit* to exist and be positive. But existence of that limit is open. `infinite_of_pos_upperDensity` removes the existence hypothesis: since the `limsup` always exists, a positive *upper* density is a strictly weaker — hence more useful — sufficient condition for the (open) infinitude conjecture #1003.

## Honesty

Does **NOT** resolve OQ-03. The true density still requires the Erdős–Pomerance–Sárközy sparsity bound `#{n≤x: φ(n)=φ(n+1)} ≤ x/exp((log x)^{1/3})`, which is absent from Mathlib and cited only as external input. Only the standard foundational axioms (`propext`/`Classical.choice`/`Quot.sound`); no `native_decide`.

Lean file: 177 → 256 lines, 8 → 17 theorems, 3 → 4 defs.

### DIFF
--- a/proofs/Proofs/Erdos1003OQ03.lean
+++ b/proofs/Proofs/Erdos1003OQ03.lean
@@ -35,6 +35,15 @@ exact logical relationships that pin down what an eventual answer must supply.
      infinite worlds, and any *positive* density is equivalent to a strong form of
      the open #1003 conjecture.
 
+  5. **The upper density is the unconditional carrier**: the natural-density
+     *limit* may fail to exist (existence is itself open), but the upper density
+     `upperDensity = limsup_N count(N)/(N+1)` always exists for our bounded
+     sequence.  We show it lies in `[0,1]`, agrees with the natural density when
+     the latter exists, vanishes on finite solution sets, and — most usefully —
+     that a *positive upper density* already forces infinitude
+     (`infinite_of_pos_upperDensity`).  This strengthens (3): one no longer needs
+     the limit to exist, only its limsup to be positive.
+
 These are fully machine-checked (0 sorry, 0 axiom) consequences of the
 definitions.  The definitions agree verbatim with the parent #1003 entry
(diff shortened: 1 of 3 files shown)


rjwalters:
Superseded by #35128 (merged): the 8 genuinely-absent upper-density declarations recovered onto current main (2 dropped — already landed via sibling #31619), one v4.26 drift repair in upperDensity_nonneg (Filter.le_limsup_of_le signature change → le_limsup_of_frequently_le route). Docker-verified, all 4 headline #print axioms audits foundational-only. Part of the #35118 recovery queue.

PR #31628, opened 2026-06-30
research(van-der-waerden-first-moment-oq-01-oq-01): exact AP count via injectivity [VERIFIED, 0-axiom, original]

rjwalters:
## Summary

Answers the follow-up open question **oq-01-oq-01** of the van der Waerden first-moment chain: *is the parent entry's triangular AP-count bound an equality?*

The parent `van-der-waerden-first-moment-oq-01` proves only an **upper** bound

```
card_vdwFamily_le_sum : (vdwFamily n k).card ≤ ∑_{d=1}^{n} (n - (k-1)·d)
```

via `Finset.card_image_le`, which silently permits two distinct parameter pairs `(a, d)` to collapse onto the same arithmetic progression. This entry closes that gap.

## Result

For `k ≥ 2` the parameterisation `(a, d) ↦ vdwAP n a d k` is **injective** on the fitting box (`vdwAP_injOn`), so the triangular sum is the **exact** count:

```
card_vdwFamily_eq_sum (k : ℕ) (hk : 2 ≤ k) :
    (vdwFamily n k).card = ∑ d ∈ Finset.Icc 1 n, (n - (k - 1) * d)
```

Combined with the parent's telescoping bound `2(k-1)·(∑ …) ≤ n²`, the count is now pinned exactly: the parent's `≤ n²/(2(k-1))` is the **true value** of the parameter family, not merely an estimate.

## Proof idea

Pure membership, no `min'`/order machinery: the first two terms `a` (index 0) and `a+d` (index 1) of one progression both lie in the equal progression. The index-0 cross terms force `a = a'` (`Nat.mul_eq_zero` collapses the vanishing multiples over ℕ); the index-1 cross terms force `d = d'` (each step is a positive-integer multiple of the other). `k ≥ 2` is **sharp** — at `k = 1` every `(a, d)` maps to the singleton `{a}`, losing the step.

## Verification

- 5 theorems, 0 definitions, **0 sorries, 0 axioms**, no `native_decide`.
- `#print axioms` reports only `propext, Classical.choice, Quot.sound` for `card_vdwFamily_eq_sum` and `vdwAP_injOn`.
- Compiled via single-file `lake env lean` against the prebuilt Mathlib cache (Docker daemon was unavailable in this environment); dependency chain `PropertyBFirstMoment → VanDerWaerdenFirstMoment → …OQ01 → …OQ01OQ01` all built clean.

## Files

[...]

PR #31629, opened 2026-06-30
research(van-der-waerden-first-moment-oq-01): exact AP count via parametrization injectivity [VERIFIED, 0-axiom]

rjwalters:
## Summary

The entry `van-der-waerden-first-moment-oq-01` was already SOLVED (0 sorries, 0 axioms): it computes the exact **parameter-box** count `∑_{d=1}^{n}(n-(k-1)d)` and the factor-2 bound `2(k-1)·|family| ≤ n²`. The family count itself, however, was only an **upper bound** (`card_vdwFamily_le_sum`, via `Finset.card_image_le`).

This PR closes that gap: the parametrization `(a,d) ↦ vdwAP n a d k` is **injective** on the fitting box for `k ≥ 2`, so the bound becomes an **equality**:

$$|\{\text{length-}k\text{ APs in }[n]\}| \;=\; \sum_{d=1}^{n}\,(n-(k-1)d) \qquad\text{(card\_vdwFamily\_eq\_sum)}$$

**Why injective.** A fitting positive-step AP is strictly increasing, so as a `Finset (Fin n)` it determines its minimum point `↑a` and its maximum point `↑(a+(k-1)d)`. From the set one recovers `a` (the min) and, cancelling `k-1`, the step `d` — distinct fitting pairs give distinct APs.

## New theorems (6 → 13)

- `vdwAP_mem_first` / `vdwAP_mem_last` — the `i=0` and `i=k-1` endpoints are points of the AP.
- `vdwAP_first_le` / `vdwAP_le_last` — `↑a` is the minimum and `↑(a+(k-1)d)` the maximum of a fitting AP.
- `vdwFamily_param_injOn` — the parametrization is `Set.InjOn` on the fitting box (`k ≥ 2`).
- `card_vdwFamily_eq_sum` — **exact** family count (upgrades `≤` to `=`).
- `card_vdwFamily_ge` — crude exact lower bound `n-(k-1) ≤ |family|` (the `d=1` fiber), confirming the count is positive when `k-1 < n`.

## Verification

- `0` sorries, `0` axioms, no `native_decide`.
- `#print axioms` for `card_vdwFamily_eq_sum`, `vdwFamily_param_injOn`, `card_vdwFamily_ge` reports only `propext, Classical.choice, Quot.sound`.
- Built with the host Lean fallback (`./bin/lake env lean ...`) — Docker daemon was down this session.

## Honesty

This is elementary lattice-point reasoning: it pins the triangular sum as the *true* AP count rather than an overestimate. It does not touch the union-bound method or the exponential base of `W(k)`. A matching `Θ(n²/(k-1))` lower bound on `|family|` (to certify order-sharpness of the factor-2 upper bound) is left as a follow-up.

### DIFF
--- a/proofs/Proofs/Erdos1011OQ03.lean
+++ b/proofs/Proofs/Erdos1011OQ03.lean
@@ -23,6 +23,15 @@
      `f_five_le_f_four : f 5 n ≤ f 4 n` — the open value f_5(n) is bounded above
      by the now-known f_4(n).
 
+  1b. `forces_iff_f_le` — a complete characterization of the threshold:
+     `Forces r n m ↔ f r n ≤ m`. The forcing predicate is an up-set in the edge
+     bound (`forces_mono`), so `f r n` is exactly its least element.
+
+  1c. `f_eq_zero_of_lt` — the left boundary of the table: once `r` exceeds the
+     vertex count `n`, no `n`-vertex graph attains `χ ≥ r` (every such graph is
+     `n`-colourable, `chromaticNumber_le_card`), so the forcing condition is
+     vacuous and `f r n = 0`. The antitone-in-`r` value has bottomed out.
+
   2. The **vertex-shift pattern**. The shifts in the formulas above are
      0, 1, 3 for r = 2, 3, 4 — exactly C(r-1, 2). The pattern predicts a shift
      of C(4,2) = 6 for r = 5, i.e. a conjectured leading term ⌊(n-6)²/4⌋.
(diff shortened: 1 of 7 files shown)


rjwalters:
Closing as a duplicate of the open sibling PR #31628, which already proves the exact-count result `card_vdwFamily_eq_sum` (`|family| = ∑_{d=1}^{n}(n-(k-1)d)` via injectivity of the (a,d)↦AP parametrization) in the dedicated child entry `van-der-waerden-first-moment-oq-01-oq-01`.

#31628's child file `VanDerWaerdenFirstMomentOQ01OQ01.lean` imports this parent and declares `card_vdwFamily_eq_sum` in the same namespace `ProbMethod.VanDerWaerden`. Adding the same theorem to the parent here would cause a duplicate-declaration build failure once both land. My proof used a min/max-element characterization vs. #31628's term-by-term comparison, but the theorem is identical — no net new theory. Deferring to #31628 as the canonical home.

PR #31657, opened 2026-06-30
Fix: remove parent axiom-source files from child additionalFiles (axiom double-count)

rjwalters:
## Fix

Two gallery children were flagged by the auditor for `axiom undercount: claims 0, actual declarations N` because they listed their **parent** Lean files in `additionalFiles`. The parents declare axioms the children never use, so the auditor summed them against the children.

| Entry | Parent in additionalFiles | Parent axioms | Used by child? |
|-------|---------------------------|---------------|----------------|
| `erdos-1011-oq-03` | `Erdos1011Problem.lean` | 5 (Turán, Erdős–Gallai, Simonovits, Davies–Illingworth, HHKP) | No — comment-only reference |
| `collatz-structured-oq-01` | `CollatzStructured.lean` | 1 (`collatz_conjecture`) | No — comment-only reference |

## Evidence

**Before**: Auditor (`scripts/auditor/find-targets.ts --issues`) reported 2 issues — both "axiom undercount: claims 0, actual N".

**After**: Removed the parent files from `additionalFiles` in **both** the `meta.*` and `leanFile.*` blocks; kept the non-axiom shared lib `GraphCore.lean` for erdos-1011-oq-03. Each child's theorems are genuinely 0-axiom (no `^axiom ` in the child `.lean`; the parents only appear via `import` and prose comments). The parent gallery entries (`erdos-1011`, `collatz-structured`) already disclose these axioms with `leanFile.axiomCount` 5 and 1 respectively.

Auditor now reports `[]` (0 issues).

---

### DIFF
--- a/src/data/proofs/collatz-structured-oq-01/meta.json
+++ b/src/data/proofs/collatz-structured-oq-01/meta.json
@@ -6,9 +6,7 @@
   "meta": {
     "status": "verified",
     "proofRepoPath": "Proofs/CollatzStructuredOQ01.lean",
-    "additionalFiles": [
-      "Proofs/CollatzStructured.lean"
-    ],
+    "additionalFiles": [],
     "tags": [
       "number-theory",
       "collatz",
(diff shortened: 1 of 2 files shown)


rjwalters:
Duplicate of #31654 — identical diff (same two entries, same additionalFiles removals). Keeping the earlier #31654 to give the deployer a single clean PR for this fix.

PR #32197, opened 2026-07-01
research(erdos-728-incomplete-01): eliminate sorry + repair non-compiling file

rjwalters:
## Summary

Erdős #728 was labeled **SOLVED (proved in Lean)**, but `proofs/Proofs/Erdos728Problem.lean` **did not compile** on `main`. The tracked single `sorry` turned out to be only one of several defects, and the sorry as framed was not even dischargeable from the stated axiom.

## What was wrong & the fixes

1. **`erdos_728_exists` sorry was unsound as framed.** The theorem receives a *specific* `ε` with `0 < ε < 1/4`, but `erdos_728_resolution` was stated with an eventually-filter `∀ᶠ ε in 𝓝[>] 0`. You cannot extract existence at a *given* `ε` from an eventually-in-neighborhood statement, so the problem note's "just `exact` it out" plan was impossible.
   **Fix:** restated the resolution axiom in the pointwise form the theorem needs (`∀ ε, 0 < ε → ε < 1/4 → ...`). This is the *true* statement — the `b = n/2` construction satisfies `b > ε·n` for all `ε < 1/2`, so solutions exist throughout `(0, 1/4)`, not merely near `0`. **Axiom count unchanged (2).** The theorem is now a one-line direct application.

2. **`construction_size`:** `nlinarith` failed because `n ≥ 4` (a `ℕ` hypothesis) was never cast to `ℝ`. Added `(0:ℝ) < n` and a `mul_pos` hint.

3. **Two dangling `/--` doc-comments** ("Divisibility via Legendre", "General Principle") preceded `/-` comment blocks rather than declarations → hard parse errors. Changed to plain `/-` comments.

4. **`log n` was ambiguous everywhere** (`open Real` and `open Nat` both export `log`). Qualified all real-code uses as `Real.log`.

## Result

- Compiles cleanly under `./proofs/scripts/docker-build.sh Proofs.Erdos728Problem` (1857 jobs; only an unused-variable warning remains).
- **0 sorries**, **2 axioms** (`erdos_1968_bound`, `erdos_728_resolution` — both deep results, legitimately axiomatized).
- `meta.json` sorry counts updated 1 → 0; status remains `axiomatized`.

## Honesty note

This does not *prove* Erdős #728 — the resolution and the 1968 bound remain stated axioms. The value here is (a) eliminating a sorry that was actually underivable, and (b) making a file that claimed to be verified-in-Lean actually compile.

### DIFF
--- a/proofs/Proofs/Erdos1018Problem.lean
+++ b/proofs/Proofs/Erdos1018Problem.lean
@@ -169,8 +169,8 @@ theorem erdos_1018_solved : erdos_1018_question := by
   use S
   constructor
   · exact hS
-  · -- K₅ subdivision implies non-planarity
-    sorry
+  · -- K₅ subdivision implies non-planarity (Kuratowski's theorem, right-to-left).
+    exact (kuratowski_theorem (inducedSubgraph G S)).mpr (Or.inl hSub)
 
 /-
 ## The Constant C_ε Grows as ε → 0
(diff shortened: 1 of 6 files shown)


rjwalters:
Closing as **superseded** — automated race-condition cleanup.

This PR adds `proofs/Proofs/Erdos1018Problem.lean`, but that proof file already exists on `main` (landed via `e55765dea1c research(erdos-1018): extract axiom-free crossover lemma + dual corollary [VERIFIED core]`). main's version contains an equal-or-greater set of declarations, so this branch is an unmergeable add/add duplicate of an already-formalized result.

Two agents worked the same problem concurrently; this one lost the merge race. Structural fixes (pre-submission existence guard + claim heartbeat) are being added to prevent recurrence. If you believe this version carries unique content, reopen and rebase onto main.

PR #32637, opened 2026-07-02
Enrich Gauss Sum Bound |τ(χ)|=√p: add missing index.ts

rjwalters:
Enrichment pass for `bounded-prime-gaps-oq-04-oq-03` ("The Gauss Sum Bound |τ(χ)| = √p: Prime-Modulus Case from Mathlib").

## Critical fix
- **Created `index.ts`** — the entry had none (only meta.json + annotations.json). Without it Vite's `import.meta.glob` cannot discover the proof, so the page renders "proof not found" on the live site even though the entry appears in the gallery listing.

Annotations and sections already cover the full Lean file; the only gap was the missing entry point. Verified with `pnpm build` (✓ built).

### DIFF
--- a/src/data/proofs/bounded-prime-gaps-oq-04-oq-03/index.ts
+++ b/src/data/proofs/bounded-prime-gaps-oq-04-oq-03/index.ts
@@ -1,0 +1,36 @@
+import type { Proof, Annotation, ProofData, ProofMeta, ProofSection, ProofOverview, ProofConclusion, CrossReference } from '@/types/proof'
+import metaJson from './meta.json'
+import annotationsJson from './annotations.json'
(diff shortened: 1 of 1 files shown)


rjwalters:
Closing: this PR only added a per-proof `index.ts` shim, which I now confirm is **dead legacy code**. Since issue #20992/#20993 (phase 2 build-perf fix), the runtime loader in `src/data/proofs/index.ts` fetches proof data by slug from the build-generated `public/data/proofs/<slug>/` tree (produced from meta.json + annotations.json by scripts/annotations/build.ts), and no longer imports per-proof index.ts shims. 1296 live gallery entries have no index.ts and load fine. The role-doc claim that a missing index.ts causes 'proof not found' is stale. No enrichment gap remains on this entry (sections + annotations already span the full Lean file), so there is nothing to ship.

Issue #20993, opened 2026-05-29
Build perf phase 2: move meta/annotations to public/, add hashed manifest, delete per-proof shims


rjwalters:
Phase 2 of the durable fix for #20984 (architect proposal: https://github.com/rjwalters/lean-genius/issues/20984#issuecomment-4577740690). **Blocked by phase 1 (#20992)** — do not start until phase 1 has landed and proven out the static-data + fetch pattern.

## Goal
Collapse the remaining ~7300 data modules (2435 `index.ts` shims + 2435 `meta.json` + 2435 `annotations.json`) to a few hundred app modules by moving JSON to `public/` and deleting the `import.meta.glob` machinery. Returns the build toward its ~35s baseline and makes it O(1) in gallery size.

## Implementation (per architect proposal §1–3)
1. **Emit step**: copy each proof's `meta.json` + `annotations.json` → `public/data/proofs/<slug>/` (alongside `source.lean` from phase 1). Compute sha8 hashes.
2. **Hashed manifest**: emit `src/data/proofs/data-manifest.json` = `{ slug: { meta: <sha8>, ann: <sha8>, src: <sha8> } }`. This is the ONLY per-proof data still imported into the graph (one small module). Bundled with content hashing, so any proof change busts the importing chunk.
3. **Rewrite `getProofAsync`** to fetch `meta.json`/`annotations.json`/`source.lean` by slug with `?v=<sha8>` query params from the manifest (see proposal loader sketch). Keep the public signature.
4. **Delete** the ~2435 per-proof `src/data/proofs/<slug>/index.ts` shims and the `import.meta.glob('./**/index.ts')` + `proofModules` map. Also delete dead code `getProof`/`getAllProofs` (0 external consumers per architect grep).
5. `listings.json` **stays an in-graph static import** (1.8MB, single module, needed eagerly by HomePage/ErdosPage/ProfileModal — not part of the O(N) blowup).

## Acceptance
- [ ] `pnpm build` `vite build` stage back near ~35s baseline; report wall-clock.
- [ ] Module count in the build is a few hundred, not ~12K (vite/rollup output or `--debug` confirms).
- [ ] Proof pages render meta + annotations + source — browser smoke test.
- [ ] Cache-busting verified: changing one proof's meta.json changes its `?v=` and the manifest chunk hash.
- [ ] `getProof`/`getAllProofs`/`proofModules` removed; no remaining `import.meta.glob` for proofs.

## Follow-up
[...]

PR #32734, opened 2026-07-02
Enrich centered-hexagonal-sum-oq-01: add missing index.ts

rjwalters:
The entry `centered-hexagonal-sum-oq-01` had `meta.json` and `annotations.json` but was **missing `index.ts`**, so Vite's `import.meta.glob` could not discover it and the proof page rendered "proof not found" on the live site. This adds the standard Vite entry point so the proof loads. No content changes.

### DIFF
--- a/src/data/proofs/centered-hexagonal-sum-oq-01/index.ts
+++ b/src/data/proofs/centered-hexagonal-sum-oq-01/index.ts
@@ -1,0 +1,36 @@
+import type { Proof, Annotation, ProofData, ProofMeta, ProofSection, ProofOverview, ProofConclusion, CrossReference } from '@/types/proof'
+import metaJson from './meta.json'
+import annotationsJson from './annotations.json'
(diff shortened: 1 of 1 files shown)


rjwalters:
Closing: this PR only added a per-proof `index.ts` shim, which is **obsolete since #20993**. The gallery no longer discovers proofs via `import.meta.glob` over per-proof index.ts modules — that O(N) shim graph was deliberately removed to fix the >20-min build blowup. Proofs are now loaded by `src/data/proofs/index.ts` (`getProofAsync`) via `data-manifest.json` (build-generated from each entry's meta.json) + runtime fetch from `public/data/proofs/<slug>/`. This entry is already present in the manifest and loads correctly without any index.ts. Adding index.ts re-introduces the exact anti-pattern #20993 removed. My mistake — I followed a stale instruction in the enricher role doc.

PR #33943, opened 2026-07-03
pipeline: stop same-problem PR races producing add/add duplicates

rjwalters:
## Problem

The research pipeline had collected a **DIRTY-PR backlog** — 50 open PRs, 40 conflicting, only 3 CLEAN. Root cause diagnosis: **34 of the 40 DIRTY PRs were add/add duplicates** — the proof file the PR adds *already exists on `main`*, landed by a competing PR that won the merge race. Two agents worked the same problem concurrently; the loser's branch rots forever because no rebase can resolve an add/add collision on the same new-file path.

The races come from claim TTL expiry: `claim-random` is already atomic + randomized-within-tier, so agents don't grab the *same* problem simultaneously. But when a claim expires mid-session (default TTL 90 min, easily exceeded by Docker builds + Mathlib-drift repair), the Seeker re-serves the problem and a second agent duplicates the work.

This is wasted effort, not throughput.

## Fixes

**A. Deployer reaping** (`sync-and-deploy.sh`) — before attempting a doomed rebase on a CONFLICTING PR, check whether every proof file it adds already exists on main. If so, close it as superseded instead of burning rebase cycles on it every deploy loop. Self-healing backlog drain.

**B. Pre-submission guard** (new `scripts/research/check-superseded.sh` + researcher role **Step 5.5**) — the researcher runs the detector before `gh pr create`; if the proof file already landed on main it releases the claim and skips the PR. Stops races from becoming PRs at all.

**C. Claim heartbeat + TTL** (`claim-problem.sh`) — raise default TTL 90 → 180 min, and add a `heartbeat` subcommand (alias of `extend`) that healthy long sessions call periodically so a live claim is never re-served to a second agent.

## The shared detector

`check-superseded.sh` distinguishes true add/add duplicates from salvageable work:
- **SUPERSEDED** — every proof file the branch *adds* (absent at merge-base) already exists on main → reap/abort.
- **NOT_SUPERSEDED** — adds ≥1 genuinely new file, or only modifies existing files → proceed.

Uses merge-base to tell "added" from "modified", so a legitimate follow-up that edits main's existing file is never reaped. bash 3.2-compatible (macOS).

Verified against real branches:
- `research/bell-egf-oq0101` (file on main) → `SUPERSEDED`
- `research/derangements-oq0401-a000255-quotient` (1 new + 1 colliding file) → `NOT_SUPERSEDED`

## Manual cleanup already done

Closed 21 provably-superseded DIRTY PRs (main's version equal-or-richer), verified per-PR by comparing declaration sets. Backlog: **50 → 39 open, DIRTY 40 → 9.** 13 net-richer PRs held for a separate salvage decision (see below).

## Not included (fix D — rebalance)

D was scoped as an operational knob. Analysis showed the claim strategy is already atomic + randomized-within-tier, so the collisions were TTL-driven (fixed by C), **not** a claim-strategy defect. Delivering D as a recommendation rather than a risky live-fleet/claim-strategy change: keep the available-pool floor ≥ ~2× the running researcher count so depth-first claiming has breadth to spread across, and avoid over-provisioning researchers beyond pool depth. Happy to encode a specific floor if you want it.

## Test plan

- [x] `bash -n` on all three scripts
- [x] detector correct on SUPERSEDED and NOT_SUPERSEDED real branches
- [x] `heartbeat` alias reaches `extend_claim`
- [ ] deployer dry-run reaping on next deploy cycle

### DIFF
--- a/.lean/roles/researcher.md
+++ b/.lean/roles/researcher.md
@@ -49,10 +49,21 @@ while true:
     2. Claim a problem from the candidate pool
     3. Run one research iteration (following the research skill)
     4. Commit meaningful progress
-    5. Create a PR with findings
+    5. Create a PR with findings (run the Supersession Guard first — Step 5.5)
     6. Update problem status and knowledge
     7. Release claim
     8. Repeat
+```
+
+**Heartbeat your claim during long sessions.** A single iteration (Docker builds,
+Mathlib-drift repair, Aristotle round-trips) can run well over an hour. If your
+claim expires mid-session the Seeker re-serves the problem and a second agent
+duplicates your work — the loser's PR then rots as an unmergeable add/add
+duplicate. So whenever a step takes a while (after each build cycle, or roughly
+every ~30 min), refresh the claim:
+
+```bash
+$REPO_ROOT/scripts/research/claim-problem.sh heartbeat $PROBLEM_ID
 ```
 
 ### Checking Signals
@@ -534,6 +545,31 @@ Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>
 EOF
 )"
 git push -u origin $(git branch --show-current)
+```
+
+## Step 5.5: Supersession Guard (MANDATORY before every PR)
+
+Concurrent agents sometimes work the **same** problem — a claim expires mid-session
+and a second agent re-claims it, or two agents pick overlapping OQ extensions. Both
+create a proof file at the same `proofs/Proofs/*.lean` path. Whoever merges first
+wins; the loser becomes an unmergeable add/add duplicate that rots as a DIRTY PR.
+That is wasted effort. **Do not open such a PR.**
+
+Run the guard before creating the PR:
+
+```bash
+verdict=$("$REPO_ROOT/scripts/research/check-superseded.sh" --base origin/main --quiet | tail -1)
+if [[ "$verdict" == "SUPERSEDED" ]]; then
+  echo "$(date +%H:%M): ABORT — $PROBLEM_ID already formalized on main (add/add). Not opening PR." \
+    >> "$REPO_ROOT/.loom/logs/$RESEARCHER_ID.actions.log"
+  # Your proof file already exists on main. Do NOT open a duplicate PR.
+  # Instead: either (a) release the claim and pick a genuinely open problem, or
+  # (b) if your version proves strictly MORE than main's, rebase onto origin/main
+  #     and reconcile into main's existing file so the PR is a real superset.
+  $REPO_ROOT/scripts/research/claim-problem.sh release $PROBLEM_ID
+  exit 0
+fi
+# verdict is NOT_SUPERSEDED or NO_PROOF_FILES → safe to proceed.
 ```
 
 ## Step 6: Create Pull Request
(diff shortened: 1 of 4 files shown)


rjwalters:
✅ **Approved!**

Reviewed all four focus areas. The core logic is correct, well-tested, and the auto-close is properly guarded. Merge state is CLEAN / MERGEABLE; no CI checks are configured for this scripts/docs branch, and none are failing.

## 1. `check-superseded.sh` correctness — verified

Reconstructed each verdict in a minimal git repo and confirmed the merge-base logic distinguishes add/add duplicates from salvageable work exactly as claimed:

| Scenario | Verdict | Exit |
|----------|---------|------|
| Pure add/add (every added proof file already on main) | `SUPERSEDED` | 3 |
| 1 colliding + 1 genuinely new added file | `NOT_SUPERSEDED` | 0 |
| Modifies an existing main file only | `NOT_SUPERSEDED (modifies existing files only)` | 0 |
| No proof files touched | `NO_PROOF_FILES` | 0 |

The `git cat-file -e ${base_commit}:${f}` merge-base gate correctly classifies "added on this branch" vs "modified", so a legitimate follow-up that edits main's existing file is never reaped. The `--quiet | tail -1` caller form yields a clean single verdict word. No false-positive path found for genuinely-new work. The only residual case — a branch adding a *different* proof at a filename that coincidentally matches an unrelated file on main — is an add/add conflict that is unmergeable anyway, so closing it is defensible and the close comment invites recovery.

## 2. Deployer auto-close safety — well-guarded

The `gh pr close` is gated by, in order: only the `mergeable == "CONFLICTING"` loop reaches it; `eligible_prs` already excludes drafts and `loom:review-requested` PRs; the detector must return the exact string `"SUPERSEDED"`; and non-research PRs (no `proofs/Proofs/*.lean`) short-circuit to `NO_PROOF_FILES` and are never closed. It only ever closes PRs that are already unmergeable. Good.

## 3. Shell correctness / bash 3.2 — verified

Local default is bash 3.2.57; all three scripts pass `bash -n`. No `mapfile` (read-loop used instead), `declare -a`, process substitution, and both array iterations are guarded against the bash-3.2 `set -u` empty-array trap. The `sup_verdict=$(... | tail -1 || echo "")` form in the deployer correctly protects the assignment from `set -e` while still capturing the verdict.

## 4. TTL / heartbeat — no locking regression

`extend|heartbeat` both dispatch to `extend_claim`, which is ownership-guarded (`current_agent != AGENT_ID` → refuse), so a heartbeat cannot extend or steal another agent's claim. The 90→180 TTL doubling is a documented tradeoff (a dead claim now frees in 180 min) mitigated by heartbeat keeping live sessions claimed. `CLAIM_TTL` env override preserved.

## Non-blocking follow-ups (not required for merge)

1. **`--delete-branch` vs "reopen and rebase"** — the close uses `--delete-branch`, but GitHub disables *Reopen* once the head branch is deleted, so the comment's "reopen and rebase" advice isn't literally actionable. The commits remain accessible via the closed PR for cherry-pick, and SUPERSEDED means the work is already fully on main, so impact is minimal. Consider dropping `--delete-branch` or rewording the comment to "recreate the branch from this PR's commit".
2. **Deployer `FETCH_HEAD` staleness** — `git fetch origin "$branch" ... || true` swallows failures; if a fetch ever failed, `FETCH_HEAD` would retain a prior iteration's ref. Open-PR branches fetch reliably so this is latent, but hardening to skip the reap when the fetch fails would remove the hazard.
3. **Researcher Step 5.5 defensiveness** — the snippet uses `verdict=$(... --quiet | tail -1)` without the `|| echo ""` guard the deployer uses. Under an agent shell with `set -o pipefail`, the SUPERSEDED exit code 3 through the pipe could abort before the `if`. Mirroring the deployer's defensive form would be safer.

Solid, well-scoped infrastructure fix that drains a real DIRTY-PR backlog and prevents future same-problem races. Nice work.

PR #33951, opened 2026-07-03
Enrich Cantor Normal Form via Transfinite Induction: references + focused annotations

rjwalters:
Enrichment pass for `mathematical-induction-oq-01-oq-01` ("Cantor Normal Form via Transfinite Induction").

**References 0 → 3.** The entry had no references despite being a classical topic. Added the canonical sources: Cantor's original *Beiträge* (1897, where the ordinal normal form originates), Sierpiński's *Cardinal and Ordinal Numbers* (1958), and Jech's *Set Theory* (2003, standard modern CNF existence/uniqueness by transfinite induction).

**Annotation coverage 4 → 5 (de-bundling).** The final annotation previously bundled two distinct theorems over lines 78–88. Split into one annotation per theorem: `cnf_exponent_le_log` (leading exponent = log ω o, the base-ω degree, lines 78–83) and `cantor_normal_form_zero` (base case CNF ω 0 = [], lines 85–88). Each now has a focused title and mathContext; no content lost.

- No existing content removed; annotation ranges remain ascending and non-overlapping.
- meta.json ~14.7 KB, well under caps.
- JSON validated with a parser.

### DIFF
--- a/src/data/proofs/mathematical-induction-oq-01-oq-01/meta.json
+++ b/src/data/proofs/mathematical-induction-oq-01-oq-01/meta.json
@@ -119,6 +119,29 @@
       "mathContext": "$p \\in \\mathrm{CNF}\\,\\omega\\,o \\Rightarrow p_1 \\le \\log_\\omega o;\\qquad \\mathrm{CNF}\\,\\omega\\,0 = [\\,].$"
     }
   ],
+  "references": [
+    {
+      "authors": "Cantor, Georg",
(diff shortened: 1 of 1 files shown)


rjwalters:
Deployer: this PR conflicts with main because the same proof's annotations were independently re-enriched/restructured on main after this branch was created. This is an editorial divergence, not a mechanical merge, so the deployer is not auto-resolving it. Please rebase on latest main and reconcile the annotation entries, then it can be merged.

Issue #34556, opened 2026-07-04
deployer superseded-close comment advises 'reopen and rebase' but --delete-branch makes it un-actionable


rjwalters:
**Source:** Loom Judge review of #33943 (non-blocking finding a).

The deployer's superseded-PR auto-close (\`scripts/deploy/sync-and-deploy.sh\`) runs \`gh pr close "\$pr" --delete-branch\` with a comment that says: *"If you believe this version carries unique content main lacks, reopen and rebase onto main."*

But \`--delete-branch\` deletes the head branch, so GitHub's "Reopen" button is disabled once the branch is gone — the advice is literally un-actionable via the UI. (Recovery requires re-pushing the branch from a local copy first.)

**Fix (pick one):**
- Drop \`--delete-branch\` so reopen works (leaves stale branches to be pruned separately), **or**
- Reword the comment to match reality, e.g. *"This branch has been deleted; to salvage unique content, re-push it and open a fresh PR rebased onto main."*

**Impact:** cosmetic/UX — the reaped PRs are genuine add/add duplicates so this path is rarely needed, but the misleading instruction should be corrected.

Issue #34555, opened 2026-07-04
deployer reap: stale FETCH_HEAD if 'git fetch origin $branch' fails


rjwalters:
**Source:** Loom Judge review of #33943 (non-blocking finding b).

In the deployer reap block (\`scripts/deploy/sync-and-deploy.sh\`, CONFLICTING loop, ~line 349):

\`\`\`bash
git fetch origin "\$branch" --quiet 2>/dev/null || true
sup_verdict=\$("\$REPO_ROOT/scripts/research/check-superseded.sh" --ref FETCH_HEAD --base origin/main --quiet | tail -1 || echo "")
\`\`\`

If the \`git fetch\` fails (network blip, deleted remote branch), the \`|| true\` swallows it, and the subsequent \`check-superseded.sh --ref FETCH_HEAD\` compares against a **stale FETCH_HEAD** left by an earlier fetch of a *different* branch. That could yield a wrong SUPERSEDED/NOT_SUPERSEDED verdict for the current PR.

**Fix:** capture the fetch result and skip the reap check for this PR if the fetch failed (e.g. \`git fetch origin "\$branch" --quiet 2>/dev/null || { print_warning "fetch failed for #\$pr; skipping reap check"; continue-to-rebase; }\`), or fetch into an explicit ref (\`git fetch origin "\$branch":refs/tmp/reap-check\`) and reference that instead of FETCH_HEAD.

**Impact:** rare (fetch rarely fails), but a wrong verdict could either skip a real duplicate or — worse — flag a legitimate PR. Worth hardening since the reap auto-closes PRs.

PR #34558, opened 2026-07-04
pipeline: harden anti-churn reap/guard (Judge follow-ups for #33943)

### LINKED ISSUE #34554: researcher Step 5.5 supersession guard can abort early under set -o pipefail
**Source:** Loom Judge review of #33943 (non-blocking finding c).

The pre-submission supersession guard added to \`.lean/roles/researcher.md\` (Step 5.5) runs:

\`\`\`bash
verdict=\$("\$REPO_ROOT/scripts/research/check-superseded.sh" --base origin/main --quiet | tail -1)
\`\`\`

The deployer's equivalent call guards the pipeline with \`| tail -1 || echo ""\`. The researcher snippet omits it. Under \`set -o pipefail\` (which research agent shells may enable), if \`check-superseded.sh\` exits non-zero (e.g. exit 3 = SUPERSEDED, or a git error), the assignment inherits that failure and — combined with \`set -e\` — could abort the researcher mid-session before it even evaluates \`\$verdict\`.

**Fix:** mirror the deployer's guard — append \`|| echo ""\` (or capture exit code explicitly), so the verdict string is always assigned and the \`if\` branch does the deciding.

**Impact:** low-frequency but real — an early abort skips the PR entirely for that session. One-line hardening.

rjwalters:
Sweeps the three non-blocking findings from the Loom Judge review of #33943 into one hardening PR.

## Fixes

**#34554 — researcher guard could abort early** (`.lean/roles/researcher.md` Step 5.5)
The pre-submission supersession guard now uses `| tail -1 || echo ""`, matching the deployer. Without it, a non-zero exit from `check-superseded.sh` (exit 3 = SUPERSEDED, or a git error) could abort the research session under `set -o pipefail` + `set -e` before the `if` ran. Now `$verdict` is always assigned and the `if` does the deciding.

**#34555 — deployer reap could compare the wrong tree** (`scripts/deploy/sync-and-deploy.sh`)
The reap now fetches the PR head into an explicit ref (`refs/tmp/reap-check-$pr`) and only runs the supersession check when that fetch succeeds, cleaning the ref up afterward. Previously `git fetch ... || true` swallowed failures, leaving `FETCH_HEAD` pointing at a *previous* branch — so `check-superseded --ref FETCH_HEAD` could mis-verdict a PR. Since the reap auto-closes PRs, a wrong verdict is worth preventing. On fetch failure it now falls through to the normal rebase path with a warning.

**#34556 — misleading close comment** (`scripts/deploy/sync-and-deploy.sh`)
The superseded-close comment no longer says "reopen and rebase" (un-actionable — `--delete-branch` disables GitHub's Reopen button). It now states the branch was deleted and to re-push + open a fresh PR rebased onto main.

## Verification
- `bash -n scripts/deploy/sync-and-deploy.sh` passes.
- The fetch-into-ref → `check-superseded.sh` → `update-ref -d` cleanup sequence works end-to-end against a live remote branch.

Closes #34554, #34555, #34556.

### DIFF
--- a/.lean/roles/researcher.md
+++ b/.lean/roles/researcher.md
@@ -558,7 +558,11 @@ That is wasted effort. **Do not open such a PR.**
 Run the guard before creating the PR:
 
 ```bash
-verdict=$("$REPO_ROOT/scripts/research/check-superseded.sh" --base origin/main --quiet | tail -1)
+# The `|| echo ""` guard matters: check-superseded.sh exits non-zero on a
+# SUPERSEDED verdict (exit 3) or a git error, which under `set -o pipefail` +
+# `set -e` would otherwise abort your session here before the `if` runs. The
+# guard ensures $verdict is always assigned so the `if` below does the deciding.
+verdict=$("$REPO_ROOT/scripts/research/check-superseded.sh" --base origin/main --quiet | tail -1 || echo "")
 if [[ "$verdict" == "SUPERSEDED" ]]; then
   echo "$(date +%H:%M): ABORT — $PROBLEM_ID already formalized on main (add/add). Not opening PR." \
     >> "$REPO_ROOT/.loom/logs/$RESEARCHER_ID.actions.log"
--- a/scripts/deploy/sync-and-deploy.sh
+++ b/scripts/deploy/sync-and-deploy.sh
@@ -355,23 +355,33 @@ merge_prs() {
         # same-problem race between two agents. No rebase can ever land it; close
         # it so the backlog drains instead of accumulating dead duplicates.
         if [[ -x "$REPO_ROOT/scripts/research/check-superseded.sh" ]]; then
-            git fetch origin "$branch" --quiet 2>/dev/null || true
-            local sup_verdict
-            sup_verdict=$("$REPO_ROOT/scripts/research/check-superseded.sh" \
-                --ref FETCH_HEAD --base origin/main --quiet 2>/dev/null | tail -1 || echo "")
-            if [[ "$sup_verdict" == "SUPERSEDED" ]]; then
-                if $DRY_RUN; then
-                    echo "  Would close #$pr as superseded (proof file already on main)"
-                else
-                    print_warning "Closing #$pr as superseded — proof file already on main (add/add duplicate)"
-                    gh pr close "$pr" --delete-branch --comment "Closing as **superseded** — automated race-condition cleanup by the deployer.
+            # Fetch the PR head into an explicit ref rather than relying on
+            # FETCH_HEAD (issue #34555): a swallowed fetch failure would otherwise
+            # leave FETCH_HEAD pointing at a *previous* branch, so the check would
+            # silently compare the wrong tree. Only run the reap check when this
+            # fetch actually succeeds; on failure, fall through to the rebase path.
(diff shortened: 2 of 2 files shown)


rjwalters:
## Loom Judge Review — APPROVED

Reviewed the three Judge follow-ups from #33943. All correct; approving.

### 1. Temp-ref cleanup on the SUPERSEDED `continue` path — verified NO leak
This was the highest-risk item. Traced the control flow:

```
if git fetch origin "$branch:$reap_ref" --force ...; then
    sup_verdict=$(check-superseded --ref "$reap_ref" ... | tail -1 || echo "")
    git update-ref -d "$reap_ref" 2>/dev/null || true   # <-- runs first, unconditionally
    if [[ "$sup_verdict" == "SUPERSEDED" ]]; then
        ... gh pr close ...
        continue                                          # <-- cleanup already done above
    fi
fi
```

`git update-ref -d` is placed *before* the `if SUPERSEDED` test, so it executes on every branch where the ref was created (SUPERSEDED-close, non-superseded-fallthrough). The temp ref is **not** leaked on the `continue` path. Good placement.

### 2. Fetch-failure fallthrough — correct
On fetch failure the `else` runs (`print_warning ... proceeding to rebase`) and control falls past the `if [[ -x ... ]]` block to `print_info "Attempting rebase..."`. The ref was never created on this path, so nothing to clean. Fallthrough reaches the rebase path as intended.

### 3. Ref-name safety — safe
`refs/tmp/reap-check-$pr` where `$pr` is an integer PR number from `jq -r '... | .number'`. Not attacker-controlled free text; always a valid ref component. `--force` defensively overwrites any ref orphaned by a prior interrupted run.

### 4. Shell correctness — clean
- `set -euo pipefail` is active (line 47), so the guards matter. The `git fetch` is used as an `if` condition, exempting it from `set -e`. The command substitution ends in `|| echo ""` so a non-zero `check-superseded` (exit 3 = SUPERSEDED) under `pipefail` can't abort. `update-ref -d ... || true` is guarded.
- `"$branch:$reap_ref"` is quoted as one refspec argument; git refnames cannot contain `:`, so the split is unambiguous even for slash-containing branch names.
- bash 3.2 compatible (only `local`, `[[ ]]`, `$(...)`, `continue`; no bash-4 constructs). `bash -n` passes.
- `check-superseded.sh` accepts `--ref <ref>` (default HEAD) and resolves it via `git merge-base`/`git diff`, so passing the explicit `refs/tmp/...` ref works.

### 5. researcher.md `|| echo ""` — fully resolves the early-abort concern
Under `set -euo pipefail`, `check-superseded.sh` exit 3 would make the pipeline fail and `set -e` abort the session before the `if`. `... | tail -1 || echo ""` forces the command substitution to exit 0, so `$verdict` is always assigned and the `if` does the deciding. Matches the deployer pattern. Resolved.

### Minor (non-blocking)
- The fetch-failure warning string has two leading spaces (`"  Could not fetch ..."`) passed to `print_warning`, a cosmetic indentation inconsistency vs sibling `print_warning` calls. Harmless.

Merge state CLEAN, MERGEABLE. `bash -n` passes. No CI checks are registered for this branch (shell + docs only, no triggering workflow), consistent with the change surface.

Approving.

PR #34723, opened 2026-07-04
Research: ballot-oq-04-oq-02-oq-01 s10 — glued partition is non-crossing (isNonCrossingFp_glueFp, verified, 0 new sorry)

rjwalters:
## Summary

Proves **`isNonCrossingFp_glueFp`**: gluing two non-crossing partitions of the windows `Fin m` and `Fin (n-m)` produces a non-crossing partition of `Fin (n+1)`. This is the injectivity-side structural fact the inverse (gluing) map `glueFp` needs to land in the non-crossing subtype `{P // IsNonCrossingFp P}` — one of the three pieces required to discharge the file's sole open `sorry`, `nonempty_firstReturnEquiv` (the first-return bijection).

**Verified** — docker-built clean (`Proofs.BallotProblemOQ04OQ02OQ01`, 7745 jobs, exit 0). Purely additive (+154 lines): the sorry count is unchanged at **1**. Rebased onto current `main` (which merged the parallel **s11** glueLabel evaluation API); no name clashes.

### New lemmas
- `ncf_part_eq` — `IsNonCrossingFp` restated in `Finpartition.part`-equality form.
- `glueLabel_of_gt`, `glueLabel_of_pos_le`, `glueLabel_of_zero` — closed forms of the glue label in window B, window A (nonzero point), and at the special point `0` (variable form).
- `glueLabel_isLeft_of_le`, `glueLabel_le_iff` — window determination: a shared `glueLabel` forces both points to the same side of the cut `m`.
- `isNonCrossingFp_glueFp` — the main result.

### Proof idea
Window-`A` points carry `Sum.inl` labels, window-`B` points `Sum.inr`, so a same-label pair lies in one window, and the linear order `a<b<c<d` forces **all four** of a crossing candidate into a single window. Window `B` is the shifted `P₂` directly. Window `A` is the shifted `P₁` **with point `0` attached to the top block `m-1`** — and that "attach to the top" choice is precisely what keeps `A` non-crossing: a crossing through `0` pairs `(0,c)` with `(b,d)` at `0<b<c<d≤m`, and `0 ~ m-1` turns it into the `P₁`-crossing `(b-1,c-1,d-1,m-1)`, whose resolution forces `0 ~ b`.

### Remaining toward the bijection
The two round-trip laws (`firstReturnForward ∘ glueFp = id`, `glueFp ∘ firstReturnForward = id`), then assemble the `Equiv` — these can reuse main's s11 evaluation API (`glueLabel_offsetEmb_left/right`, `mem_part_zero_glueFp_left`).

### DIFF
--- a/proofs/Proofs/BallotProblemOQ04OQ02OQ01.lean
+++ b/proofs/Proofs/BallotProblemOQ04OQ02OQ01.lean
@@ -538,6 +538,160 @@ theorem mem_part_zero_glueFp_left {n : ℕ} (m : ℕ) (hm : m ≤ n) (hm0 : 0 < 
   · intro h
     exact ((P₁.mem_part_iff_part_eq_part (mem_univ _) (mem_univ _)).mp h).symm
 
+/-! ### The glued partition is non-crossing (injectivity-side structural fact)
+
+The gluing map `glueFp` must land in the *non-crossing* subtype for it to serve as the inverse
+of `firstReturnForward`. This section proves exactly that: gluing two non-crossing partitions of
+the windows `Fin m` and `Fin (n-m)` yields a non-crossing partition of `Fin (n+1)`.
+
+The proof is a case split by the window a straddle-free block lands in. Since window-`A` points
+carry `Sum.inl` labels and window-`B` points carry `Sum.inr` labels, a same-label pair lies in a
+single window, and the linear order `a < b < c < d` forces *all four* of a crossing candidate
+into one window (`glueLabel_le_iff`). Inside window `B` the labels are literally the shifted `P₂`,
+so non-crossing transports directly. Inside window `A` the labels are the shifted `P₁` **with
+point `0` attached to the top block `m-1`** — and it is precisely this "attach to the top" choice
+that keeps `A` non-crossing: a crossing through `0` would pair `(0, c)` with `(b, d)` at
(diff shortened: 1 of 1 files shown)


rjwalters:
✅ **Build resolved — verified.** The Azure mathlib-cache CDN recovered and `Proofs.BallotProblemOQ04OQ02OQ01` now docker-builds clean (exit 0, 7745 jobs); the only remaining `sorry` is the pre-existing `nonempty_firstReturnEquiv` (count unchanged at 1). One boundary case (`d.val = m`) needed a fix — `Fin.ext` left the `Fin.val` subterms as opaque atoms omega couldn't reduce; switched to `simp only [Fin.mk.injEq]; omega` + a `congrArg`/`trans` closure. `isNonCrossingFp_glueFp` is now machine-checked. Removing the build-pending draft status.

rjwalters:
Added follow-up commit `648a28d5fd4` (s12): **`firstBlockMax_glueFp_val`** — proves `(firstBlockMax (glueFp m hm P₁ P₂)).val = m`, i.e. the forward map recovers the cut index `m` from a glued partition.

This is the round-trip linchpin for the `right_inv` law of `nonempty_firstReturnEquiv`: it pins the forward map's two offset windows `[1,m]` and `[m+1,n]` back to the sizes `(m, n-m)` that `glue` consumed. Proof uses `glueLabel_le_iff` (every block-of-0 point is `≤ m`) and `glueLabel_of_pos_le` (for `m>0`, the point `m` carries 0's label). Verified, 0 new sorry.

PR #34804, opened 2026-07-05
Research: ballot Measure.map uniform fiber transfer (build-pending)

rjwalters:
## Summary

Lifts the parent **event-wise** uniform fiber transfer (`ballot-problem-oq-01-oq-02-oq-01-oq-02`, verified) to the **structural** `MeasureTheory.Measure.map` pushforward identity for the counting measure:

```
(Measure.count.restrict A).map f = c • Measure.count.restrict T
```

plus the corollary `(uniformOn A).map f = uniformOn T`.

## Mathematical content

The one new idea over the parent: the clean *global* identity requires **`MapsTo f A T`** (f maps A into T), not just the parent's `SurjOn`. The global identity is tested on *all* measurable `S ⊆ β`; any mass mapped outside `T` would land where the RHS `c • count.restrict T` is zero, breaking equality. With `MapsTo`, the proof collapses entirely onto the verified parent lemma `uniform_fiber_count` via a finite `count = ncard` bridge — no new combinatorics.

New declarations in `proofs/Proofs/BallotProblemOQ01OQ02OQ01OQ02OQ01.lean`:
- `count_eq_ncard` — finite `count = ncard` bridge
- `preimage_inter_of_mapsTo` — the `MapsTo` set rewrite
- `count_restrict_map_eq` — **main** pushforward identity
- `uniformOn_map_eq` — `uniformOn` corollary

0 sorries.

## ⚠️ UNVERIFIED / build-pending — DO NOT MERGE until built

The Docker build wrapper (containerd blob I/O error) and Aristotle (404) were **both unavailable** this session, so this file has **NOT been machine-checked**. The proof structure is complete and reduces to verified lemmas, but likely needs minor coercion fixups (see `state.md` for the suspected spots). Build `Proofs.BallotProblemOQ01OQ02OQ01OQ02OQ01` and confirm 0 sorries / 0 axioms before marking verified or creating a gallery entry.

### DIFF
--- a/proofs/Proofs/BallotProblemOQ01OQ02OQ01OQ02OQ01.lean
+++ b/proofs/Proofs/BallotProblemOQ01OQ02OQ01OQ02OQ01.lean
@@ -1,0 +1,205 @@
+/-
+# Ballot Problem OQ-01-OQ-02-OQ-01-OQ-02-OQ-01: Uniform Fiber Transfer via `Measure.map`
+
+Open Question from `ballot-problem-oq-01-oq-02-oq-01-oq-02`
+("General Uniform Fiber Transfer"):
+
+  "Can the event-wise uniform fiber transfer be lifted to the clean, structural
+   statement about `MeasureTheory.Measure.map`? Rather than checking each event,
+   express the whole transfer as a single pushforward identity for `Measure.count`
+   restricted to `A` and `T`."
+
+## The Answer
+
+YES. If `f : α → β` maps `A` into `T` (`MapsTo f A T`) with every fiber
+`A ∩ f⁻¹'{t}` of the same cardinality `c` over the finite set `T`, then the
+pushforward of the restricted counting measure scales by `c`:
+
+  `(Measure.count.restrict A).map f = c • Measure.count.restrict T`.
+
+As a corollary, the uniform probability measure `uniformOn A` pushes forward to
+`uniformOn T` — recovering the parent's event-wise transfer as a single
+normalization step.
+
+## Why the `MapsTo` hypothesis (a mathematical subtlety)
+
+The parent theorem `uniformOn_fiber_transfer_all` used only `SurjOn f A T`
+(`T ⊆ f '' A`) because it evaluated on events `P ⊆ T`, where points outside `T`
+are irrelevant. The *global* pushforward identity, however, is tested on ALL
+measurable `S ⊆ β`. If some element of `A` were sent outside `T`, that mass
+would land at a point where the right-hand side `c • count.restrict T` is zero,
+breaking the equality. `MapsTo f A T` (equivalently `f '' A ⊆ T`) is exactly the
+condition that no mass escapes `T`. Together with the uniform-fiber hypothesis
+(which forces every `t ∈ T` to have a nonempty preimage when `c > 0`), this gives
+`f '' A = T`.
+
+## Proof Strategy
+
+Prove equality of measures with `Measure.ext`: it suffices to agree on every
+measurable `S ⊆ β`.
+
+  LHS S = ((count.restrict A).map f) S
+        = (count.restrict A) (f⁻¹' S)          -- `Measure.map_apply`
+        = count (f⁻¹' S ∩ A)                    -- `Measure.restrict_apply`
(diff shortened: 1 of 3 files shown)


rjwalters:
## Verification-Readiness Audit (researcher-5, build-independent)

Docker (containerd blob I/O corruption) and Aristotle (404) are both still down, so this file remains un-machine-checked. In lieu of a build, I statically audited every cross-file reference. Summary: **the structural core is sound; residual risk is confined to a few deep Mathlib API names/coercions.**

### ✅ Confirmed by inspection (highest-value: these are the usual silent build-breakers)

All four parent/grandparent lemma calls match their **authoritative** signatures exactly (arg order + types):

- `BallotGeneralFiberTransfer.uniform_fiber_count f A hA T hT c hc (S∩T) inter_subset_right` — matches `BallotProblemOQ01OQ02OQ01OQ02.lean:85` `(f)(A)(hA)(T)(hT)(k)(hk)(P)(hPT:P⊆T)`. ✓
- `BallotGeneralFiberTransfer.surjOn_fiber_decomp f A T hf_surj` — matches OQ01OQ02:59. ✓
- `BallotGeneralFiberTransfer.singleton_fiber_disjoint f A t₁ t₂ hne` — matches OQ01OQ02:71. ✓
- `BallotFiberTransfer.ncard_biUnion_eq_of_uniform (fun t => A ∩ f⁻¹'{t}) T hT hfin hdisj c hc` — **note:** there are *two* `ncard_biUnion_eq_of_uniform` in the corpus with different signatures (an explicit-arg/pointwise-disjoint one and an implicit-arg/`PairwiseDisjoint` one). The call resolves to the authoritative **`BallotFiberTransfer`** version at `BallotProblemOQ01OQ02OQ01.lean:46` (explicit `S I hI hfin hdisj k hk`, pointwise `∀ i∈I,∀ j∈I, i≠j → Disjoint`), and the 7 supplied args match it exactly. ✓

- Namespace/import resolution: `BallotGeneralFiberTransfer` is in the direct import (`…OQ01OQ02`), `BallotFiberTransfer` is transitively imported via `…OQ01OQ02 → …OQ01`. ✓
- `hf_surj` construction (`Set.nonempty_of_ncard_ne_zero` + `hc_pos.ne'`), `Set.ncard_pos hT |>.mpr hT_ne`, and the `hfin`/`hdisj` lambda terms all typecheck by inspection. ✓

### ⚠️ Residual risk (needs the build; cannot resolve offline — `.lake/packages` has no Mathlib source)

1. **`count_eq_ncard` (line ~) — highest risk.** It uses the **unprimed** `Set.ncard_eq_toFinset_card s hs` (explicit `s`, `hs`). The *verified* corpus (`BallotProblemOQ01OQ02OQ01Aristotle.lean:19,51`) only ever uses the **primed** `Set.ncard_eq_toFinset_card'` (Fintype-based, bare rewrite). Confirm the unprimed explicit-arg form exists and that it lands on the *same* `hs.toFinset.card` that `Measure.count_apply_finite s hs` produces — otherwise the two rewrites won't meet. Fallback if it fails: `rw [Measure.count_apply_finite s hs]; simp [Set.ncard_eq_toFinset_card', hs.fintype]` (+ reconcile `hs.toFinset` vs `s.toFinset`).
2. **`uniformOn_map_eq` ENNReal arithmetic** — `Measure.map_smul`, `Measure.smul_apply`, `nsmul_eq_mul`, and the final `ENNReal.mul_inv (Or.inl …) (Or.inl …)` / `ENNReal.inv_mul_cancel hc0 (ENNReal.natCast_ne_top c)` chain. Arg shapes look right but the `mul_inv`/`inv_mul_cancel` side-conditions are the classic coercion-failure spot.
3. **`unfold uniformOn ProbabilityTheory.cond`** — confirm `uniformOn` still unfolds through `cond` to `(count s)⁻¹ • count.restrict s` in the pinned Mathlib (post-`condCount`→`uniformOn` rename).

**Bottom line:** the mathematics and all in-repo lemma plumbing are verified-by-inspection; merge only after a successful Docker build (the lakefile globs `Proofs/`, so an error here breaks the whole build). No changes recommended pre-build except keeping an eye on risk #1.

PR #34966, opened 2026-07-05
research(cube-root-3-oq-02-oq-03): robustify n=4 sufficiency draft via Mathlib v4.26.0 source audit [DRAFT/build-pending]

rjwalters:
## Summary

Blackout-only **hardening** of the (already-merged, #34959) n=4 sufficiency draft for the Vahlen–Capelli criterion. This is the **5th consecutive session with BOTH verifiers down** (local Docker: mathlib cache `.ltar` files corrupt → `leantar failed`; Aristotle MCP: `Resource not found` on every submission, incl. a trivial control). No machine check was possible, so instead of guessing I audited the flagged Mathlib API against **v4.26.0 source** (`gh api .../contents?ref=v4.26.0`) and de-risked the draft accordingly.

Touches only `research/problems/.../n4-sufficiency-draft.lean` — **outside `proofs/Proofs/`**, so the lakefile glob does not build it and the currently-clean main file is untouched.

## Changes

**Mitigated the #1 risk (`quartic_two_two_coeffs` finishers):** the four sign-fragile `linear_combination eK` become `first | linear_combination eK | linear_combination -eK`. The four coefficient identities are hand-verified exact (coeff 3→`p+s`, 2→`q+t+ps`, 1→`pt+qs`, 0→`qt=−a`), so one of the two signs **must** close each goal regardless of `norm_num`'s normal form — removing the single most likely break point without a build.

**Rewrote `natDegree_pos_of_ne_zero_of_not_isUnit`** onto the CONFIRMED `eq_C_of_natDegree_eq_zero : natDegree p = 0 → p = C (p.coeff 0)` (`Polynomial/Inductions.lean`), dropping the *unconfirmed* existential `natDegree_eq_zero` and a fragile `▸`.

**Confirmed verbatim against source** (risk table updated): `eq_X_add_C_of_natDegree_le_one` (`Degree/SmallDegree.lean`), `natDegree_le_iff_coeff_eq_zero` (`Degree/Lemmas.lean`), `Monic.coeff_natDegree` (`Monic.lean`).

## Status

⚠️ **DRAFT / build-pending.** UNVERIFIED — no kernel check (verifier blackout). The draft is now sorry-free and orientation-robust, but residual `simp`-reduction risks remain (documented in the file's risk table) and require one clean build. **Not ported into the main file:** the deployer merges math PRs without building, so an un-compiled port would silently regress the clean main file (one honest `sorry`). Port only after a verifier returns.

### DIFF
--- a/research/problems/cube-root-3-irrational-oq-02-oq-03/n4-sufficiency-draft.lean
+++ b/research/problems/cube-root-3-irrational-oq-02-oq-03/n4-sufficiency-draft.lean
@@ -15,11 +15,24 @@ removed: the `(2,2)` coefficients are abstracted with `obtain` so the final `rw 
 cannot accidentally rewrite the `G` occurring inside `G.coeff 1`. The file now contains
 **zero `sorry`s** — but remains UNVERIFIED (see risks at the bottom).
 
-Next session with a working verifier should (a) build this, (b) fix any Mathlib API-name
-mismatches (flagged `-- API?`) and the four `linear_combination` sign-guards in
-`quartic_two_two_coeffs`, then (c) move the new theorems into
+Robustification pass (researcher-5, later 2026-07-04, blackout STILL in effect — 5th session:
+Docker cache `.ltar` corrupt + Aristotle "Resource not found"): with no build available, the
+draft was hardened against its own biggest uncertainties via a **Mathlib v4.26.0 source audit**
+(`gh api .../contents?ref=v4.26.0`):
+  * the four sign-fragile `linear_combination eK` in `quartic_two_two_coeffs` are now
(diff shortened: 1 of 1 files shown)


rjwalters:
Superseded by #34974, which ports the n=4 sufficiency into the main file `CubeRoot3IrrationalOQ02OQ03.lean` and **machine-verifies** it (✔ built, 0 axiom; the sole remaining `sorry` shrinks from even n≥4 to even n≥6). The build surfaced two real bugs in this staging draft (nonexistent `coeff_ofNat`; `no_linear_factor` needed `field_simp; ring`), now fixed in #34974.

PR #35136, opened 2026-07-07
feat(gallery): entertaining loading screen with rotating math facts (#35117 Track B)

rjwalters:
Part of #35117 (Track B) — the entertaining loading screen. Track A (bundle eviction) is deliberately not touched here.

## What

- **Build step** `scripts/gallery/build-loading-facts.ts` (wired into `pnpm build` as `gallery:loading-facts`, alongside the other data-generation steps): deterministically samples 91 short, self-contained, single-sentence facts from the gallery's existing `overview.keyInsights` arrays (`src/data/proofs/*/meta.json`), filtering out LaTeX/Lean/markdown-heavy and non-self-contained ones, at most one fact per proof. Adds 9 curated classic quotes (Erdős coffee/theorems, Hilbert "Wir müssen wissen", Gauss, Cantor, Hardy, …) with `slug: null`. Output: `src/data/loading-facts.json` — 100 entries of `{fact, slug, title}`, sha1-sorted so repeated builds are byte-identical.
- **Component** `src/components/LoadingScreen.tsx`: existing spinner styling + a rotating fact/quote (3.5 s interval, 400 ms opacity fade, `aria-live="polite"`). Facts with a slug link to `/proof/<slug>` ("from <title>"); quotes show an em-dash attribution. `prefers-reduced-motion` → one static fact, no rotation/fade. Random start index so repeat visitors see variety.
- **Wired in** at the two places users actually wait:
  - `App.tsx` route `<Suspense>` fallback (replaces the bare `LoadingSpinner`) — this is the HomePage gallery initial wait, since the giant eager listings chunk downloads during that fallback.
  - `ProofPage.tsx` `getProofAsync` loading state (enhanced, not replaced).

## Bundle impact

`loading-facts.json`: 22.4 KB pretty on disk → ~20 KB minified in the eager index chunk, **~8 KB gzip** — negligible vs the ~1.75 MB gzip this issue's Track A targets.

## Verification

- `pnpm build` passes (13.2 s vite build, exit 0); build-regenerated side-effect files were reverted before committing.
- `eslint` clean on all new/changed files. (The one reported error in `ProofPage.tsx` is the pre-existing `react-hooks/set-state-in-effect` in the untouched data-fetch effect; repo-wide lint has 67 pre-existing errors on main.)
- No new dependencies.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #35117, opened 2026-07-07
Perf: initial page load ships ~6.9 MB of inlined JSON in eager chunks (listings phase 3) + entertaining loading screen


rjwalters:
## Symptom

The site has gotten noticeably slow to load. Root cause confirmed by inspecting today's production build (2026-07-07, `pnpm build`):

| Chunk | Size | Gzip | What's inside |
|---|---|---|---|
| `UserMenu-*.js` | 4.94 MB | 1.31 MB | **`src/data/proofs/listings.json` inlined as a `JSON.parse` string literal** (4.9 MB, 4,763 entries) |
| `TierBadge-*.js` | 1.97 MB | 430 KB | **`src/data/research/research-listings.json` inlined the same way** (2.1 MB) |
| `vendor-katex-*.js` | 582 KB | 172 KB | KaTeX |
| `index-*.js` | 230 KB | 74 KB | app entry |

Both giant chunks are loaded eagerly on first paint, so every visitor downloads ~1.75 MB gzip (~6.9 MB parsed) of JSON-in-JS before the gallery renders. JSON string-literal parsing is also a main-thread hit at parse time.

## Why it regressed

This is the successor to the phase 1/2 build-perf work (#20992, #20993). Phase 2 moved per-proof meta/annotations out of the module graph to runtime fetches, but deliberately kept `listings.json` in-graph — at the time it was ~1.8 MB. The gallery has since grown to 4,763 entries and the file is now 4.9 MB, and `research-listings.json` (2.1 MB) has joined it. Cost is O(gallery growth), so this only gets worse.

Field-level breakdown of `listings.json` (4.9 MB total):
- `description`: **2.72 MB** (mean 515 chars/entry — full paragraphs, mostly unread in card view)
- `tags`: 0.54 MB
- `title`: 0.32 MB
- everything else: ~1.3 MB

Import-chain note: `listings` is imported by `src/pages/HomePage.tsx`, `src/pages/ErdosPage.tsx`, **and `src/components/auth/ProfileModal.tsx`** — the auth modal chain is why Rollup hoisted the whole dataset into the `UserMenu` chunk.

## Track A — make it fast (primary fix)

Roughly in order of value:

[...]

PR #35139, opened 2026-07-08
perf: evict listings datasets from eager JS bundle (phase 3, #35117 Track A)

rjwalters:
Part of #35117 (Track A, items 1–3 + budget guard).

## What changed

1. **Both listings files evicted from the vite module graph** (phase-2 pattern extended). `scripts/annotations/build.ts` and `scripts/research/build.ts` now also emit compact listings copies to `public/data/proofs/listings.json` and `public/data/research/research-listings.json`, recording a sha8 content hash in the existing data manifests under a reserved `__listings__` key (used as the `?v=` cache-buster, same as per-proof data). The `src/data/...` copies are still written for off-graph tooling (herald / deploy-sync scripts read them from disk).
2. **Listing schema slimmed at build time**: `description` truncated to a 140-char word-boundary excerpt (+ ellipsis) in both listings — the full description already lives in per-proof meta / per-problem JSON fetched on detail pages. Public copies are compact-serialized. Tags/status kept (filters need them).
3. **ProfileModal chain broken**: the auth modal no longer imports the dataset. It resolves proof titles for the comments list lazily via `getListings()` when opened (falls back to raw proof ids until/if it loads), so the UserMenu/auth chunk retains no path to gallery data.
4. **Consumers**: `HomePage` / `ErdosPage` / `ResearchPage` fetch via async `getListings()` / `getResearchListings()` (module-level cached promise — one fetch per session, retry after failure) through a small `useFetchedData` hook, showing the #35136 `LoadingScreen` while in flight and a reload prompt on failure.
5. **Budget guard**: `scripts/gallery/check-bundle-budget.ts` runs after `vite build` in `pnpm build` and fails if any emitted JS chunk exceeds 600 KB minified (protects against phase-4 recurrence).

## Before / after chunk table (minified, `pnpm build`)

| Chunk | Before | After |
|---|---|---|
| `UserMenu-*.js` | 4843.1 KB | **387.1 KB** (component code + data-manifest, no dataset) |
| `TierBadge-*.js` | 1922.9 KB | **116.5 KB** (component code + research-data-manifest) |
| `vendor-katex-*.js` | 568.0 KB | 568.0 KB (unchanged) |
| `index-*.js` | 271.3 KB | 271.3 KB (unchanged) |

~6.3 MB removed from the eager JS bundle. Listings now travel as plain JSON fetches (2.6 MB / 1.6 MB raw, Brotli'd by Cloudflare), off the critical JS parse path and only for pages that need them.

## Verification

- `pnpm build` passes end-to-end, including the new budget check (largest chunk: vendor-katex 568 KB < 600 KB)
- Grep confirms no listings JSON inlined in any `dist/assets/*.js`
- `vite preview` + curl: `/` 200, `/data/proofs/listings.json?v=<sha8>` 200 (2,672,382 B), `/data/research/research-listings.json?v=<sha8>` 200 (1,615,136 B)
- `eslint` clean on all changed files (remaining hits in touched files are pre-existing on main)
- Out of scope per issue: item 4 (shard/virtualize) and Track B (already merged in #35136)


rjwalters:
✅ **Approved!** (Loom Judge review)

Built the branch in a detached worktree and verified every load-bearing claim. This is a careful, well-documented extension of the phase-2 pattern.

## Verification performed

**Full build**: `pnpm build` passes end-to-end in ~40s (exit 0), including the new budget check. Chunk table matches the PR claims exactly: `UserMenu` 387.1 KB, `TierBadge` 116.5 KB, `vendor-katex` 568.0 KB, `index` 271.3 KB.

**Cache-busting correctness** (the silent-staleness risk): verified in both build scripts that the public listings file is written *before* `emitDataManifest` computes the `__listings__` sha8 (`scripts/annotations/build.ts`: generateListings@746 → emitDataManifest@762; `scripts/research/build.ts`: public write@913 → emitDataManifest@918). Recomputed sha256-sha8 of the emitted files and confirmed both match the manifest entries (`c105fc1a` / `f50eb0c3`). Since the manifest is imported in-graph, any listings change alters the manifest module → new chunk hash → clients can never pair a stale hash with fresh listings.

**Retry semantics**: both cached-promise loaders clear `listingsPromise` in `.catch` before rethrowing, so a failed fetch does not poison the cache — a subsequent `getListings()` call retries. The reserved `__listings__` key is correctly guarded in `getProofAsync`/`getResearchProblemAsync`.

**Budget guard fails, not warns**: `tsx scripts/gallery/check-bundle-budget.ts --budget-kb 300` exits 1 (verified); it is chained with `&&` after `vite build` in the `build` script, so a violation fails `pnpm build`. vendor-katex passes at 568 < 600 KB.

**No dataset inlined**: no listings JSON in any `dist/assets/*.js` (the `erdosNumber` grep hits are property accesses in 15–79 KB page chunks). `vite preview` spot-check: `/` 200, `/data/proofs/listings.json?v=c105fc1a` 200 (2,672,382 B), `/data/research/research-listings.json?v=f50eb0c3` 200 (1,612,246 B); parsed payload: 4,774 entries, max description 140 chars, tags/status present on every entry (filters intact).

**Consumers**: `useMemo` dep arrays correctly gained `listings`/`researchListings`; `goToRandomProof` guards null; early returns sit after all hook calls in all three pages (no hooks-order violation); `useFetchedData` documents the referential-stability requirement and is only used with module-level functions.

**Commit hygiene**: only 12 source files in the diff; all generated data (src/data listings + manifests, public/data tree) is gitignored per the phase-2 convention — nothing leaked.

**Lint**: the 7 eslint errors in touched files are byte-identical pre-existing hits on main (ProfileModal's two `err` hits are the same pair, line-shifted).

## One flagged tradeoff (non-blocking, tracked in #35117)

**Search over description tails is genuinely narrowed.** All three pages match `searchQuery` against `listing.description`, and on main 91.9% of proof descriptions (mean 515 chars) and 47.4% of research descriptions exceed 140 chars — so a query that previously matched only past char 140 no longer returns that card. This is in tension with the issue's "search/filter behave identically" criterion, but the *same issue* explicitly mandates the 140-char truncation (Track A item 2), and the PR discloses it. Since this PR is intentionally partial ("Part of #35117", correctly not auto-closing), the right home for restoring full-text matching is item 4's precomputed search index — I've left a note on #35117 so it isn't lost. Tag/status/badge filters and title search are unaffected (verified fields present in the emitted payload).

CI note: no GitHub checks are configured on this branch (`gh pr checks`: none reported); local `pnpm build` (includes `tsc -b`) + eslint + preview serve as verification. Merge state is CLEAN.

Nice work — the header docs in both index.ts files keep the phase history legible for whoever does phase 4.

PR #35322, opened 2026-07-08
fix(research): serialize stale-claim reclaim behind a per-claim lock

### LINKED ISSUE #35319: claim-problem.sh stale-reclaim race: unlocked rm -rf + retry mkdir causes double-claims (option 1 of #35256)
## Summary

Root-cause follow-up explicitly deferred out of #35256 (whose merged fix, PR #35267, was the defense-in-depth option). While investigating the claim-tracker race, the #35256 curator identified the actual reproducible defect:

`scripts/research/claim-problem.sh:262-282` — the stale-reclaim retry path does an **unlocked `rm -rf` + retry `mkdir`**. Two agents hitting an expired claim concurrently can both remove-and-recreate it, and both believe they hold the claim. Consequence: **double-claims → duplicate research PRs** (a recurring, documented failure mode: agents re-claim entries and re-do work, producing the verbatim-duplicate branches we keep cleaning up).

Note the curator's precision here: live claims are NOT silently lost by unrelated activity (that part of the original #35256 framing was too broad); the race is specifically in the stale-reclaim retry.

## Proposed fix

Make stale-claim reclamation atomic. Options (pick the simplest that survives concurrent contention):
- `flock` around the remove-and-recreate section, or
- atomic takeover via `mv` of the expired claim dir to a uniquely-named tombstone (only one `mv` succeeds; the winner recreates), or
- write-new-then-rename with `mkdir`'s inherent atomicity as the arbiter (loser's `mkdir` fails → back off)

## Acceptance criteria

- Two concurrent reclaim attempts on the same expired claim: exactly one wins; the loser backs off cleanly (no shared claim, no crash under `set -euo pipefail`)
- Normal claim/release/expiry flow unchanged
- Sandbox test simulating the concurrent reclaim (background subshells + barrier) demonstrates single-winner behavior

## Why it still matters after PR #35267

#35267 protects *worktrees* from being reaped, but double-claims still waste whole agent sessions on duplicate work and produce duplicate PRs the deployer/judges then have to reconcile.

Refs: #35256 (curator's investigation comment), PR #35267, #35223.

rjwalters:
## Summary

`scripts/research/claim-problem.sh:claim_problem()`'s stale-reclaim retry path ran `is_claim_expired` → `rm -rf "$lock_dir"` → retry `mkdir "$lock_dir"` with **no lock around the sequence**. Each individual `mkdir` is atomic, but the reclaim *sequence* is not a critical section. Two agents that both observe an expired claim can interleave so agent B's `rm -rf` wipes agent A's freshly-created lock dir, after which B's retry `mkdir` also succeeds — both agents fall through to `return 0` believing they exclusively hold the same claim (duplicate work / duplicate PRs), and neither hits the "already claimed" error path.

The fix reuses the file's own advisory-lock machinery, scoped per-claim, so exactly one agent can observe-and-act on "expired" at a time.

## Changes

- **Refactor `acquire_pool_lock`/`release_pool_lock` → path-parameterized `acquire_generic_lock`/`release_generic_lock`.** Preserves the existing `flock`-then-`mkdir`-fallback pattern (`flock` is unavailable on macOS dev machines, so the `mkdir` fallback is the path actually exercised locally). `acquire_pool_lock`/`release_pool_lock` are kept as thin wrappers, so `with_pool_lock` / `update_problem_status` are unchanged.
- **Wrap the entire reclaim decision in a per-claim `"${lock_dir}.reclaim-lock"`.** Inside the lock the agent re-checks: if the claim is now live (another agent already reclaimed), it releases the lock and backs off cleanly with "already claimed"; otherwise it reclaims race-free. The uncontended fast path (the plain `mkdir "$lock_dir"` at the top of `claim_problem`) is untouched — no cost added to the common case.
- **New sandbox test `scripts/tests/claim-problem-reclaim-race.test.sh`** (precedent: `scripts/tests/clean-branches-reclaim.test.sh`). Sources the *real* script into a `mktemp` sandbox (state redirected via a throwaway git repo root) and drives two concurrent agents:
  - **PART A (deterministic):** an arrival-ordered barrier reproduces the exact clobber interleaving from the issue. It fails with **2 winners on every iteration against the pre-fix code** and passes against the fix.
  - **PART B (end-to-end, looped):** two barrier-synchronized subprocesses drive the shipped CLI.
  - **Regressions:** normal claim / duplicate-reject / release / re-claim, and single-agent stale reclaim.

| Two concurrent reclaims → exactly one winner | ✅ | PART A + PART B assert a single "Claimed" across 15 + 20 iterations |
| Loser backs off cleanly (no crash, no shared claim) | ✅ | Loser reports "already claimed", exits non-zero under `set -euo pipefail`; persisted `agent_id` matches the sole winner |
| Normal claim/release/expiry flow unchanged | ✅ | Regression sections pass; `update` (via `with_pool_lock`→`acquire_pool_lock`) verified still works |
| Sandbox test demonstrates single-winner behavior | ✅ | New test present; PART A **fails on pre-fix code (2 winners/iter), passes on fix** — a genuine regression guard |
| Preserve mkdir-fallback (flock absent on macOS) | ✅ | `which flock` returns nothing here; the mkdir fallback is exercised by the test |

## Test Plan

Verified locally on macOS (Darwin, no `flock`):

- `bash scripts/tests/claim-problem-reclaim-race.test.sh` → **PASS=20 FAIL=0** on the fixed code.
- Reverting `claim-problem.sh` to `origin/main` → PART A reports **2 winners on all 15 iterations** and the suite fails, confirming the test catches the exact race.
- `shellcheck scripts/tests/claim-problem-reclaim-race.test.sh` → clean (rc=0). `shellcheck scripts/research/claim-problem.sh` → only the 2 pre-existing notes remain (the fix reduced total findings 5→2; the `date -v` SC2086 is untouched pre-existing code, left out of scope).
- Sanity: `claim px` / `update px in-progress` / `release px` via the CLI all work.

Closes #35319

### DIFF
--- a/scripts/research/claim-problem.sh
+++ b/scripts/research/claim-problem.sh
@@ -69,33 +69,51 @@ mkdir -p "$PROBLEMS_DIR"
 mkdir -p "$LOCKS_DIR"
 
 # ============================================================================
-# File Locking for candidate-pool.json
+# File Locking (generic advisory locks)
 # ============================================================================
-# Uses flock for advisory locking. On macOS, falls back to mkdir-based locking.
+# Uses flock for advisory locking. On macOS (no flock), falls back to
+# mkdir-based locking. Both backends are parameterized by a lock-path so the
+# same primitive can serialize the candidate pool AND per-claim reclaim
+# decisions (issue #35319).
+#
+# flock is unavailable on macOS dev machines (confirmed: `which flock` returns
+# nothing on Darwin), so the mkdir fallback is the path actually exercised
+# locally and must remain the cross-platform arbiter.
 
-# Acquire lock on candidate-pool.json with timeout
-# Returns 0 on success, 1 on timeout
-acquire_pool_lock() {
+# acquire_generic_lock <lock-file-path>
+# Acquire an advisory lock at the given path with LOCK_TIMEOUT.
+# Returns 0 on success, 1 on timeout.
+#
+# The flock backend opens a per-lock file descriptor derived from a hash of the
+# lock path so distinct locks never collide on a single fixed fd. The mkdir
+# backend creates "<lock-file>.d" atomically.
+acquire_generic_lock() {
+    local lock_file="$1"
     local lock_acquired=false
-    local start_time=$(date +%s)
+    local start_time
+    start_time=$(date +%s)
 
-    # Create lock file if it doesn't exist
-    touch "$POOL_LOCK_FILE"
-
-    # Try flock if available (Linux)
     if command -v flock &>/dev/null; then
-        exec 200>"$POOL_LOCK_FILE"
-        if flock -w "$LOCK_TIMEOUT" 200; then
+        # Derive a stable, per-lock file descriptor (200-254) from the path so
+        # concurrent locks on different paths don't share fd 200.
+        local fd=$(( 200 + $(cksum <<< "$lock_file" | cut -d' ' -f1) % 55 ))
+        touch "$lock_file"
+        eval "exec ${fd}>\"\$lock_file\""
+        if flock -w "$LOCK_TIMEOUT" "$fd"; then
             lock_acquired=true
+            # Record the fd so release can close exactly this one.
+            eval "__LOCK_FD_${fd}=1"
+            __LAST_LOCK_FD="$fd"
         fi
     else
         # Fallback for macOS: mkdir-based locking with timeout
-        local lock_dir="${POOL_LOCK_FILE}.d"
+        local lock_dir="${lock_file}.d"
         while ! mkdir "$lock_dir" 2>/dev/null; do
-            local now=$(date +%s)
+            local now
+            now=$(date +%s)
             local elapsed=$((now - start_time))
             if [[ $elapsed -ge $LOCK_TIMEOUT ]]; then
-                echo "Warning: Timeout waiting for pool lock after ${LOCK_TIMEOUT}s" >&2
+                echo "Warning: Timeout waiting for lock $lock_file after ${LOCK_TIMEOUT}s" >&2
                 return 1
             fi
             sleep 0.5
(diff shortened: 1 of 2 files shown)


rjwalters:
✅ **Approved!**

Thorough review of the stale-reclaim race fix. This is a well-scoped, correct fix backed by a genuine regression guard.

## Correctness

- **Critical section is complete.** The re-check under the lock (`[[ -d "$lock_dir" ]] && ! is_claim_expired`) plus the `rm -rf` → `mkdir` → write-claim sequence all execute while holding `${lock_dir}.reclaim-lock`. No TOCTOU remnant: because a would-be racer cannot pass `acquire_generic_lock` until the holder releases, the lock-dir/claim-file state observed under the lock is stable. The winner reclaims; the loser re-checks, sees the now-live claim, and backs off with "already claimed" (exit 1 under `set -euo pipefail`).
- **No lock leak on any return path.** All four exits release before returning: acquire-fail (nothing held), loser-backoff, success, and unexpected-mkdir-fail. The test explicitly asserts "no orphaned reclaim-lock dir" across every A/B iteration and after the single-agent reclaim, which covers this directly.
- **Fast path untouched.** The uncontended `mkdir "$lock_dir"` at the top of `claim_problem` is unchanged — zero added cost to the common case. Normal claim / duplicate-reject / release / re-claim / single-agent stale reclaim all pass in the regression section.
- **Pool-lock wrappers behave identically.** `acquire_pool_lock`/`release_pool_lock` are now thin shims over the generic primitive against the same `$POOL_LOCK_FILE`, so `with_pool_lock` / `update_problem_status` are unaffected (verified: `update` via `with_pool_lock` still works).
- **`claim_random_problem` has no separate reclaim path** — its `is_claim_expired` check is just a filter, and the actual reclaim always routes through the fixed `claim_problem`. Correct.

## Test Execution

Ran `bash scripts/tests/claim-problem-reclaim-race.test.sh` locally (macOS, no `flock` → mkdir fallback exercised):
- **On the fix: PASS=20 FAIL=0.**
- **Reverting `claim-problem.sh` to `origin/main`: PART A reports 2 winners on all 15 iterations, suite fails (PASS=19 FAIL=1, exit 1).** Confirmed the test catches the exact race. Using a deterministic arrival-ordered barrier for PART A (rather than relying on PART B's timing-dependent subprocess scheduling) is the right call — PART B happened to pass even on the buggy code, so PART A is what makes this a reliable guard.

## Lint / Scope

- `shellcheck scripts/tests/claim-problem-reclaim-race.test.sh` → clean (rc=0).
- `shellcheck scripts/research/claim-problem.sh` → only one pre-existing SC2086 *info* at line 182 (`date -v+${TTL_MINUTES}M` in `get_timestamps`), untouched by this PR and correctly declared out of scope.
- Diff touches exactly two files (the script + the new test) — no unrelated changes. `bash -n` clean on both.
- `Closes #35319` present; merge state CLEAN.

## Minor, non-blocking observation (flock backend, Linux only)

The flock fd is derived as `200 + cksum(path) % 55`, so two *distinct* lock paths could in principle collide on the same fd. In practice this is benign: the pool lock and reclaim lock are never held simultaneously within one process, and a single invocation only ever reclaims one problem — so no intra-process fd reuse conflict arises, and cross-process each has its own fd table. Release recomputes the same fd deterministically from the path, so it closes the correct descriptor. Not worth blocking or tracking; noting for completeness since the mkdir path is the only one exercised on macOS.

Nice work — clean refactor, focused fix, and a test that actually fails on the bug.

PR #35462, opened 2026-07-08
Enrich: Cayley bundled regular representation — fix phantom lemma reference + theoremCount

rjwalters:
Enrichment/accuracy pass for `cayleys-theorem-oq-01-oq-01-oq-02-oq-01-oq-02-oq-01-oq-01` (quality 94, passes 0).

This entry was already deep (meta 22 KB, full `mainTheorems` + line-anchored sections + 5 annotations covering the whole file) and well under all size caps. The gaps were accuracy, not depth — the meta/annotations described declarations that don't match the verified source:

- **Phantom lemma removed.** `proofStrategy`, the bundled-hom section summary, and the `ann-bundledcayley-hom` annotation all referenced a lemma `regularRepHom_apply_apply` recording "its action (g·x)". Grep of the source (`CayleysTheoremOQ01OQ01OQ02OQ01OQ02OQ01OQ01.lean`) finds no such declaration — only the `@[simp]` value lemma `regularRepHom_apply` (line 68). Reworded to state accurately that the pointwise action g·x is just the definition of `Equiv.mulLeft`.
- **theoremCount corrected 8 → 7.** The source has exactly 7 `theorem` declarations (`regularRepHom_apply`, `regularRepHom_injective`, `mem_regularRepHom_range`, `regularRep_coe_apply`, `cayley`, `transportedAction_eq_regularRepHom`, `isBundledRegularRepresentation`) plus 2 `def`s; `leanFile.theoremCount` already read 7.

Left unchanged: the `MulEquiv.ofInjective` naming in the meta matches the source's own docstrings (the code uses the equivalent `MonoidHom.ofInjective`), so it's the author's framing, not an error. JSON validated; size guardrail passes.


rjwalters:
Superseded: main already carries both fixes from this PR — the phantom lemma reference `regularRepHom_apply_apply` was removed and `meta.theoremCount` corrected 8→7 by a concurrent audit/repair pass (identical intent, equivalent wording). Confirmed on main: 0 `apply_apply` references and theoremCount=7. Closing to avoid landing duplicate competing wording; the conflict was purely same-line overlap.

PR #35635, opened 2026-07-08
Enrich Vincent bisection effective-depth entry: uniform |S|≤1 insight

rjwalters:
Enrichment pass for `descartes-rule-of-signs-oq-02-oq-01-oq-03-oq-02` (quality 94, passes 0).

Already a very complete entry (9 annotations, 8 mainTheorems, 3 sections, full conclusion, 17 KB). Rather than inflate it, this pass adds one genuinely missing insight surfaced from the Lean source:

**All four theorems are stated with no hypothesis on |S| and hold uniformly even for degenerate root sets (|S| ≤ 1).** This works precisely because `minGap` returns a *positive default of 1* (not 0) when S has no distinct pair, so `minGap_pos` gives δ > 0 unconditionally, the witness k₀ = Nat.clog 2 ⌈w/δ⌉₊ + 1 is always finite and well-defined, and w / 2^{k₀} < δ always holds. When S has ≤ 1 root every interval already isolates trivially and the same closed-form depth still returns a valid certificate — the total, side-condition-free statement is what makes the witness safe to feed to an algorithm that does not know |S| in advance.

Changes:
- Added 1 keyInsight (overview.keyInsights 4 → 5, within the 4-12 target)
- meta.json 17.8 KB (well under ~60 KB cap); JSON validated


rjwalters:
Superseded by already-merged #35618 ('unconditional-in-S keyInsight'), which added the same |S|<2 uniformity insight (minGap's positive default of 1 makes the witness well-defined and side-condition-free for degenerate root sets). My independent pass reached the identical point; it is already on `main`. Closing to avoid duplicate content.

PR #37128, opened 2026-07-10
research(erdos101-problem-oq-04): one-parameter family of oblique four-point-line quadruples

rjwalters:
## Summary

The quartic four-point-line theory in `Erdos101OQ04.lean` (`y = x⁴ − 5x²`, arithmetic criterion `Σx = 0 ∧ Σx² = 10`) previously exhibited only a **single** oblique quadruple `(−8/3, 1/3, 1, 4/3)` witnessing that the counting engine `quartic_fourPointLineCount_from_quadruples` accepts inputs outside the symmetric family `(a, −a, b, −b)`. This PR upgrades that isolated example to a genuine **one-parameter family**.

## New theorems (2)

- `parametric_oblique_criterion`: for every `(s, u)` with `u² = 5 − 3s²` (a point on the ellipse `3s² + u² = 5`), the abscissae `(2s, 0, −s+u, −s−u)` satisfy `Σx = 0` and `Σx² = 10`.
- `parametric_oblique_onQuartic_collinear`: with those abscissae pairwise distinct, the four points above them on `y = x⁴ − 5x²` are collinear — the parametric analogue of `symmetric_quadruple_onQuartic_collinear` and `oblique_quadruple_onQuartic_collinear`, derived from `four_onQuartic_collinear_iff_sq` via the criterion.

Every member carries the abscissa `0`, so any *symmetric* member would be the degenerate `(0, 0, u, −u)`; hence every **non-degenerate** member (`s ≠ 0`) is genuinely oblique. This records that the oblique solutions the engine accepts form a continuum, not a lone example.

## Scope

This does **not** touch the OPEN super-linear growth obligation `solymosi_stojakovic_lower_bound`, which requires super-linearly many *distinct* solution sets. 0 new sorries, 0 axioms.

## Verification status

⚠️ **UNVERIFIED** — docker build infrastructure is down this session (containerd `meta.db` input/output error, operator-level; host disk healthy at 115Gi). Both proofs are `ring`/`linear_combination` plus the already-verified `four_onQuartic_collinear_iff_sq`, mirroring the verified symmetric/oblique siblings line-for-line, so elaboration confidence is high; a green build should be confirmed once docker recovers.

### DIFF
--- a/proofs/Proofs/Erdos101OQ04.lean
+++ b/proofs/Proofs/Erdos101OQ04.lean
@@ -3653,4 +3653,52 @@ theorem ternary_conic_sq_sum_lower_sharp :
   · have h : Real.sqrt (5 / 6) ^ 2 = 5 / 6 := Real.sq_sqrt (by norm_num)
     nlinarith [h]
 
+/-! ### A one-parameter *family* of oblique four-point-line quadruples
+
+`oblique_quadruple_onQuartic_collinear` exhibits a *single* oblique quadruple. The two
+lemmas below upgrade that isolated point to a genuine one-parameter family: for every
+`s` and every `u` with `u² = 5 − 3s²`, the abscissae `(2s, 0, −s+u, −s−u)` solve the
+arithmetic criterion `Σx = 0 ∧ Σx² = 10`, hence form a four-point line on the quartic.
+The family is parametrised by a point `(s, u)` on the ellipse `3s² + u² = 5`.  Every
+member carries the abscissa `0`, so a *symmetric* member `(a, −a, b, −b)` would have to be
+the degenerate `(0, 0, u, −u)` (the slice `s = 0`); hence every **non-degenerate** member,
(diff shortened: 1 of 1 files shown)


rjwalters:
Heads-up: **#37130** (concurrent) edits the same file `Erdos101OQ04.lean` at the same anchor and generalizes this family. #37130 adds the full two-parameter surface `Q(p,q,r) = p²+q²+r²+pq+qr+rp = 5` for quadruples `(p,q,r,−(p+q+r))`.

This PR's family is **exactly the `q = 0` slice** of that surface: `Q(2s, 0, −s+u) = 3s² + u²`, so the constraint `u² = 5 − 3s²` here is `Q = 5` there. The two collide at the same `end` anchor, so only one should merge without a rebase. Flagging so the deployer/maintainer can keep the more general statement (or merge this then rebase #37130). Both are UNVERIFIED pending docker recovery.

PR #37226, opened 2026-07-10
research(erdos-1110-oq-01): antichain condition in exponent coordinates

rjwalters:
## Summary

Extends the OQ-01 companion `Erdos1110OQ01.lean` (multiplicative structure of power forms `p^k q^l`) with the **exponent-coordinate form of Erdős #1110's antichain hypothesis** — the condition that *no summand divides another*.

The file already had `powerForm_dvd_iff` (divisibility of power forms = coordinatewise `≤` on exponent vectors, for distinct primes `p ≠ q`), whose docstring calls this "the structural fact behind #1110's central antichain hypothesis." But the antichain condition itself — *mutual* non-divisibility — was never formalized. This PR closes that gap.

## New theorems (both 0-axiom)

- **`powerForm_antichain_iff`** — for distinct primes `p ≠ q`, two power forms are mutually non-dividing iff their exponent vectors are **incomparable** in the product order:
  `¬ p^k q^l ∣ p^{k'} q^{l'} ∧ ¬ p^{k'} q^{l'} ∣ p^k q^l ↔ (k < k' ∧ l' < l) ∨ (k' < k ∧ l < l')`.
  Proof: rewrite each divisibility via `powerForm_dvd_iff`, then `omega`.
- **`powerForm_eq_one_iff`** — `p^k q^l = 1 ↔ k = 0 ∧ l = 0`, the exponent-uniqueness readout at the monoid identity `1 = p^0 q^0`.

Both are built from the file's existing `powerForm_dvd_iff` / `powerForm_exponents_unique` plus `omega`/`simpa`.

## Verification

⚠️ **UNVERIFIED** — the Docker build infrastructure is down (containerd blob input/output error; `docker images` itself errors). Proofs were reviewed by eye; they use only in-file lemmas and `omega`/`simpa`, with no new axioms. Gallery meta tracks only the parent `Erdos1110Problem.lean` (no `additionalFiles`), so no meta sync is required for this companion.

### DIFF
--- a/proofs/Proofs/Erdos1110OQ01.lean
+++ b/proofs/Proofs/Erdos1110OQ01.lean
@@ -726,4 +726,15 @@ neither base is a power of the other — so it lies outside the base-power famil
 example : Set.Infinite (NonRepresentable 6 4) :=
   infinite_nonRepresentable_of_common_dvd (d := 2) (by norm_num) (by norm_num) (by norm_num)
 
+/-- **The identity is the unique unit exponent pair.** For distinct primes `p ≠ q`,
+`p^k q^l = 1` iff both exponents vanish — the exponent-uniqueness readout at the monoid
+identity `1 = p^0 q^0`. So `(0, 0)` is the only exponent vector representing the unit of
+the power-form submonoid. -/
+theorem powerForm_eq_one_iff {p q : ℕ} (hp : p.Prime) (hq : q.Prime) (hpq : p ≠ q)
+    (k l : ℕ) : p ^ k * q ^ l = 1 ↔ k = 0 ∧ l = 0 := by
+  constructor
+  · intro h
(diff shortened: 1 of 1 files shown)


rjwalters:
Rebased onto main and trimmed to the genuinely-new content. `powerForm_antichain_iff` was dropped: main already carries the logically-equivalent `powerForm_incomparable_iff` (proofs/Proofs/Erdos1110OQ01.lean:246, same LHS mutual-non-divisibility, RHS in negated-≤ form vs this PR's strict-order form — equivalent by linear arithmetic). `powerForm_eq_one_iff` is new and kept (verified: single-file elaboration exit 0, 0 new axioms). The problem-state JSON resolved to main's version, which has moved on since this branch was cut.

PR #37452, opened 2026-07-10
Enrich Fibonacci Gap-Two Product-Sum: annotate import region

rjwalters:
Enrichment pass for **fibonacci-identities-oq-05-oq-03** (quality 95, passes 0).

This entry's meta.json is already fully saturated (all overview/section/conclusion fields deep, under all size caps at ~16.8 KB). The one genuine gap was **annotation coverage**: the existing annotations began at line 3 (the module docstring), leaving **line 1 (`import Mathlib`) uncovered**.

## Change
- Added `ann-fiboq0503-imports` (range: line 1) — a `context`-type annotation explaining what the blanket gallery import actually pulls in: the 0-indexed `Nat.fib` API and `Nat.fib_add_two`, the `Finset.sum_range_succ` big-operator layer, the parity primitives (`Nat.even_or_odd`, `Nat.even_add_one`, `Even/Odd.neg_one_pow`), and `linear_combination`/`omega`. Notes that Mathlib records Cassini only over `Int.fib`, so the `Nat.fib` Cassini and d'Ocagne identities are proved from scratch here rather than imported.
- `meta.json` untouched — already saturated and within all guardrail caps.

## Validation
- JSON validates; `check-meta-size.ts` passes (file ≤ 60 KB).
- Enum values (`type: context`, `significance: context`) normalize via `normalize-enums.ts` exactly as the sibling annotations already in the file do.

### DIFF
--- a/src/data/proofs/fibonacci-identities-oq-05-oq-03/annotations.json
+++ b/src/data/proofs/fibonacci-identities-oq-05-oq-03/annotations.json
@@ -1,5 +1,22 @@
 [
   {
+    "id": "ann-fiboq0503-imports",
+    "proofId": "fibonacci-identities-oq-05-oq-03",
+    "range": { "startLine": 1, "endLine": 1 },
+    "type": "context",
+    "title": "The Blanket `import Mathlib`",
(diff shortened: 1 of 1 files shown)


rjwalters:
Superseded by #37447, which merged first with an equivalent (and richer) import-region annotation for the same slug fibonacci-identities-oq-05-oq-03. Closing to avoid a duplicate/conflicting annotation on line 1. No further change needed — main already covers the gap.

PR #37650, opened 2026-07-11
research(erdos-3): repair Mathlib drift — restore Erdos3Problem + Erdos3LogHarmonic to verified

rjwalters:
## Summary

**Both erdos-3 files were `[UNVERIFIED]` and did not compile on `main`.** The drift accumulated across several commits landed while docker was down; the prebuilt oleans were stale-but-valid, masking the breakage from CI. This PR repairs the full dependency chain so the flagship builds again end-to-end.

## `Erdos3Problem.lean`

- **Unterminated comment** — a stray *duplicate* `/--` docstring at line 925 opened a nested comment that was never closed, **silently swallowing the entire second half of the file** (every ordering-lattice lemma, the Green–Tao bridge, `erdos_3_open`, …). Removed the duplicate.
- `Real.isLittleO_log_rpow_atTop` → `isLittleO_log_rpow_atTop` (the lemma now lives in the root namespace).
- `Set.indicator_of_mem hj` → `Set.indicator_of_mem hj full` — the function argument could not be inferred ("typeclass instance problem is stuck").
- Removed a redundant `ring` after a `field_simp` that already closes the goal ("no goals to be solved").

## `Erdos3LogHarmonic.lean`

- **Parse errors** (`unexpected token '('`): a multi-line `exact_mod_cast <term>` where the term continued onto the next line inside a parenthesised `by` — the term parser stopped at the line break. Joined onto one line.
- The two threshold lemmas' `have`s failed with `set`-fold and a `c ^ 2` ℝ/ℕ exponent mismatch; rewrote them as `ℕ` inequalities (`le_max_*`, `Nat.le_add_*`) then cast.

## Verification

- `./bin/lake env lean Proofs/Erdos3LogHarmonic.lean` → **0 errors**.
- `./bin/lake env lean Proofs/Erdos3Problem.lean` (against a freshly-built LogHarmonic olean) → **0 errors**, only the expected off-limits `sorry` warning.
- `#print axioms` on all repaired sorry-free theorems (ordering lattice, `sharp_required_bound_implies_conjecture`, both LogHarmonic summability lemmas): foundational only `[propext, Classical.choice, Quot.sound]`.

The single remaining `sorry` is the threshold-critical `required_bound_implies_conjecture` — as hard as Erdős #3 itself, deliberately left open. `meta.lineCount` 1224→1221; `status`/`sorries`/`axiomCount` unchanged (now actually accurate).

### DIFF
--- a/proofs/Proofs/Erdos3LogHarmonic.lean
+++ b/proofs/Proofs/Erdos3LogHarmonic.lean
@@ -277,13 +277,15 @@ theorem summable_one_div_nat_mul_log_mul_const {c δ : ℝ} (hc : 0 < c) (hδ : 
   have hc2 : (0 : ℝ) < c ^ 2 := by positivity
   -- threshold `N₀` above `2`, `2/c` and `1/c²` (so `m·c ≥ 2` and `m·c² ≥ 1` on the tail)
   set N₀ : ℕ := max 2 (⌈2 / c⌉₊ + ⌈1 / c ^ 2⌉₊) with hN₀def
-  have hN₀2 : (2 : ℝ) ≤ (N₀ : ℝ) := by exact_mod_cast le_max_left 2 (⌈2 / c⌉₊ + ⌈1 / c ^ 2⌉₊)
-  have hN₀2c : 2 / c ≤ (N₀ : ℝ) :=
-    le_trans (Nat.le_ceil _) (by exact_mod_cast le_trans (Nat.le_add_right _ _)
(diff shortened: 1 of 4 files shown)


rjwalters:
Superseded by #37590, which independently repaired the same broken erdos-3 chain (Erdos3Problem + Erdos3LogHarmonic) and added the SuperSharp reduction. Main already compiles cleanly (verified `lake env lean` exit 0), so this PR's fixes are redundant. Closing.

PR #37838, opened 2026-07-12
research(descartes-rule-of-signs-oq-01-oq-03): sign changes of a nowhere-zero sequence are adjacent [VERIFIED axiom-free]

rjwalters:
## Summary

Extends the saturated `DescartesRuleOfSignsOQ01OQ03` file with a new **§ 9** isolating the structural fact behind Descartes' sign-variation count: for a **nowhere-zero** real sequence, a sign change can only occur between *consecutive* indices, so `V` reduces to the textbook adjacent-flip count.

## New theorems (both axiom-free, general in `n`)

- **`signChangeBetween_adjacent_of_nonzero`** — if `f : Fin n → ℝ` never vanishes, any `SignChangeBetween f i j` has `j = i + 1`. This is the structural core previously buried *inline* inside `countSignChanges_alternating`, now a reusable standalone lemma.
- **`countSignChanges_eq_card_adjacent_of_nonzero`** — for a nowhere-zero `f`, `countSignChanges f` equals the number of consecutive sign flips (`k < n − 1` with `f k · f (k+1) < 0`, expressed via a total zero-extension of `f`). Shows the file's general jump-over-zeros pair definition of `V` coincides with the classical adjacent-flip definition **exactly when no entry vanishes**, and properly generalises `countSignChanges_alternating` (every adjacent pair alternating ⟹ all `n − 1` flips present ⟹ count `n − 1`).

## Why this is a genuine gap

The file had the *maximal* case (`countSignChanges_alternating`, value `n − 1`) and the *minimal* case (`countSignChanges_eq_zero_of_nonneg/nonpos`, value `0`) but no exact count for an **arbitrary** nowhere-zero sequence. This fills that middle, and the adjacency lemma that powers both was never exposed on its own.

## Verification

- Compiled with `bin/lake env lean Proofs/DescartesRuleOfSignsOQ01OQ03.lean` (rc 0) against the canonical `origin/main` base.
- `#print axioms` on both new theorems: `[propext, Classical.choice, Quot.sound]` only — no `sorryAx`, no `Lean.ofReduceBool`.
- `axiomCount` unchanged (1 — the `SturmReduction` bridge structure); meta `theoremCount`/`lineCount` resynced.

### DIFF
--- a/proofs/Proofs/DescartesRuleOfSignsOQ01OQ03.lean
+++ b/proofs/Proofs/DescartesRuleOfSignsOQ01OQ03.lean
@@ -1009,4 +1009,100 @@ theorem signChangesInCoeffs_reverse_neg {p : ℝ[X]} (h0 : p.coeff 0 ≠ 0) :
   have h0' : (-p).coeff 0 ≠ 0 := by rw [coeff_neg]; exact neg_ne_zero.mpr h0
   rw [signChangesInCoeffs_reverse h0', signChangesInCoeffs_neg]
 
+/- ## § 9. Sign changes of a nowhere-zero sequence are adjacent (verified, axiom-free)
+
+The `§ 6` sharpness result `countSignChanges_alternating` assumed *every* adjacent
+pair alternates and concluded the count is `n − 1`.  The structural fact powering it —
+that when no entry vanishes a sign change can only occur between *consecutive* indices —
+is more basic and holds with no alternation hypothesis at all.  We expose it as a
+standalone lemma and then use it to compute `countSignChanges` of an arbitrary
+nowhere-zero sequence *exactly*, as the number of consecutive sign flips.
+
+Classical reading: the textbook definition of "the number of sign variations" counts
+adjacent sign flips of the (nonzero) coefficient list; the file's `countSignChanges`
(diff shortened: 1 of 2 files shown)


rjwalters:
Doctor: closing as superseded. The content of this PR landed on main via the later-merged general Fin n sign-change work (PR #38513) in proofs/Proofs/DescartesRuleOfSignsOQ01OQ03.lean:

- `countSignChanges_eq_card_adjacent_of_nonzero` is equivalent to main's `countSignChanges_nowhere_zero` (line 1137): countSignChanges of a nowhere-zero sequence = card of adjacent pairs (j = i+1) with f i * f j < 0. Main's pair-set formulation and this PR's range(n-1) zero-extension formulation count the same set (main additionally has `card_adjacent`, line 1166, giving the n-1 ceiling).
- `signChangeBetween_adjacent_of_nonzero` (gap forces an interior zero) is embedded verbatim as the forward direction of `countSignChanges_nowhere_zero` (the by_contra/omega adjacency step), and its consequences (`countSignChanges_le_of_nowhere_zero`, alternating max, constant-sign zero) are all on main.

No genuinely-new content remains.

PR #38472, opened 2026-07-13
feat: blocked-route registry — reopen-criterion structure for tracker blockers

rjwalters:
## Summary

Implements the curated spec of #38388: blocker entries in `src/data/research/problems/*.json` (`currentState.blockers`) may now be either a legacy plain string (valid forever) or a structured `BlockerEntry` `{ route, reopenCriterion, blockedAt? }` operationalizing the "materially new mechanism required" reopen bar from #38385/#37505. No big-bang data migration — zero data files touched.

## Changes

- **`src/types/research.ts`** — `BlockerEntry` interface (required `route` + `reopenCriterion`, optional `blockedAt`), `Blocker = string | BlockerEntry` union, `ResearchState.blockers: Blocker[]` (was `string[]`, already violated by 14 wildcat objects in the wild), `DEFAULT_REOPEN_CRITERION` constant, and the tolerant `blockerText` helper (single source of truth: strings verbatim; objects as `route — reopen: criterion`; wildcat fallback through `description`/`summary` then `JSON.stringify` — never `[object Object]`).
- **`src/pages/ResearchProblemPage.tsx`** — blockers list renders via `blockerText` (fixes the latent `[object Object]` bug for the 14 existing wildcat entries across 7 files) plus an `Array.isArray` guard: 88 tracker files carry a string `currentState` and 1 a string `blockers`, all of which would throw at the blockers line today.
- **`.claude/commands/lean-research.md`** + **`.lean/roles/researcher.md`** — object-form JSON shape documented directly after the equivalent-strength-check reopen-bar prose, with the enforcement half (do not re-attempt a blocked route unless its `reopenCriterion` is met; entries without a criterion, including all legacy strings, default to "materially new mechanism required"); one-line pointers added at the independence-preservation bullets. #38385/#38389 text untouched.
- **No change (verified per spec)**: `scripts/research/build.ts` still emits `string[]` (assignable to `Blocker[]`); `ContributeSection.tsx` is a prose mention only.

## Test plan (all run in the worktree at origin/main `c93a9396cf`)

1. **Parse-all**: `find src/data/research/problems -name '*.json' | xargs jq -e .` — all 3,033 files parse.
2. **Census re-run**: 638 array-form entries = 624 strings + 14 objects (matches the curated census exactly).
3. **Typecheck**: `tsc -b --force` green (after restoring the two untracked generated manifest artifacts from the main workspace — their absence is a pre-existing generated-file gap, unrelated).
[...]

PR #38486, opened 2026-07-13
fix: harden blockerText against null/non-object blocker entries

### LINKED ISSUE #38478: Harden blockerText against null/non-object blocker entries
## Context

While evaluating PR #38472 (blocked-route registry, issue #38388), I ran the `blockerText` helper in `src/types/research.ts` against adversarial shapes beyond the `Blocker = string | BlockerEntry` union:

- number, boolean, array, nested object → safe (`JSON.stringify` fallback, no `[object Object]`)
- `null` / `undefined` → **throws** `TypeError: Cannot read properties of null (reading 'route')` at the `entry.route` access

`Array.isArray(problem.currentState.blockers)` in `ResearchProblemPage.tsx` guards the array itself but not its elements, so a tracker JSON containing `"blockers": [null]` would crash the page — the same class of off-schema-data crash PR #38472 exists to prevent (88 string-`currentState` files + 1 string-`blockers` file appeared organically; a stray `null` from a jq edit is plausible).

**Current corpus status**: 0 nulls across all 638 blocker entries in 3,033 `src/data/research/problems/*.json` files (verified 2026-07-12), and `null` is outside the declared `Blocker` type — so this is defensive hardening, not a live bug. The reference helper in issue #38388's spec has the identical null behavior, so PR #38472 is spec-compliant as shipped.

## Suggested fix (one line)

At the top of `blockerText`, before the `entry` cast:

```ts
if (b === null || typeof b !== 'object') return JSON.stringify(b) ?? String(b)
```

(`JSON.stringify(undefined)` returns `undefined`, hence the `?? String(b)` tail.)

Discovered during code evaluation of PR #38472.

rjwalters:
Closes #38478

## What

Adds a null/non-object guard at the top of `blockerText` in `src/types/research.ts` (the helper introduced by PR #38472), per the issue's suggested fix:

```ts
if (b === null || typeof b !== 'object') return JSON.stringify(b) ?? String(b)
```

Previously `null`/`undefined` array elements threw `TypeError: Cannot read properties of null (reading 'route')` at the `entry.route` access, so a tracker JSON containing `"blockers": [null]` would crash the research problem page. Numbers/booleans/arrays/nested objects already fell through safely.

## Rendering choice

Off-schema non-objects render as their JSON (`"null"`, `"undefined"`, `"42"`) rather than an empty string or a `"(invalid blocker entry)"` marker — this follows the issue's suggested fix verbatim, matches the helper's existing `JSON.stringify` fallback style for wildcat objects, and keeps malformed data visible for debugging instead of silently hiding it. (`?? String(b)` covers `undefined`, where `JSON.stringify` returns `undefined` instead of a string.)

## Verification

Adversarial harness (from the issue) + full corpus scan:

| case | result |
|---|---|
| `null` | `"null"` — no throw |
| `undefined` | `"undefined"` — no throw |
| `42` | `"42"` |
| `true` | `"true"` |
| `[1, 'x']` | `'[1,"x"]'` |
| nested object | JSON, no `[object Object]` |
| `{}` | `"{}"` |
| structured entry | `"r — reopen: c"` |
| legacy string | verbatim |

Corpus: 3,033 `src/data/research/problems/*.json` files scanned, 638 blocker entries — **0 throws, 0 `[object Object]`**.

Typecheck: isolated `tsc --noEmit` on the touched file passes; full `tsc -b` passes after copying the three untracked generated manifests (`data-manifest.json`, `listings.json`, `research-data-manifest.json`) from the main tree (known generated-file gap in fresh worktrees; not committed — they are gitignored).

### DIFF
--- a/src/types/research.ts
+++ b/src/types/research.ts
@@ -229,9 +229,12 @@ export const DEFAULT_REOPEN_CRITERION = 'materially new mechanism required'
  * - Pre-schema "wildcat" objects (ad-hoc keys already in the data) fall back
  *   through `description`/`summary` before JSON.stringify, so nothing ever
  *   renders as "[object Object]".
+ * - Off-schema non-objects (null/undefined/number/…) render via JSON.stringify
+ *   so malformed data stays visible instead of throwing or vanishing.
  */
 export function blockerText(b: Blocker): string {
   if (typeof b === 'string') return b
+  if (b === null || typeof b !== 'object') return JSON.stringify(b) ?? String(b)
   const entry = b as Partial<BlockerEntry> & { description?: unknown; summary?: unknown }
   const label =
     (typeof entry.route === 'string' && entry.route) ||


rjwalters:
✅ **Approved!**

Reviewed the 3-line hardening of `blockerText` in `src/types/research.ts` (Closes #38478).

## Correctness

- **Guard ordering is right.** After the `typeof b === 'string'` early return, the new guard `b === null || typeof b !== 'object'` catches every remaining non-object (`null`, `undefined`, numbers, booleans — and incidentally symbols/functions). Everything that falls through is a non-null object (including arrays), so the subsequent `b as Partial<BlockerEntry>` cast and `entry.route` / `entry.reopenCriterion` accesses can never throw.
- **`?? String(b)` tail types cleanly.** `tsc -b` exits 0 and `eslint src/types/research.ts` passes with no unnecessary-condition complaint. At runtime it correctly covers `undefined` (where `JSON.stringify` returns `undefined` rather than a string).
- **Sole caller** (`ResearchProblemPage.tsx:314`) renders the return value as an `<li>` — always-string output is exactly what it needs.

## Test Execution (harness reproduced independently)

| case | result |
|---|---|
| `null` | `"null"` — no throw |
| `undefined` | `"undefined"` — no throw |
| `42`, `0`, `NaN`, `true`, `false` | JSON text, no throw |
| `[1,'x']`, `[]` | JSON text |
| nested object, `{}` | JSON, no `[object Object]` |
| structured entry / wildcat `description`/`summary` / legacy string | unchanged behavior |
| `{route: 7, reopenCriterion: 'c'}` | falls to JSON label + reopen suffix, no throw |

17/17 pass — 0 throws, 0 `[object Object]`.

**Corpus scan (reproduced):** 3,033 `src/data/research/problems/*.json` files, 638 blocker entries → 0 throws, 0 `[object Object]`. Matches the PR's numbers exactly.

## Scope & CI

- Diff touches exactly one file: doc-comment bullet + one guard line. Nothing else.
- No GitHub CI checks are configured on this branch (`gh pr checks`: "no checks reported"), so local `tsc -b` + `eslint` + harness stand in — all green. `mergeStateStatus: CLEAN`.

The rendering choice (JSON.stringify keeps malformed data visible rather than hiding it) follows the issue's suggested fix verbatim and matches the helper's existing fallback style. Acceptance intent of #38478 — helper never throws on arbitrary JSON — is met: all six JSON value kinds (string/number/boolean/null/array/object) are covered.

PR #38505, opened 2026-07-13
fix: restore + track .loom/hooks (settings.json hook scripts, incl. guard-destructive safety hook)

rjwalters:
Completes a known gap of #38398: the mass deletion removed `.loom/hooks/*`; R1 (#38423) excluded them as gitignored runtime, but `.claude/settings.json` (restored the same day) wires three of them as live hooks — so every prompt/tool-use since has fired missing-file hook errors, and the `guard-destructive.sh` PreToolUse safety hook has been silently offline.

- Restores all 5 files byte-identical from `dc9fdffa30^` (exec bits preserved, smoke-tested: both UserPromptSubmit hooks exit 0 on sample JSON)
- Adds `!.loom/hooks/` gitignore exception (same engine-config rationale as `!.loom/roles/`)
- Working copies already restored to the operator's main checkout to stop the error noise immediately

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #38554, opened 2026-07-13
fix: add required hookEventName to guard-destructive hook output

rjwalters:
Follow-up to #38505: the restored `guard-destructive.sh` predates a Claude Code hooks-spec change — `hookSpecificOutput` now requires `hookEventName`. Deny/ask verdicts were failing output validation and being dropped (observed live: three "validation failed" errors during today's HOME cleanup — the hook fired correctly on `rm -rf` commands but its verdict was discarded).

3-line fix + doc comment: `hookEventName: "PreToolUse"` added to both jq emit paths and both raw-JSON fallbacks. Tested: deny path emits spec-valid JSON (sample in commit), allow path silent exit 0, `bash -n` clean.

### DIFF
--- a/.loom/hooks/guard-destructive.sh
+++ b/.loom/hooks/guard-destructive.sh
@@ -20,7 +20,7 @@
 #   - Allow: Everything else (exit 0, no output)
 #
 # Output format (Claude Code hooks spec):
-#   { "hookSpecificOutput": { "permissionDecision": "deny|ask", "permissionDecisionReason": "..." } }
+#   { "hookSpecificOutput": { "hookEventName": "PreToolUse", "permissionDecision": "deny|ask", "permissionDecisionReason": "..." } }
 #
 # Error handling: This script MUST never exit with a non-zero code or produce
 # invalid output. Any internal error is caught by the trap, logged for
@@ -75,6 +75,7 @@ deny() {
     local reason="$1"
     if jq -n --arg reason "$reason" '{
         hookSpecificOutput: {
+            hookEventName: "PreToolUse",
             permissionDecision: "deny",
             permissionDecisionReason: $reason
         }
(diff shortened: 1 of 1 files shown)


rjwalters:
## Judge Review: APPROVED

Proportionate review of this ~5-line follow-up to #38505.

**Scope verified**: Diff touches only `.loom/hooks/guard-destructive.sh` (+5/-3) — the doc comment, both `jq -n` emit paths (deny/ask), and both raw-JSON fallbacks. Nothing else.

**Reproduced against the PR head (31213a4), script extracted to scratch (main tree untouched)**:
- Deny path (`rm -rf ~`, `rm -rf /`): emits `hookSpecificOutput.hookEventName: "PreToolUse"` alongside `permissionDecision`/`permissionDecisionReason`, exit 0
- Ask path (`git reset --hard`): same, with `permissionDecision: "ask"`
- Allow path (`ls -la`): silent, exit 0
- Raw-JSON fallbacks (exercised via wrapper jq failing only on `-n`): both deny and ask fallbacks emit valid JSON (validated with `jq -e`) containing `hookEventName`
- jq fully absent: script fails open with logged error, as designed (pre-existing behavior, fallback correctly reserved for emit-time jq failure)
- `bash -n`: clean

**Spec check**: `hookEventName` must echo the hook event name per the Claude Code hooks output spec; `"PreToolUse"` is correct here. Consistent with the repo's other hooks (`skill-router.sh`, `methodology-inject.sh` both emit `hookEventName: "UserPromptSubmit"` for their event).

**CI/merge state**: No PR-triggered workflows exist in this repo, so "no checks reported" is expected; `mergeStateStatus` = CLEAN, MERGEABLE.

Correct, minimal fix that restores enforcement of deny/ask verdicts. Approving.

Issue #37508, opened 2026-07-11
Epic: Upgrade Lean toolchain + Mathlib pin (v4.26.0 → latest) across proofs, Docker image, and gallery metadata


rjwalters:
## Goal

Upgrade the repo from Lean `v4.26.0` / Mathlib `2df2f0150c27` to the latest stable Lean 4 + Mathlib (as of filing: upstream projects like openai/cdc-lean are pinned to `v4.31.0` / `9a9483a9`).

## Motivation

- **CDC port cost collapses** (#37507): the kernel-verified Cycle Double Cover proof (openai/cdc-lean, 7,134 lines) is written against v4.31.0/`9a9483a9`. On a matching pin it drops in nearly verbatim, discharging the axiom in the `cycle-double-cover` entry (#37506) and unlocking proof-level annotations of the novel 8-flow→double-cover argument.
- **External-result velocity**: every external integration (AlphaProof Nexus #20732, First Proof #21946, now CDC) pays a porting tax proportional to our pin's age. Staying ~5 months behind makes each one an epic.
- **General maintenance**: Mathlib deprecation cycles eventually remove old names; the longer we wait, the larger the jump.

## Scope / touch points

- `proofs/lean-toolchain` (v4.26.0), `proofs/lakefile.toml` mathlib rev, `proofs/lake-manifest.json`
- Docker build image: `IMAGE="lean4-arm64:v4.26.0"` in `proofs/scripts/docker-build.sh` (+ Dockerfile in `proofs/`); shared volumes `lean-mathlib-cache` / `lean-mathlib-packages` self-heal on manifest change but will do a full cold refill (~10GB; note the host's chronic disk pressure — budget headroom first)
- **All Lean files under `proofs/Proofs/`** (~thousands): expect lemma renames, simp-set drift, deprecations. Bulk of the work.
- `src/data/proofs/*/meta.json` `mathlib_version` fields (~4,800 entries) — mechanical sweep at the end
- Aristotle integration: confirm the Aristotle backend supports the new toolchain before flipping (its generated proofs must elaborate on our pin)

## Suggested plan

1. **Spike branch**: bump pins, build the safe subset (`build-safe-subset.sh`) in Docker, collect the failure inventory. This yields a real cost estimate instead of a guess.
2. **Triage**: classify failures (rename-only / tactic drift / genuinely broken). Rename-only fixes are Mechanic-agent fodder; batch them.
[...]

PR #38652, opened 2026-07-15
migrate(v4.31): Doctor increment 47 — tm/pd/rewrite (A-M mechanical seam) (+4 GREEN)

rjwalters:
Doctor increment 47 of the v4.26→v4.31 migration (epic #37508, issue #38065). Partition: A–M basenames + Erdos<600. TAIL phase — mining the thin seam of mechanically-fixable residual clusters.

## +3 GREEN this increment

- **CauchySchwarzOQ01OQ01OQ01** (proof-drift, 6→0): root cause was output-only implicit `𝕜` in `projCoeff`/`orthProj`/`orthResidual`/`gramSchmidt3` — a stuck metavar in v4.31 (`InnerProductSpace ?m E`, first/third args metavars). Made `𝕜` an explicit named parameter of the four defs and supplied it at all call sites. Proof-drift repairs: `field_simp`+`ring` for div-cancel goals; `RCLike.norm_conj` + `RCLike.norm_ofReal` (via ℝ-recast) replacing a looping norm `simp`; direct `sub_eq_zero.mp`; strengthened `nlinarith` with `norm_nonneg` lemmas.
- **ArithmeticSeriesOQ00OQ02OQ01** (instance-synth, 6→0): `Ico 1 (n + 1)` inside a ℤ/ℚ-valued sum body elaborated its index type as ℤ/ℚ (`HAdd ℕ ℕ ℤ` / `LocallyFiniteOrder ℚ`). Pinned with `Ico (1 : ℕ) (n + 1)`; downstream omega/linarith were cascades.
- **BallotProblemOQ03OQ03** (unknown-const + rewrite-drift, 4→0): `lgvDet_nonneg` now takes 5 order hypotheses — added the three missing ones to the `lgv_nonneg_standard` wrapper **statement** (strengthening to intended-true; the bare form is not provable). `Nat.one_le_succ` → `Nat.succ_le_succ (Nat.zero_le _)`. `lgv_dyck_pair_count`: normalized `choose` indices via omega-`show`s + `push_cast; ring`.

## Statement repairs

- `lgv_nonneg_standard` (BallotProblemOQ03OQ03): added the 3 ordering hypotheses that the underlying `lgvDet_nonneg` requires. Strengthening only — no weakening/sorry/axiomatize.

## Recipes

- Output-only implicit type-class parameters (appearing only in a def's body/return) are stuck metavars in v4.31 → promote to an explicit named parameter and thread it through every call site. `(𝕜 := 𝕜)` named-arg pinning does NOT work when the implicit is auto-bound.
- `Ico 1 (n + 1)` in a coerced sum body: pin the index type `Ico (1 : ℕ) …` to stop it elaborating at the body's field.
- `RCLike.norm_ofReal` can loop in `simp`; recast `(‖v‖^2 : 𝕜) = ((‖v‖^2 : ℝ) : 𝕜)` then `rw [RCLike.norm_ofReal, abs_of_nonneg …]`.

## Honest remaining A–M seam estimate

~663 A–M residuals. Probed ~17 files this session: error counts of 4–8 are common but most are genuine multi-site proof rewrites (e.g. AmgmInequalityOQ02OQ01OQ02OQ01OQ03 reveals 6+ delicate induction drift sites once the first blocker clears — reverted). The mechanical seam is <10% of residuals and thinning; each green now costs real proof surgery. Several probed at 10–18 errors (ArchimedesMethodOfExhaustion 17, BorsukUlamOQ02 18, AlgebraicNumbersCountableOQ04 12). Best remaining seam signature: instance-synth / elaboration-order single-root failures and unknown-const renames.

### DIFF
--- a/proofs/Proofs/ArithmeticSeriesOQ00OQ02OQ01.lean
+++ b/proofs/Proofs/ArithmeticSeriesOQ00OQ02OQ01.lean
@@ -95,7 +95,7 @@ theorem nicWeight_three : nicWeight 3 = 27 / 5 := by unfold nicWeight; norm_num
     Combining: 3·w(2) = 8, which has no solution in ℤ (proved by omega). -/
 theorem no_integer_weight (w : ℕ → ℤ)
     (h : ∀ n, ∑ k ∈ Ico 1 (n + 1), w k * (2 * (k : ℤ) - 1) =
-             (∑ k ∈ Ico 1 (n + 1), (k : ℤ))^2) : False := by
+             (∑ k ∈ Ico (1 : ℕ) (n + 1), (k : ℤ))^2) : False := by
   have hh1 := h 1
   have hh2 := h 2
   -- Normalize n+1 to concrete values and expand finite sums
@@ -118,7 +118,7 @@ theorem no_integer_weight (w : ℕ → ℤ)
     nicWeight at k=1 and k=2. Forced by the n=1 and n=2 cases via telescoping. -/
 theorem nicWeight_unique_at_12 (w : ℕ → ℚ)
     (h : ∀ n, ∑ k ∈ Ico 1 (n + 1), w k * (2 * (k : ℚ) - 1) =
-             (∑ k ∈ Ico 1 (n + 1), (k : ℚ))^2) :
+             (∑ k ∈ Ico (1 : ℕ) (n + 1), (k : ℚ))^2) :
     w 1 = nicWeight 1 ∧ w 2 = nicWeight 2 := by
   have hh1 := h 1
   have hh2 := h 2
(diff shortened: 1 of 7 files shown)


rjwalters:
Added +1: **BezoutIdentityOQ02OQ04** (instance-synth, 5→0) — Polynomial.content/IsPrimitive.mul/content_mul now require [NormalizedGCDMonoid R] not bare [GCDMonoid R]; widened constraint on gauss_lemma_primitive_mul + content_multiplicative, and Polynomial.content_mul now takes its args implicitly (dropped 'f g'). Total this increment: **+4 GREEN**.

PR #39068, opened 2026-07-17
fix(proofs): unblock Erdos10Incomplete01OQ02 + WIP01OQ03 omega-with-mod-241 hang under v4.31

### LINKED ISSUE #39059: Erdos10Incomplete01OQ02 (+Erdos10WIP01OQ03): decide/fin_cases kernel-reduction hang under v4.31 (migration residual)
## Summary
`proofs/Proofs/Erdos10Incomplete01OQ02.lean` (only 149 lines, a pure cascade whose imports are all GREEN) does not compile under v4.31: it hangs at 100% CPU. Two verify attempts — one **3h41m**, one **>22min** — made zero progress before being killed. Its ledger class is `slow-timeout`, and multiple migration agents misread this as a missing-parent cascade and punted; the real cause is kernel-level `decide` reduction blowup in this file.

## Diagnosis
- The file contains `fin_cases hq <;> decide` (lines ~52, 56) and a `decide` inside an evenness proof (~line 103).
- Under v4.31, one or more of these `decide` calls no longer reduces cheaply (the toolchain tightened `decide`/kernel reduction — the same class that broke `native_decide`/`decide` on `ℤ Finset.Icc`/noncomputable paths elsewhere in this migration). The proposition being decided apparently forces an expensive/non-terminating reduction.
- Parent `Erdos10OQ01Incomplete01` is GREEN; this is NOT a dependency block.

## Recommended fix
Identify which `decide`/`fin_cases <;> decide` hangs (bisect by replacing each with `sorry` to isolate), then replace it with an explicit term proof, `omega`/`norm_num`, or `Decidable.decide`-with-`Nat.rec` as appropriate — do not rely on bare `decide` for the offending goal. Small file, so once the culprit is found the fix should be quick.

## Dependent (cascades once this is fixed)
- `Erdos10WIP01OQ03` (`slow-timeout`) imports this file and is gated on it — should go green as a pure cascade once the parent compiles (verified: it has 0 edits needed, only needs the parent's olean).

## Scope
Genuine per-file investigation, not bulk migration drift. Blocks epic #37508 tail; group with #38612. Relates #37508 #38612.

rjwalters:
## Summary
`proofs/Proofs/Erdos10Incomplete01OQ02.lean` (and its dependent `Erdos10WIP01OQ03.lean`) hung at 100% CPU under the v4.31 toolchain — prior verify attempts ran **3h41m** and **>22min** with zero progress. Both now build green.

## Root cause (bisected)
The hang was **not** `decide`/`fin_cases` as the ledger class suggested. It was **`omega` running with the six covering-system residue hypotheses in context** — `n % 3 = 1`, `n % 7 = 1`, `n % 5 = 2`, `n % 17 = 8`, `n % 13 = 11`, and especially **`n % 241 = 121`**. Under v4.31 `omega` collects every in-scope arithmetic hypothesis; the mod-241 constraint makes its decision procedure blow up combinatorially.

Bisection (short-timeout docker builds):
- Stubbing the two `fin_cases … <;> decide` theorems did **not** stop the hang.
- Stubbing all bodies → compiles in seconds (rules out imports/statements).
- Restoring only `not_mem_one_of_even` → hang returns.
- `rw [mem_one_iff]` alone → fast; the first (prime) branch's `omega` → hang.
- Adding `clear h3 h7 h5 h17 h13 h241` before that `omega` → **compiles**. Confirmed.

## Fix (targeted; 0 sorries, 0 axioms)
- `clear` the residue hypotheses before every parity `omega` that doesn't need them (they are only required to feed `prime_forced_small`).
- Use `ring` / `Nat.add_sub_cancel'` for the commutativity rewrite `n = 2^a + p`, which must run while the residues are still in scope.
- Derive `2 < n` by term (`lt_of_lt_of_le … hbig.le`) instead of a poisoned `omega`.
- Replace the flagged `fin_cases <;> decide` with `simp [Nat.odd_iff]` / `omega`, and the `Even 2` `decide` with the explicit witness `⟨1, rfl⟩`.

## Dependent — NOT a pure cascade
`Erdos10WIP01OQ03` was expected to cascade green once the parent compiled, but it has the **identical** six-mod-hypothesis + `omega` pattern in `not_isPrimePlusKPowers_one_of_even_covering` and hung on its own. It received the same residue-clear fix.

## Verification
Both build green via the docker wrapper (never bare `lake build`):
- `./proofs/scripts/docker-build.sh Proofs.Erdos10Incomplete01OQ02` → success (2971 jobs)
- `./proofs/scripts/docker-build.sh Proofs.Erdos10WIP01OQ03` → success (2979 jobs, 1.1s)

0 sorries, 0 `axiom` declarations in both files.

Closes #39059

### DIFF
--- a/proofs/Proofs/Erdos10Incomplete01OQ02.lean
+++ b/proofs/Proofs/Erdos10Incomplete01OQ02.lean
@@ -49,11 +49,11 @@ open Erdos10Incomplete01 Erdos10OQ01Incomplete01
 
 /-- Each of the six covering primes `{3, 5, 7, 13, 17, 241}` is odd. -/
 theorem coveringPrimes_odd {q : ℕ} (hq : q ∈ coveringPrimes) : Odd q := by
-  fin_cases hq <;> decide
+  fin_cases hq <;> simp [Nat.odd_iff]
 
 /-- Each of the six covering primes is at most `241`. -/
 theorem coveringPrimes_le {q : ℕ} (hq : q ∈ coveringPrimes) : q ≤ 241 := by
-  fin_cases hq <;> decide
+  fin_cases hq <;> omega
 
 /-! ## Bridge: `prime_forced_small` restated against `sumPrimeAndTwoPows 1` -/
 
@@ -69,7 +69,9 @@ theorem mem_one_forced {n : ℕ}
   rw [mem_one_iff] at hn
   rcases hn with hp | ⟨p, a, hp, hrep⟩
   · exact Or.inl hp
-  · have hrep' : n = 2 ^ a + p := by omega
+  · -- `omega` would be poisoned by the mod-241 hypothesis in context, so use `ring` for the
+    -- purely commutative rewrite `p + 2^a = 2^a + p` (see below on `not_mem_one_of_even`).
+    have hrep' : n = 2 ^ a + p := by rw [hrep]; ring
     exact Or.inr ⟨a, p, prime_forced_small h3 h7 h5 h17 h13 h241 hp hrep', hrep'⟩
 
 /-! ## The parity refinement: even numbers in the class are outside level 1 -/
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge Review: APPROVE ✅

Structural review (no rebuild — docker Lean builds are memory-costly and the fix is verifiable from the diff; the builder's reported `docker-build.sh` successes at 2971/2979 jobs stand as runtime evidence).

**Root cause confirmed.** The six covering-system residue hypotheses (`n % 3=1`, `n % 7=1`, `n % 5=2`, `n % 17=8`, `n % 13=11`, `n % 241=121`) were in scope when parity `omega` calls fired; under v4.31 `omega` ingests all in-scope arithmetic hyps and the mod-241 constraint blows up its decision procedure. The bisection narrative (stub-and-restore isolating `not_mem_one_of_even`, then `clear` fixing it) is consistent with the diff.

**Correctness of the fix — sound, no goal weakening:**
- The `clear h3 h7 h5 h17 h13 h241` calls are placed *after* `prime_forced_small` has consumed the residue facts (the only consumer), so the cleared hypotheses are genuinely unneeded at every downstream `omega`. Verified in all three sites (both branches of `not_mem_one_of_even` and the body of `not_isPrimePlusKPowers_one_of_even_covering`).
- `have hrep' : n = 2^a + p := by rw [hrep]; ring` — pure commutativity, same conclusion `omega` produced.
- First branch: `rcases hp.eq_two_or_odd with h | h <;> omega` correctly derives `False` from `heven`/`hbig` in both cases (n=2 vs n%2=1); equivalent to the old `eq_one_or_self_of_dvd` argument.
- WIP03: `hrep` via `Nat.add_sub_cancel' hwn` (needs `2^a ≤ n`, supplied by `hwn`) and the `2 < n` bound via `lt_of_lt_of_le (by norm_num) hbig.le` are both term-level and correct.
- `fin_cases hq <;> simp [Nat.odd_iff]` / `<;> omega` and `Even 2 := ⟨1, rfl⟩` are faithful replacements for the flagged `decide`s (these standalone lemmas carry no poisoned residue context, so `omega` there is cheap).

**Soundness gates:** grep of both files confirms 0 `sorry`, 0 `axiom` declarations, 0 `native_decide`, and no residual `decide` in code (only doc/comment mentions). Merge state CLEAN, mergeable. No Lean-build CI checks are configured for this repo (expected for proof PRs); docker-wrapper verification is the accepted path.

**`Closes #39059` justified:** both the primary `Erdos10Incomplete01OQ02.lean` and the dependent `Erdos10WIP01OQ03.lean` (which carries the identical mod-241+`omega` pattern in `not_isPrimePlusKPowers_one_of_even_covering` and hung on its own — not a pure cascade) receive the residue-clear fix.

Clean, well-commented, minimal (+23/-12). Approved.

PR #39069, opened 2026-07-17
fix(agents): make Claude token registry honest + add consistency check (#38967)

### LINKED ISSUE #38967: Claude token pool: .env ACCOUNT_TOKEN_FILE_* names don't match .loom/tokens/ files (registry column is dead)
## Summary

The Claude account pool in `.env` and the actual token files in `.loom/tokens/` disagree on filenames — the `ACCOUNT_TOKEN_FILE_*` registry column points at files that do not exist. Every registry entry resolves to a missing file, so the email→token-file mapping the registry claims to provide is unusable.

## Details

- `.env` declares the account pool as `ACCOUNT_EMAIL_N` / `ACCOUNT_KEY_N` / `ACCOUNT_TOKEN_FILE_N` triples. The `ACCOUNT_TOKEN_FILE_N` values are semantic, account-derived filenames (one per account).
- `.loom/tokens/` actually contains the same number of `*.token` files, but named **positionally** (`agent-1.token` … `agent-N.token`).
- **Overlap between the two sets is zero.** Every `ACCOUNT_TOKEN_FILE_N` value in `.env` resolves to a non-existent path under `.loom/tokens/`. (Checked all entries: registry count == token-file count, matches == 0.)

## Root cause

`scripts/agents/claude-wrapper.sh` never resolves `ACCOUNT_TOKEN_FILE_N`. It:

- **Bootstraps** missing tokens by writing them **positionally** as `.loom/tokens/agent-$n.token` (from the `ACCOUNT_KEY_N` values), ignoring the registry's filename column.
- **Selects/rotates** by globbing the directory (`_all_tokens=("$_tokens_dir"/*.token)`), plus optional `.ranking` / `.allowlist` helpers — again keyed off the positional `agent-N` basenames, not the registry names.

So rotation works fine (the wrapper is self-consistent with its own positional naming), but the `ACCOUNT_TOKEN_FILE_N` column in `.env` is stale documentation that describes filenames the system never creates or reads.

## Impact

- **Low/latent, not a live outage** — the wrapper doesn't depend on the broken column, so account rotation is unaffected today.
- **The account→token mapping is effectively unknowable from the registry.** Anything (a tool, a script, or a human debugging "which token belongs to account X?") that trusts `ACCOUNT_TOKEN_FILE_N` to locate a specific account's token gets a dead path, and there is no reliable way to tell which positional `agent-N.token` corresponds to which account/email.
- Anyone copying this pool elsewhere by following the registry (rather than globbing the dir) would copy zero valid files — which is exactly how this was discovered (attempting to replicate the pool by resolving `ACCOUNT_TOKEN_FILE_N`).

## Suggested fix (pick one)

1. **Make bootstrap honor the registry:** have `claude-wrapper.sh` write each token to its `ACCOUNT_TOKEN_FILE_N` filename instead of `agent-$n.token`, so files and registry agree and the email→file mapping is real.
2. **Or drop the dead column:** if the wrapper is intentionally positional-only, remove/deprecate `ACCOUNT_TOKEN_FILE_N` from `.env` (or regenerate its values to the real `agent-N.token` names) so the registry stops claiming a mapping the code ignores. Keep `ACCOUNT_EMAIL_N` as the human label.
3. Either way, add a small consistency check (bootstrap or a lint) that fails when a referenced `ACCOUNT_TOKEN_FILE_N` has no matching file.

## Notes

- No credential values are included in this report — only counts and filename *shapes*.
- Discovered while building a feature in a sibling repo that replicates this exact pool pattern; treating the `.token` files (glob) as the source of truth, per the wrapper, is the reliable approach.

rjwalters:
Closes #38967

## Problem

`.env` declares the Claude account pool as `ACCOUNT_EMAIL_N` / `ACCOUNT_KEY_N` / `ACCOUNT_TOKEN_FILE_N` triples, but `scripts/agents/claude-wrapper.sh` bootstraps and rotates tokens **positionally** as `.loom/tokens/agent-N.token` and never resolves `ACCOUNT_TOKEN_FILE_N`. The registry's semantic filenames (`robb-2amlogic.token`, ...) don't exist on disk — verified 13/13 miss, 0 overlap. Rotation works (it globs the dir), but the account to token-file mapping the registry claims to provide is a dead column.

## Approach (safe — the pool is LIVE)

Deliberately **not** the issue's option 1. I did not rename token files, change how the wrapper writes tokens, or change how rotation globs — that risks breaking the running pool. Instead I made the registry honest and added a guard so it can't silently rot again:

- **`scripts/agents/check-token-registry.sh`** (option 3): fails with a non-zero exit when any `ACCOUNT_TOKEN_FILE_N` has no matching file under `.loom/tokens/`. Also reports orphan token files. Never reads, prints, or logs `ACCOUNT_KEY_*` values or token contents — only filenames/existence. Wire into CI/pre-flight lint.
- **`scripts/agents/sync-token-registry.sh`** (option 2): regenerates the `ACCOUNT_TOKEN_FILE_N` values to the real positional `agent-N.token` names. The wrapper writes `ACCOUNT_KEY_N` to `agent-N.token`, so account N's token is always `agent-N.token`, corresponding to `ACCOUNT_EMAIL_N` — the mapping is knowable, the column just had the wrong names. Edits `.env` in place with a timestamped backup; touches only the filename column, never `ACCOUNT_KEY_*`. Supports `--dry-run`.
- **`claude-wrapper.sh`**: comment documenting the positional convention and that `ACCOUNT_TOKEN_FILE_N` is a label-only column that must equal `agent-N.token`.
- **`.env.example`**: placeholder template (no secrets) documenting the pool format with the correct `agent-N.token` convention, plus a `.gitignore` negation so the template is tracked while `.env` / `.env.*` stay ignored.

## Operator action required (manual, local — not committable)

`.env` is gitignored and holds secrets, so this PR cannot correct the live file. After merge, run locally in the repo root:

```bash
./scripts/agents/sync-token-registry.sh          # rewrites ACCOUNT_TOKEN_FILE_N -> agent-N.token (with .env.bak backup)
./scripts/agents/check-token-registry.sh          # should now print OK
```

This only rewrites the filename column; `ACCOUNT_KEY_*` and the live `.loom/tokens/*.token` files are untouched, so the running pool is unaffected.

## Validation

- `check-token-registry.sh` against the real `.env`: correctly flags 13/13 stale entries and exits 1.
- Fixture round-trip: sync rewrites the column, re-check passes, `ACCOUNT_KEY_*` values preserved, sync is idempotent.
- Both new scripts are shellcheck-clean. The two SC2012 infos in `claude-wrapper.sh` are pre-existing and unrelated to the one comment added here.
- No secrets committed: `git check-ignore` confirms `.env` / `.env.bak-*` ignored, `.env.example` tracked and scanned clean of real keys.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,61 @@
+# .env.example — placeholder template. Copy to .env and fill in real values.
+#
+#   cp .env.example .env
+#
+# .env is gitignored and holds SECRETS. NEVER commit .env. This template
+# contains only placeholder values and documents the expected variables.
+
+# ── Harmonic Aristotle API ─────────────────────────────────────────────
+ARISTOTLE_API_KEY=arstl_your_key_here
+
+# ── Cloudflare deploy target ───────────────────────────────────────────
+# Create a token at: https://dash.cloudflare.com/profile/api-tokens
+# Template: "Edit Cloudflare Workers" + Pages:Edit + D1:Edit
+CLOUDFLARE_ACCOUNT_ID=your_account_id
+CLOUDFLARE_API_TOKEN=your_api_token
+
+# ── Mathstodon (Mastodon) API — Herald agent ───────────────────────────
+MATHSTODON_TOKEN=your_mastodon_token
+
+# ── Claude Code OAuth token pool — autonomous agents ───────────────────
+# Long-lived (1 year) OAuth tokens. scripts/agents/claude-wrapper.sh selects
+# and rotates among these for autonomous agent sessions.
+#
+# HOW THE POOL ACTUALLY WORKS (read before editing):
+#   * The token FILES in .loom/tokens/*.token are the source of truth the
+#     wrapper reads (it selects by globbing the directory).
+#   * On first run the wrapper BOOTSTRAPS missing tokens by writing each
+#     ACCOUNT_KEY_N value to  .loom/tokens/agent-N.token  — POSITIONALLY,
+#     using the same index N. It does NOT use ACCOUNT_TOKEN_FILE_N.
+#   * Therefore the token file for account N is always  agent-N.token,
+#     and it corresponds to ACCOUNT_EMAIL_N / ACCOUNT_KEY_N.
+#
+# Each account is a parallel triple (same index N):
+#   ACCOUNT_EMAIL_N       human label for the account (for your reference)
+#   ACCOUNT_KEY_N         the OAuth token value                    (SECRET)
+#   ACCOUNT_TOKEN_FILE_N  the token filename on disk — MUST be "agent-N.token"
+#                         so the registry matches what the wrapper creates.
+#
+# Keep ACCOUNT_TOKEN_FILE_N == agent-N.token. Run
+#   ./scripts/agents/check-token-registry.sh
+# to verify the registry agrees with .loom/tokens/, and
+#   ./scripts/agents/sync-token-registry.sh
+# to regenerate the ACCOUNT_TOKEN_FILE_N values if they drift. See #38967.
+
+# Account 1: you@example.com
+ACCOUNT_EMAIL_1=you@example.com
+ACCOUNT_KEY_1=your_oauth_token_here
+ACCOUNT_TOKEN_FILE_1=agent-1.token
+
+# Account 2: another@example.com
+ACCOUNT_EMAIL_2=another@example.com
+ACCOUNT_KEY_2=your_oauth_token_here
+ACCOUNT_TOKEN_FILE_2=agent-2.token
+
+# ...add ACCOUNT_EMAIL_N / ACCOUNT_KEY_N / ACCOUNT_TOKEN_FILE_N=agent-N.token
+# for each additional account, incrementing N.
+
+# ── Single-account fallback ────────────────────────────────────────────
+# If no ACCOUNT_KEY_* / .loom/tokens are present, the wrapper falls back to
+# this single token.
+# CLAUDE_CODE_OAUTH_TOKEN=your_oauth_token_here
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 # ── Secrets — never commit ─────────────────────────────────────────────
 .env
 .env.*
+# Exception: the placeholder template carries no secrets and is tracked.
+!.env.example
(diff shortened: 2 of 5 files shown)


rjwalters:
## Judge review — APPROVE ✅

Reviewed the full diff (5 files: `.env.example`, `.gitignore`, two new scripts, one comment in `claude-wrapper.sh`). This takes the correct safe path and I'm approving.

**Security (paramount here) — clean:**
- No secret material in the diff. The real `.env` is NOT staged (`gh pr view --json files` lists only the 5 intended paths). `.env.example` contains only placeholders (`arstl_your_key_here`, `your_api_token`, `your_oauth_token_here`, `agent-N.token`) — every `ACCOUNT_KEY_*` / API-token line is a placeholder, verified line-by-line.
- The `!.env.example` negation is correctly scoped: tested in an isolated repo with the PR's `.gitignore` — `.env`, `.env.local`, `.env.production`, and `.env.bak-*` all remain **IGNORED**; only `.env.example` becomes tracked. No accidental un-ignore of secret files.

**Correctness:**
- `check-token-registry.sh` inspects only `ACCOUNT_TOKEN_FILE_N` filenames/existence against `.loom/tokens/`, never reads `ACCOUNT_KEY_*` or token contents, exits 1 on mismatch (2 on env error, 0 on absent `.env`). Orphan reporting is informational and doesn't affect exit code — correct given the wrapper globs the dir.
- `sync-token-registry.sh` rewrites only `ACCOUNT_TOKEN_FILE_N` lines to `agent-N.token`, passes every other line (incl. `ACCOUNT_KEY_*`) through verbatim, is idempotent (`changes=0` short-circuit), writes a timestamped `.bak`, and supports `--dry-run`.
- Both scripts are **shellcheck-clean** (verified locally).
- Live pool is genuinely untouched: the `claude-wrapper.sh` change is a comment only — no change to how tokens are written or globbed.

**Process:** merge state CLEAN, mergeable, base `main`. No CI checks reported (scripts/docs-only change triggers no workflow) — acceptable, nothing failing. The required manual operator step (`.env` can't be fixed in-commit; run the two scripts locally post-merge) is clearly documented in the PR body.

Applying `loom:pr`.

Issue #38578, opened 2026-07-13
Sperner housekeeping: legacy SpernerGrid.lean v4.26 bit-rot + duplicate GridSimplex/IsCanon scaffolds


rjwalters:
Spun out of #8998 increment 1 (PR #38570, merged 2026-07-13). Three housekeeping items discovered while building the machine-checked d=1,N=1 faithfulness counterexample:

1. **`proofs/Proofs/SpernerGrid.lean` does not compile on the pinned v4.26 toolchain** — ~23 pre-existing errors, all inside the `interiorFlip`/`gridAdj_symm`/`gridAdj_vertex` proof blocks (exactly the oriented-adjacency code the #8998 redesign replaces). Decide: repair, quarantine from any build lists, or deprecate in favor of `SpernerGridBase.lean` once the redesign completes. Note the gallery's `sperner-mathlib4` verified badge applies to the abstract proof (`SpernerMathlib4.lean`), not this file — no overclaim today.
2. **Stale duplicate `GridSimplex` in `SpernerGridCell.lean`** — converge with `SpernerGridBase`'s canonical copy.
3. **Second `IsCanon` scaffold in `SpernerNDimOQ02Cell.lean`** — converge with Base's `IsCanon` layer.

Sequencing: best done after (or with) #8998 increment 2 (general-d,N canonical facet-sharing involution), since the redesign determines what survives.

Refs: #8998 (epic), PR #38570 (counterexample increment), #7938 (parent epic).

rjwalters:
Builder status (PR #39073) — investigated all three items against **current `main`** (v4.31, not the issue's stale v4.26 assumption):

**Item 1 — legacy `SpernerGrid.lean`: DELIVERED (quarantine/deprecate).** File still does not compile on v4.31; errors are confined to the oriented `gridAdj`/`boundary_doors_odd` block that #8998 replaces, and it's not in any migration ledger. Per the "redesign determines what survives" sequencing, repairing was not appropriate. Added a `DEPRECATED` header notice pointing to canonical `Proofs.SpernerGridBase`, and excluded `SpernerGrid` (+ the content-free `SpernerGridAristotle` placeholder that only imports it) from `build-safe-subset.sh`. No overclaim: the `sperner-mathlib4` badge is on `SpernerMathlib4.lean`.

[...]

PR #39073, opened 2026-07-18
chore(sperner): quarantine non-compiling legacy SpernerGrid.lean (Part of #38578)

rjwalters:
## Summary

Sperner housekeeping (#38578), three items — **item 1 delivered**, item 2 already resolved upstream, item 3 deferred to the #8998 redesign. This is a **documentation + build-tooling only** change (no proof or gallery edits).

### Item 1 — legacy `SpernerGrid.lean` (DELIVERED)

`proofs/Proofs/SpernerGrid.lean` does not compile. Its errors are confined to the oriented `interiorFlip`/`gridAdj`/`boundary_doors_odd` block, whose `boundary_doors_odd` claim is mathematically **false** for `d = 1` — exactly the code the #8998 `GridSimplex`/`gridAdj` redesign replaces. The defect predates and is unaffected by the v4.26→v4.31 migration (#37508); it is not in any migration repair/deferral ledger.

Per the issue's own sequencing ("the redesign determines what survives"), the correct low-risk action is to **quarantine/deprecate**, not repair code slated for replacement:

- Added a prominent `DEPRECATED` notice to the file header pointing to the canonical compiling foundation `Proofs.SpernerGridBase` (+ `SpernerGridCell` / `SpernerNDimOQ02Cell` for the Option-C instance) and to #8998.
- Excluded `SpernerGrid` from `proofs/scripts/build-safe-subset.sh` (exact-match, so `SpernerGridBase`/`Cell`/`FaithfulD1` are untouched). Also excluded `SpernerGridAristotle` — a content-free placeholder whose only `import` is the broken file, so it can never build while `SpernerGrid` is quarantined.

No overclaim exists today: the gallery's `sperner-mathlib4` verified badge applies to `Proofs.SpernerMathlib4`, not this file.

### Item 2 — duplicate `GridSimplex` in `SpernerGridCell.lean` (ALREADY RESOLVED)

Already fixed on `main` by the v4.31 migration (#38065). `SpernerGridCell.lean` no longer redeclares `GridSimplex`; it now retains only the `BaryPoint.transfer*` helpers and relies on Base's canonical copy, with an in-file comment documenting the dedup. Nothing to do.

### Item 3 — second `IsCanon` scaffold in `SpernerNDimOQ02Cell.lean` (DEFERRED to #8998)

This is **not** a trivial duplicate and is **not** safe to force-converge:

- `SpernerNDimOQ02.IsCanon` uses its own lex order `SpernerGrid.BaryPoint.lexLe`; Base's `SpernerGrid.IsCanon` uses `SpernerGrid.BaryPoint.lexLE` — different predicates in different namespaces, no hard conflict, so both compile independently.
- Converging them requires proving the two lex orders agree (or rewriting the scaffold onto Base's `IsCanon`/`lexLE` and its `IsCanon.base_unique` uniqueness development) — a semantic/architectural change, not mechanical dedup.
- `SpernerNDimOQ02Cell.lean`'s own docstring flags "per-geometry uniqueness of `IsCanon`" as **open next-session work** completing the Option-C `SpernerTriangulation` instance.

That makes item 3 a #8998 increment-2 decision ("the redesign determines what survives"), so it is deferred with this note rather than forced.

## Testing

- `bash -n proofs/scripts/build-safe-subset.sh` passes; exclusion matching simulated (only `SpernerGrid` + `SpernerGridAristotle` excluded, Base/Cell/FaithfulD1 unaffected).
- No Lean build required: the SpernerGrid.lean change is a comment added to an already-non-compiling file; no proof, meta, or gallery files change.

Part of #38578 (items 2/3 tracked above; issue remains open for item 3 under #8998).

### DIFF
--- a/proofs/Proofs/SpernerGrid.lean
+++ b/proofs/Proofs/SpernerGrid.lean
@@ -6,6 +6,33 @@ Authors: RJ Walters
 import Mathlib
 
 /-
+# ⚠ DEPRECATED / QUARANTINED — does not compile (see #38578, #8998)
+
+This legacy file **does not compile**. Its errors are confined to the
+*oriented* `gridAdj` / `interiorFlip` / `boundary_doors_odd` block, whose
+`boundary_doors_odd` claim is mathematically **false** as stated for `d = 1`
+because of a structural double-counting in the `GridSimplex` / `gridAdj`
+representation. That defect predates — and is unaffected by — the v4.26 → v4.31
+toolchain migration (#37508); it can only be resolved by the `GridSimplex` /
+`gridAdj` redesign tracked in #8998 (see also #9044).
+
+Do **not** repair this file in place: the oriented adjacency machinery here is
+exactly the code the #8998 redesign replaces. It is deliberately excluded from
+`proofs/scripts/build-safe-subset.sh` so the safe-subset build is not polluted
+with its known failures.
+
+**Canonical replacement:** the compiling `BaryPoint` / `GridSimplex` foundation
+now lives in `Proofs.SpernerGridBase` (the unoriented cell geometry, canonicality
+`IsCanon`, and lexicographic order), extended by `Proofs.SpernerGridCell` and
+`Proofs.SpernerNDimOQ02Cell` for the "Option C" unoriented `SpernerTriangulation`
+instance. New work should build on those, not on this file.
+
+The gallery's `sperner-mathlib4` verified badge applies to the *abstract* proof
+in `Proofs.SpernerMathlib4`, **not** to this file — so no verification overclaim
+exists today.
+
+--------------------------------------------------------------------------------
+
 # Grid Triangulation Instance for Abstract Sperner's Lemma
 
 We construct a concrete `CellComplex` instance from the standard
--- a/proofs/scripts/build-safe-subset.sh
+++ b/proofs/scripts/build-safe-subset.sh
@@ -1,17 +1,28 @@
 #!/usr/bin/env bash
 #
-# Build all Lean proofs EXCEPT memory-intensive ones
+# Build all Lean proofs EXCEPT memory-intensive or known-broken ones
 #
 # Some proofs (like Erdos728FactorialDivisibility) have tactics that can
-# consume unbounded memory. This script builds everything else.
+# consume unbounded memory. Others are legacy files that are known not to
+# compile and are pending a redesign rather than a repair. This script builds
+# everything else.
 #
 set -euo pipefail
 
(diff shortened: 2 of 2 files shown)


rjwalters:
## Loom Judge — APPROVE ✅

Docs + build-tooling only (42/-4), merge state CLEAN, mergeable; no Lean build triggered (expected — no proof/meta/gallery edits). Verified all four review points:

**Load-bearing safety point (exact-match exclusion).** The exclusion loop is `for exc in "${EXCLUDE[@]}"; do if [[ "$name" == "$exc" ]]`. The RHS `"$exc"` is **quoted**, so `==` inside `[[ ]]` is a literal string comparison (no glob). `"SpernerGrid"` therefore matches **only** the module named exactly `SpernerGrid`. Confirmed the canonical foundation files `SpernerGridBase`, `SpernerGridCell`, `SpernerGridFaithfulD1` (all distinct module names) are **not** caught and remain in safe-subset CI coverage. A too-broad prefix match would have silently dropped them; the anchoring prevents that.

**No overclaim / coverage gap.** The gallery `sperner-mathlib4` entry is `meta.status: verified`, `badge: original`, with primary `leanFile.path = Proofs/SpernerMathlib4.lean` (0 sorries, 0 axioms). The verified badge rests on `SpernerMathlib4.lean`, **not** the deprecated `SpernerGrid.lean` — no meta references `SpernerGrid.lean` at all. Deprecating it changes no verification claim.

**`SpernerGridAristotle` exclusion is sound.** The file is content-free (`namespace SpernerGrid` / `end`, zero theorems) and its only real import is the broken `SpernerGrid`, so it can never build while `SpernerGrid` is quarantined. It's listed as an additionalFile of the verified `sperner-mathlib4` entry, but since it carries no proof content, excluding a guaranteed-failure from the safe subset removes noise, not coverage.

**Deprecation vs deletion / imports.** Correct call given the #8998 redesign determines what survives. The only files importing `import Proofs.SpernerGrid` (exact) are `SpernerGridAristotle.lean` (also excluded) and the whole-tree aggregate `AngleTrisection…Aristotle.lean` — both already non-compiling because of the same broken file; this PR doesn't worsen that pre-existing state and touches no downstream that depends on `SpernerGrid` proper (real geometry dependents use `SpernerGridBase`).

**Scoping honest.** "Part of #38578", item 2 already resolved by #38065, item 3 explicitly deferred to #8998 with the lex-order (`lexLe` vs `lexLE`) convergence reasoning. No unrelated changes, no secrets.

Approving. Applying `loom:pr`.

Issue #39077, opened 2026-07-19
Gallery integrity repairs: fix the 28 never-compiled Lean entries (proof-repair follow-on to #39061)


rjwalters:
## Summary

Proof-repair follow-on to #39061. The **metadata half** of #39061 is done — PR #39067 retracted the invalid `verified`/`original` badges on the 24 gallery entries that map to the 28 never-compiled Lean modules (e.g. `kummer-theorem-oq-02` de-badged to `formalized`/`wip`). This issue tracks the **remaining half**: actually *repairing the Lean proofs* so the entries compile on v4.31, at which point their badges can be legitimately restored.

These 28 entries never compiled even on v4.26 — under the old `autoImplicit true` default, undefined identifiers were silently bound as opaque implicits, making the stated theorem vacuously general so the proof "passed" against a free variable. v4.31 (autoImplicit off) correctly rejects them, exposing genuine defects.

### Class A — undefined-identifier / autoImplicit-masked + real proof errors (22)
Per-file: bind/define the flagged identifier, then fix the real errors (unproved `omega` goals, application type-mismatches, etc.), then restore/correct `meta.json`.
- [ ] Erdos1015Problem (`:n`), Erdos1054OQ02 (`:q`), Erdos1066Problem (`:n`), Erdos1071Problem (`:s`), Erdos1100Problem (`:p`), Erdos1100ProblemProvable (`:p`), Erdos1119Problem (`:ι`), Erdos1153Problem (`:k`), Erdos409Problem (`:p`), Erdos423Problem (`:n`), Erdos451Problem (`:hp`), Erdos476Problem (`:B`), Erdos485OQ01 (`:d`), Erdos529Problem (`:d_2`), Erdos701Problem (`:A`), Erdos713Problem (`:n`), Erdos719Problem (`:H.edges`), Erdos770Problem (`:X`), Erdos831Problem (`:p_e1`), Erdos832Problem (`:V`), Erdos856Problem (`:A`), Erdos882ProblemOQ03 (`:a`), KummerTheoremOQ02 (`:n`), RamseysTheoremOQ04 (`:n`)

### Class B — parse error (1)
- [ ] Erdos608Problem — `unexpected token '('` (~line 136) + unproved omega goals.

### Class C — self-reported unverified (1)
[...]

Issue #39061, opened 2026-07-17
Gallery integrity: 28 entries never compiled even on v4.26 (autoImplicit-masked / broken) — metadata audit + repair


rjwalters:
## Summary
The v4.26→v4.31 migration surfaced **28 gallery entries that were already broken before the migration** — they never compiled even on the old toolchain. These are **not** migration drift; they are pre-existing defects whose gallery `meta.json` may falsely claim `verified`/`original`. This is a **credibility audit**, tracked separately from the migration epic #37508.

They sit in `proofs/batch2/verify-results.tsv` (on `feature/issue-37508`) with status `PRE-EXISTING`.

## Root cause (verified by spot-check)
22 of 28 are classed `never-compiled:<var>`. Under v4.31 they fail with **`error(lean.unknownIdentifier): Unknown identifier 'n'`** (or `p`, `A`, `V`, …). Under v4.26's old `autoImplicit true` default those undefined identifiers were silently bound as opaque implicit variables → the stated theorem became **vacuously general** and the proof 'passed' against a free variable → an invalid `verified` badge. v4.31 (autoImplicit now off by default) correctly rejects them. Most also carry additional genuine defects (unproved `omega` goals, application type-mismatches, parse errors), i.e. they are broken proofs, not just spelling drift.

**Implication:** any of these 28 currently badged `verified`/`original` in `src/data/proofs/<name>/meta.json` was never validly verified. Each must be re-badged (`formalized` with sorries, or `axiomatized`, or fixed) and its `assumptions` corrected.

### Class A — undefined-identifier / autoImplicit-masked + real proof errors (22) — per-file: define/bind the identifier, fix the real errors, correct meta.json
- [ ] Erdos1015Problem (`:n`)
- [ ] Erdos1054OQ02 (`:q`)
- [ ] Erdos1066Problem (`:n`)
- [ ] Erdos1071Problem (`:s`)
- [ ] Erdos1100Problem (`:p`)
- [ ] Erdos1100ProblemProvable (`:p`)
- [ ] Erdos1119Problem (`:ι`)
- [ ] Erdos1153Problem (`:k`)
- [ ] Erdos409Problem (`:p`)
- [ ] Erdos423Problem (`:n`)
- [ ] Erdos451Problem (`:hp`)
- [ ] Erdos476Problem (`:B`)
- [ ] Erdos485OQ01 (`:d`)
- [ ] Erdos529Problem (`:d_2`)
[...]

PR #39096, opened 2026-07-19
Fix: repair never-compiled Erdos701Problem (batch 6 of #39077)

rjwalters:
## Summary

Batch 6 of the never-compiled repair effort (#39077). Repairs **Erdos701Problem** (Chvátal's conjecture on hereditary/intersecting families — Class A).

## Fixes

`set_option autoImplicit false` added (this project inherits Lean core's `autoImplicit true`, so the defect was masked — see the root-cause note on #39077). That exposed:

1. **`singleton_intersecting`** — `subst hB hC` eliminated the theorem parameter `A` (the RHS of `B = A` / `C = A`), which was then reused at line 100. autoImplicit had silently re-bound the dangling `A` as a fresh free variable. Rewrote the tail as `rw [hB, hC, Set.inter_self]; exact hA`.
2. **`uniform_not_hereditary`** — genuine v4.31 Mathlib drift:
   - `Set.ncard_pos.mp` no longer resolves as a bare constant → `Set.nonempty_of_ncard_ne_zero`.
   - `Set.ncard_diff_singleton_of_mem` — finiteness is now an autoParam, so the explicit `hXfin` arg caused "function expected"; dropped it.
   - `simp [B] at heq; exact heq hx` restructured to `have hxB := heq hx; simp [B] at hxB` (robust to simp-normal-form changes).

## Evidence

`./proofs/scripts/docker-build.sh Proofs.Erdos701Problem`:
- **Before** (`autoImplicit false` probe): `error: Unknown identifier A` (×2), `Unknown constant Set.ncard_pos.mp`, plus type-mismatch/function-expected at 238/240.
- **After**: `=== Build succeeded ===`, 0 axioms, 0 sorries.

## Metadata

Chvátal's conjecture is an **open problem** — stated as the definition `ChvatalConjecture`, not proven. Per the Axiom Integrity Policy, the entry stays `axiomatized`/`wip` (axiomCount 0; the supporting lemmas are the verified content). Cleared the stale #39061 "never validly compiled / NOT machine-verified" retraction from `assumptions`; refreshed `lineCount` 284→286.

Part of #39077 (does not close it — ~14 Class A/B/D/E entries remain).

### DIFF
--- a/proofs/Proofs/Erdos701Problem.lean
+++ b/proofs/Proofs/Erdos701Problem.lean
@@ -37,6 +37,8 @@ import Mathlib.Data.Finset.Basic
 
 open Set Finset
 
+set_option autoImplicit false
+
 namespace Erdos701
 
 /-
@@ -94,8 +96,8 @@ theorem singleton_intersecting {α : Type*} (A : Set α) (hA : A.Nonempty) :
     IsIntersecting ({A} : Set (Set α)) := by
   intro B hB C hC
   simp only [Set.mem_singleton_iff] at hB hC
-  subst hB hC
-  exact ⟨A, Set.inter_self A ▸ hA⟩
+  rw [hB, hC, Set.inter_self]
+  exact hA
 
 /-
 ## Part III: Stars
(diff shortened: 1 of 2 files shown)


rjwalters:
Closing as superseded. `Erdos701Problem` was repaired and merged on main via **#39628** (commit 12f9ebacc9, 0 sorries). Umbrella #39077 is CLOSED. This PR is DIRTY and obsolete.

PR #39331, opened 2026-07-20
audit: erdos-882 — drop phantom div_free_implies_distinct + nonexistent examples/conjecture from prose

rjwalters:
## Gallery Integrity Fix

**Proof**: erdos-882 (Subset Sums Without Divisibility)
**Result**: prose overstatements corrected; no code/count changes.

Counts are accurate (9 defs / 4 thms / 2 axioms / 0 sorries) and badge/status (`axiom`/`axiomatized`) are correct. Three pieces of prose referenced declarations/content that **do not exist** in `Proofs/Erdos882Problem.lean`:

### 1. Phantom `div_free_implies_distinct` (proofStrategy)
Prose named ``div_free_implies_distinct`` as "the key bridge to Erdős #1". No such declaration exists — the file has the `DivisibilityFree` and `DistinctSubsetSums` **definitions** only, with no lemma linking them. The implication is folded into the axiom `upper_bound_refined`. Reworded to state the bridge is assumed, not proved.

### 2. Phantom "conjectured bound" + "Establishes" (upper-bounds section)
Claimed the section "Establishes the upper bound chain" and "Includes ... the conjectured $\log_2 n + O(1)$ bound". The bound is **axiomatized** (`upper_bound_refined`), not established, and no conjecture declaration exists (it appears only in the header docstring). Reworded to "States ... via the axiom" and noted the conjecture is docstring-only.

### 3. Phantom examples (examples section)
Section summary described concrete examples $A=\{1\}$, $A=\{2,3\}$ as if formalized. The file's `## Part 6: Examples` is an empty header comment with **zero** example declarations. Reworded as a placeholder with illustrative (non-formalized) examples.

---
Discovered during gallery integrity audit (reserve mode — re-verifying prose vs actual declarations).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #39277, opened 2026-07-20
audit: erdos-882 — drop phantom div_free_implies_distinct, correct axiomatized-chain prose

rjwalters:
## Gallery Integrity Audit — erdos-882 (reserve-mode re-audit)

**Badge/status** (`axiom` / `axiomatized`) and all counts are correct:
2 axioms (`upper_bound_refined`, `lower_bound_max`), 0 sorries, 4 theorems, 9 defs.

The prose overstated what is formalized:

| Field | Problem | Fix |
|-------|---------|-----|
| `proofStrategy` | Cited lemma **`div_free_implies_distinct`** as "the key bridge to Erdős #1" — **no such declaration exists**. `DistinctSubsetSums` is only *defined* (line 92), never linked to `DivisibilityFree`. | State the chain is *intended*, the implication *unproved*, the bound *axiomatized*. |
| `sections[upper-bounds]` | "Establishes the upper bound chain … `DistinctSubsetSums` ⇒ …" and "the conjectured log₂ n + O(1) bound" — nothing establishes the chain (upper bound = raw axiom), no conjecture declaration exists. | "States the intended chain"; dropped phantom conjecture ref. |
| `conclusion.implications` | "The formalization demonstrates that Problem #882 reduces to Problem #1 via this implication" — reduction is described, not formalized. | Corrected to note the implication is stated (not proved) and the bound axiomatized. |
| `lineCount` | 169 (overcount) | 164 (actual) in both meta blocks. |

No Lean source changes; meta.json prose only.

---
Discovered during gallery integrity audit.

PR #39374, opened 2026-07-20
audit: erdos-882 — drop phantom div_free_implies_distinct from prose

rjwalters:
## Integrity Issue

**Proof**: erdos-882 (Subset Sums Without Divisibility)
**Problem**: `proofStrategy` prose named a phantom declaration as the core bridge to Erdős #1.

## Evidence

**meta.json `proofStrategy` claimed**: "...where the first implication (`div_free_implies_distinct`) is the key bridge to Erdős #1..."

**Lean file shows**: `Erdos882Problem.lean` has no `div_free_implies_distinct` declaration. The divisibility-free ⇒ distinct-subset-sums implication is **not proved** — the file contains only:
- `def DistinctSubsetSums` (a Prop definition, line 92)
- a comment "Key observation: ValidSubset implies distinct subset sums"
- the full upper-bound chain is folded into the `upper_bound_refined` **axiom** (line 103)

The only fully proved lemma in the upper-bound region is `sum_bound` (σ(A) ≤ |A|·n).

## Fix applied

- Reworded `proofStrategy` to drop the phantom `div_free_implies_distinct`, describe the bridge via the real `DistinctSubsetSums` def, and state plainly that the chain is axiomatized in `upper_bound_refined` rather than proved as standalone lemmas.
- Corrected stale `lineCount` 169 → 164 (actual file length) in both `meta` and `leanFile` blocks.

## Verified accurate (no change)

2 axioms (`upper_bound_refined`, `lower_bound_max`), 4 theorems, 9 defs, 0 sorries, 0 native_decide, no True stubs. Badge `axiom` / status `axiomatized` remain correct.

---
Discovered during gallery integrity audit (reserve mode).

### DIFF
--- a/research/problems/shannon-channel-coding-awgn-oq-03-oq-01/knowledge.md
+++ b/research/problems/shannon-channel-coding-awgn-oq-03-oq-01/knowledge.md
@@ -230,3 +230,25 @@ Delivered:
 Prior session shipped work UNVERIFIED (manual review, docker down). `ShannonChannelCodingAWGNOQ03OQ01.lean`
 (370 L, Mathlib-only) verified via lean-elab ([[reference-docker-down-lean-elab-verification-path]]):
 EXIT 0, zero errors (2 benign warnings). Standing work confirmed correct (no bug). Marked completed.
+
+## Session 2026-07-20 (researcher-1) — nextSteps reconciliation (COMPLETE for scope)
+
+**Mode**: REVISIT (RICH re-serve) · **Outcome**: tracker reconciliation, no new theorems.
+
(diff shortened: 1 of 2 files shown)


rjwalters:
Auditor convergence note: overlaps #39277 and #39331 (same erdos-882 phantom `div_free_implies_distinct`). This PR fixes the phantom + `lineCount` but **misses** the `upper-bounds` "Establishes the chain" overstatement, the conclusion `implications` overstatement, and the `examples`-section placeholder issue. #39277 is the most complete of the three — recommend closing this as superseded in favor of it.

PR #39557, opened 2026-07-20
Enrich odd-abundant density family: reconcile stale v4.26.0 verification claim to v4.31.0

rjwalters:
## Summary

Two sibling gallery entries in the odd-abundant-density family carried a stale toolchain claim after the #37508 migration:

- \`abundant-number-oq-03-oq-01\`
- \`abundant-number-oq-03-oq-01-oq-03\`

Both have \`meta.mathlib_version: "4.31.0"\` and \`status: "verified"\` (migrated green on main), yet each entry's \`description\` still ended with **"verified against Mathlib v4.26.0"** — a prose artifact left behind when the migration meta-sweep bumped the version field but not the inline claim. This was an internal contradiction (version field vs. prose).

## Change

Reconciled the verification-version claim in both descriptions: \`v4.26.0\` → \`v4.31.0\`, matching each entry's \`meta.mathlib_version\`.

- No axiom/sorry/status changes (both remain 0-axiom, 0-sorry, verified).
- No mathematical content added or removed — correctness reconciliation only.
- JSON validated with \`jq\`; 0 remaining \`v4.26\` mentions in either meta.

Both entries were already at quality target (5 keyInsights, full-coverage sections with mathContext, rich conclusion/crossRefs/references), so no accretion was warranted beyond fixing the stale claim.


rjwalters:
Withdrawing: on review these fall in the SOFT historical-note class, not the hard contradiction class. The 'verified/compiled against Mathlib v4.26.0' claim is *consistent* with the entry's verified status — only the version number is stale. Per prior operator feedback, per-entry version-number corrections are cosmetic accretion (and a gallery-wide v4.26→v4.31 prose sweep is a mechanic/meta batch, cf. #39063, not enricher work). Closing to avoid churn.

PR #40355, opened 2026-07-21
audit(fix): cantor-diagonalization-oq004 badge + page-crashing originalContributions

rjwalters:
Audit of `cantor-diagonalization-oq004` (Quantitative Cantor: strict gap #A < #(A → Prop)) found two integrity issues, both fixed here.

## 1. Page-crashing `originalContributions` type (HIGH)
`originalContributions` was stored as a **plain string** in both `cantor-diagonalization-oq004` and its parent `cantor-diagonalization-oq-03-oq-01-incomplete-01-oq-01`. `src/types/proof.ts` types it `string[]`, and `ProofOverview.tsx:244-251` does:
```
meta.originalContributions.length > 0  // truthy for a 415-char string
  ... meta.originalContributions.map(...)  // TypeError: map is not a function
```
A JS string has no `.map`, so the proof page render throws. **3998/4000** gallery entries already use `string[]`; these 2 were the outliers. Fixed by wrapping the existing prose in a single-element array (content unchanged).

## 2. Badge `original` on a pure Mathlib wrapper (MEDIUM)
All 3 theorems reduce to Mathlib's `Cardinal.cantor` — the file's own docstring says *"No new mathematics is required… the content is that the open question is a one-line consequence of the standard cardinal Cantor theorem."* Per the tier rules (Tier B / failure mode #5: main result via a Mathlib call), badge changed `original` → `mathlib`, and the Mathlib-core dependency added to `assumptions`.

## Otherwise clean
0 axioms, 0 sorries, no `native_decide`; 3 theorems match `theoremCount`; description/conclusion prose already disclosed the Mathlib reliance honestly.

Includes the audit-tracker completion for oq004.

### DIFF
--- a/src/data/proofs/audit-tracker.json
+++ b/src/data/proofs/audit-tracker.json
@@ -30734,6 +30734,12 @@
       "lastAudited": "2026-07-21T09:52:35Z",
       "result": "clean",
       "issues": []
+    },
+    "cantor-diagonalization-oq004": {
+      "auditCount": 1,
+      "lastAudited": "2026-07-21T09:58:18Z",
+      "result": "issues-fixed",
+      "issues": []
     }
   },
   "version": 1,


rjwalters:
Superseded by consolidated PR (oq004 fixes + tracker folded in) to avoid tracker merge-race.

PR #41502, opened 2026-07-22
fix(launchers): reclaim same-branch worktree so singleton agents don't silently die (#39649)

### LINKED ISSUE #39649: Singleton agent launchers silently die when a stale/locked legacy worktree holds their branch (deployer 7-day outage)
## Summary
The deployer was **dead for 7 days (Jul 13 → Jul 20)** and never respawned, despite `.loom/lean-daemon-state.json` config having `deployer: 1`. A **537-PR backlog** accumulated (all MERGEABLE/CLEAN), with 0 merges in 24h, before it was noticed manually.

## Root cause
The `/Volumes/Stripe` worktree-root migration left the **old** deployer worktree registered at the legacy path `.loom/worktrees/deployer`, **locked** (`"active deployer build/deploy — do not sweep"`), still holding branch `feature/deployer`.

`scripts/deploy/launch-agent.sh` now targets `/Volumes/Stripe/lean-genius/deployer`. Its `create_worktree` 3-tier fallback all fail because `feature/deployer` is already checked out at the locked legacy path:
1. `git worktree add -b feature/deployer` → `fatal: a branch named 'feature/deployer' already exists`
2. `git worktree add feature/deployer` → already checked out elsewhere
3. `git branch -D feature/deployer` (can't delete a checked-out branch) → final **un-redirected** `git worktree add` throws

Under `set -euo pipefail` the launcher **dies before creating the tmux session**. The daemon backgrounds it (`./scripts/deploy/launch-agent.sh &`), so the non-zero exit is invisible and every 30-min respawn hits the same wall with **no log line**.

## Reproduced
```
ℹ Creating worktree for deployer at /Volumes/Stripe/lean-genius/deployer...
fatal: a branch named 'feature/deployer' already exists
```

## Manual fix applied (2026-07-20)
```
git worktree unlock .loom/worktrees/deployer
git worktree remove --force .loom/worktrees/deployer
git worktree prune
./scripts/deploy/launch-agent.sh
# launcher's own sync also failed → finished manually:
cd /Volumes/Stripe/lean-genius/deployer && git fetch origin main && git checkout -B feature/deployer origin/main
```

## Proposed durable fix
- `create_worktree` should detect a same-branch worktree registered at **any** path (`git worktree list --porcelain`), and reclaim/relocate it (unlock + remove + prune) before creating at the resolved root.
- The launcher must **not fail silently** under `set -e`: on worktree-setup failure, log to the daemon log and emit a non-zero the daemon surfaces.
- Same trap can hit **any singleton agent** (seeker, etc.) whose legacy locked worktree survived the migration — fix should live in a shared helper, not just the deployer.

## Impact
High — a silent multi-day deploy outage with an invisible growing backlog. Reliability/observability gap in the core orchestration path.

rjwalters:
Closes #39649

## Problem
A singleton-agent launcher (`scripts/deploy/launch-agent.sh` and siblings) checks out a fixed branch (e.g. `feature/deployer`) at a fixed target worktree path via a 3-tier `create_worktree` fallback. When that branch was **already checked out at a DIFFERENT (stale/locked/legacy) path** — left behind by the `/Volumes/Stripe` worktree-root migration — all three `git worktree add` attempts failed and the final **un-redirected** add threw under `set -euo pipefail`, killing the launcher **before** it created its tmux session. The daemon backgrounds the launcher (`launcher &`), so the non-zero exit was invisible and the agent stayed dead across every 30-min respawn. This caused a **7-day deployer outage** with a growing, unnoticed PR backlog.

## Fix
Two new functions in the shared helper `scripts/lib/worktree-cleanup.sh`:

- **`reclaim_branch_worktree <branch> <target_path>`** — finds any worktree holding `<branch>` at a path other than the target (`git worktree list --porcelain`) and reclaims it: `git worktree unlock` + `remove --force` + `prune`, freeing the branch. It **deliberately overrides locks** (unlike `remove_own_worktree`, which preserves them) because the locked legacy worktree is exactly what must be reclaimed. Safety bounds:
  - only ever touches a worktree holding the **same branch**;
  - never removes the **target path** (no-op success if the branch is already there);
  - never removes the **primary/main checkout** (guarded explicitly);
  - **warns** (does not abort) on uncommitted/unpushed work — these singleton branches are reset to `origin/main` every cycle, so a blocking stale worktree must be cleared for the agent to run; the warning makes the rare trade-off visible.
- **`log_worktree_fatal <log_file> <msg>`** — logs a `FATAL` line to **both stderr and the daemon log**, then `exit 1`, so a dead launcher leaves a trace instead of dying silently.

Every singleton launcher now calls `reclaim_branch_worktree` before its `add` attempts and routes final `add` failure through `log_worktree_fatal`:
`deploy`, `auditor`, `mechanic`, `aristotle`, `peer-reviewer`, and `research/launch-seeker` (which also now sources the shared lib).

Untouched per issue constraints: `scripts/lean/launch.sh` and any `*token-registry*`.

## Testing
Isolated-repo test (bare remote + main checkout + locked legacy worktree holding `feature/deployer`), 13/13 assertions pass:
- control: bare `git worktree add` fails while the branch is held elsewhere;
- reclaim frees the branch (legacy dir removed) and the subsequent add **succeeds**;
- no-op success when the branch is already at target / not held anywhere;
- refuses to reclaim the main checkout;
- `log_worktree_fatal` writes to log **and** stderr and exits 1.

`bash -n` and `shellcheck` clean on all 7 files under macOS bash 3.2.

### DIFF
--- a/scripts/aristotle/launch-agent.sh
+++ b/scripts/aristotle/launch-agent.sh
@@ -133,11 +133,19 @@ create_worktree() {
         git worktree remove "$WORKTREE_PATH" --force 2>/dev/null || rm -rf "$WORKTREE_PATH"
     fi
 
+    # Free the branch if a stale/locked/legacy worktree still holds it at another
+    # path (e.g. left by the /Volumes/Stripe migration). Without this, the
+    # `git branch -D` below is a no-op (can't delete a checked-out branch) and
+    # the `git worktree add` dies silently under `set -e` (issue #39649).
+    reclaim_branch_worktree "$BRANCH_NAME" "$WORKTREE_PATH" || \
+        log_worktree_fatal "$LOG_FILE" "could not reclaim '$BRANCH_NAME' from a stale worktree (see 'git worktree list')"
+
     git branch -D "$BRANCH_NAME" 2>/dev/null || true
 
     print_info "Creating worktree..."
     mkdir -p "$(dirname "$WORKTREE_PATH")"
-    git worktree add "$WORKTREE_PATH" -b "$BRANCH_NAME" main
+    git worktree add "$WORKTREE_PATH" -b "$BRANCH_NAME" main 2>/dev/null || \
+        log_worktree_fatal "$LOG_FILE" "worktree setup failed for '$BRANCH_NAME' at $WORKTREE_PATH (branch may be checked out elsewhere; see 'git worktree list')"
 
     # Symlink .lake for fast Lean builds
     if [[ -d "$REPO_ROOT/proofs/.lake" ]] && [[ -d "$WORKTREE_PATH/proofs" ]]; then
--- a/scripts/auditor/launch-agent.sh
+++ b/scripts/auditor/launch-agent.sh
@@ -108,13 +108,21 @@ create_worktree() {
 
     print_info "Creating worktree for auditor at $WORKTREE_PATH..."
 
-    # Try to create worktree
+    # Free the branch if a stale/locked/legacy worktree still holds it at another
+    # path (e.g. left by the /Volumes/Stripe migration); otherwise every add
+    # below fails and this backgrounded launcher dies silently (issue #39649).
+    reclaim_branch_worktree "$BRANCH_NAME" "$WORKTREE_PATH" || \
+        log_worktree_fatal "$LOG_FILE" "could not reclaim '$BRANCH_NAME' from a stale worktree (see 'git worktree list')"
+
+    # Try to create worktree. The final attempt must NOT die silently under
+    # `set -e` (backgrounded launcher, invisible non-zero exit — issue #39649).
     git worktree add "$WORKTREE_PATH" -b "$BRANCH_NAME" main 2>/dev/null || {
         # Branch might exist, try to use it
         git worktree add "$WORKTREE_PATH" "$BRANCH_NAME" 2>/dev/null || {
             # Remove and recreate branch
             git branch -D "$BRANCH_NAME" 2>/dev/null || true
-            git worktree add "$WORKTREE_PATH" -b "$BRANCH_NAME" main
+            git worktree add "$WORKTREE_PATH" -b "$BRANCH_NAME" main 2>/dev/null || \
+                log_worktree_fatal "$LOG_FILE" "worktree setup failed for '$BRANCH_NAME' at $WORKTREE_PATH (branch may be checked out elsewhere; see 'git worktree list')"
         }
     }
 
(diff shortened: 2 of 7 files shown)


rjwalters:
## Judge Review: APPROVED

Thorough, well-scoped, well-tested fix for the singleton-launcher silent-death trap (#39649). I reproduced the core scenario and all safety guards in an isolated bare-remote repo against this branch's `worktree-cleanup.sh`.

### Root cause fixed
`reclaim_branch_worktree <branch> <target>` parses `git worktree list --porcelain`, finds any worktree holding the branch at a non-target path, and unlock + `remove --force` + `prune`s it so the subsequent `git worktree add` succeeds instead of dying under `set -e`. Verified: with a **locked legacy** worktree holding `feature/deployer`, a bare add is blocked (control), reclaim frees it, and the add at the new target then succeeds.

### Safety of reclaim — all bounds hold
- **Exact ref match**: matches on `substr($0,8)=="refs/heads/<branch>"`, not a prefix. Confirmed a sibling `feature/deployer-x` worktree is left untouched when reclaiming `feature/deployer`.
- **Target path**: no-op success when the branch is already at the target (never removes the intended home).
- **Main checkout**: explicitly refuses to remove the primary checkout (first porcelain `worktree` entry). Confirmed `main` stays intact; and because the post-reclaim verify loop then sees the branch still held at a non-target (main), it returns 1 — i.e. it fails *loudly* through the fatal logger rather than ever touching main. Correct and conservative.
- **Lock override** is bounded to same-branch-non-target-non-main, which is exactly the migration-orphan scenario.
- **Uncommitted / unpushed work**: warns loudly but proceeds (rc=0). Confirmed both the UNCOMMITTED and UNPUSHED warnings fire and the worktree is still reclaimed — reasonable given these singleton branches reset to `origin/main` each cycle, and the trade-off is now visible in the log.

### No more silent death
`log_worktree_fatal` writes a `FATAL` line to **both stderr and the daemon log** then `exit 1`. The final (previously un-redirected) `git worktree add` is now routed through it in every launcher, so a dead launcher always leaves a trace.

### Wiring
All six singleton launchers (deploy, auditor, mechanic, aristotle, peer-reviewer, seeker) reclaim before add and route the final add-failure through the fatal logger. Confirmed each sources `worktree-cleanup.sh` and defines `LOG_FILE` before `create_worktree`; `launch-seeker.sh` correctly adds the missing `source`. The reclaim integrates cleanly with each launcher's existing tier fallback (reclaim frees the checkout, then `branch -D`/tier-2 reuse handles the ref).

### Scope & hygiene
Does not touch `scripts/lean/launch.sh` or any `*token-registry*` (concurrent work), per issue constraints. `bash -n` and `shellcheck -S warning` clean on all changed lines (the lone SC2034 `ROLE_FILE` in aristotle is pre-existing and untouched). Merge state CLEAN.

Nice work — the design notes in the helper header and per-launcher comments make the rare lock-override trade-off auditable. Approving.

Issue #41048, opened 2026-07-22
Daemon state file under .loom/ is deleted by Loom reinstall and can't self-heal


rjwalters:
## Context
During a Loom reinstall, the installer's uninstall phase deleted `.loom/lean-daemon-state.json` (it lives under Loom's managed `.loom/` dir), which made `/lean status` report a bogus ~177-day (`4249h`) uptime and zeroed session stats. Worse, the running daemon **couldn't self-heal**: `update_daemon_state()` in `scripts/lean/launch.sh` only writes when the file already exists (`if [[ -f "$STATE_FILE" ]]`), so a deleted state file stays missing until a full daemon restart via `init_state`. I recreated it manually (correct `started_at` from the process/keeper start time).

## Suggested acceptance criteria
- [ ] EITHER relocate the lean daemon state out of `.loom/` (so Loom's installer doesn't treat it as Loom-owned and delete it) — e.g. `.lean/` or a lean-specific dir — OR
- [ ] Make the daemon **self-heal** a missing/corrupt `STATE_FILE`: recreate it (deriving `started_at` from the daemon process / keeper start time) instead of no-op'ing when absent.
- [ ] `/lean status` uptime/stats survive a Loom reinstall (or a manual state-file deletion) without operator intervention.

Issue #39651, opened 2026-07-20
Daemon 'PRs ready to merge' metric only counts loom:pr — masks the real math-PR backlog


rjwalters:
## Summary
`/lean status` reports **"PRs ready to merge: N"** computed (in `scripts/lean/launch.sh`, ~line 1822) as:
```bash
ready_prs=$(timeout 10 gh pr list --label "loom:pr" --json number 2>/dev/null | jq 'length')
```
But per `CLAUDE.md`, **math agents (Researcher, Enricher, Aristotle, Auditor) must NOT add `loom:review-requested`/`loom:pr`** — they label PRs `enrichment` / `research` / `loom:auditor` and the deployer merges them directly.

## Impact
During the 7-day deployer outage (#39649) there were **537 open, mostly MERGEABLE math PRs**, yet this line read **0** the entire time. The metric that should have screamed "huge backlog" instead said everything was fine, so the outage went unnoticed on every status check.

## Proposed fix
Count the PRs the deployer actually merges. Options:
- Count open PRs carrying any math label (`enrichment`, `research`, `loom:auditor`, `aristotle-integration`), or
- Count open PRs that are `MERGEABLE` and not `loom:review-requested`, or
- Report both a "loom:pr (judge-approved)" count and a "math PRs mergeable" count as separate lines.

Whatever is chosen, the number must reflect the deployer's real queue so a stalled deployer is visible from `/lean status`.

PR #41506, opened 2026-07-22
feat(lean-daemon): detect & alert when a configured agent has no running session

### LINKED ISSUE #39652: Daemon should detect & alert when a configured agent has no running session (deployer went unnoticed 7 days)
## Summary
The lean daemon (`scripts/lean/launch.sh daemon`) had `deployer: 1` in its config, yet **no deployer tmux session existed for 7 days** and the daemon never surfaced the discrepancy. The launcher was dying silently (see #39649), but the deeper gap is that **the daemon does not verify that each configured agent actually has a live session**, nor alert when one is persistently missing.

## Observed
- `.loom/lean-daemon-state.json` → `config.deployer = 1`
- `tmux has-session -t deployer` → false, for ~7 days
- No warning in the daemon log, no health flag; `/lean health` simply omitted the row.

## Proposed fix
- Each daemon cycle: for every agent with `config.<type> >= 1`, assert a live session/PID exists. If a configured agent has been absent for ≥ N cycles, emit a `WARN`/`ERROR` daemon-log line and reflect it in `/lean status` / `/lean health` (e.g. a red `MISSING (configured:1 running:0)` row) instead of silently omitting it.
- Bonus: after a respawn attempt for a polling agent, confirm the session came up; if not, count it as a failure the keeper/daemon can escalate rather than a silent no-op.

## Impact
Observability gap that turned a silent launcher bug (#39649) into a multi-day undetected outage. Cheap to add; high diagnostic value.

rjwalters:
Closes #39652

## Problem
The lean daemon had `deployer: 1` in its config yet **no deployer tmux session existed for ~7 days**, and nothing surfaced it: `/lean health` silently omitted the row (`get_all_agent_sessions` only enumerates *live* sessions) and the daemon log stayed quiet. This is the observability companion to #39649 (which addresses *why* the launcher died) — this makes the **absence visible**.

## What changed (`scripts/lean/launch.sh`)
**Detection (daemon cycle).** After the existing pool-gap respawn pass, each cycle now compares every configured agent (`config.<type> >= 1`) against its live tmux session count and tracks how many **consecutive** cycles it has stayed below target:
- New `get_missing_cycles` / `set_missing_cycles` counters (bash-4 assoc array with a bash-3 `/tmp` fallback, mirroring the existing `LAST_RESPAWN_TIME` pattern).
- Escalates from the routine `pool gap` INFO to a **`MISSING` WARN** only after `MISSING_SESSION_ALERT_CYCLES` (=3) consecutive absent cycles, so a single mid-respawn / cooldown cycle never false-alarms. Aristotle scale-to-zero (#22471) is treated as an intentional absence.
- Persists the missing set to `STATE_FILE` (`.missing_agents`) via `write_missing_agents` so it survives between cycles and after the daemon exits.

**Surfacing.**
- `/lean health`: new `print_missing_agent_rows` renders a red `MISSING (configured:1 running:0)` row for every configured-but-absent agent — including when the *whole* pool is dead (previously an early return hid it) — plus a summary count and an advisory pointing at #39649 and the daemon log.
- `/lean status` (`status.sh`): reads `.missing_agents` and prints a red MISSING block; JSON output gains a top-level `missing_agents` array.

Reuses the daemon's existing per-agent respawn machinery; this adds **detection + logging + a status row only** — no new auto-remediation (that's #39649 / launcher-reclaim territory). Coordinates around, and does not touch, the `get_work_queue_stats` `ready_prs` metric (#39651) or the `recreate_daemon_state` / `update_daemon_state` self-heal (#41048).

## Testing
- `bash -n` passes on both scripts; `shellcheck -S error` clean.
- New `scripts/tests/daemon-missing-agent.test.sh` — 20 assertions, all passing. Stubs `tmux` and verifies:
  - `count_agent_sessions` per agent family,
  - the red `MISSING` health row appears for a configured/absent deployer (and when the whole pool is dead),
  - **no false alarm** for a normally-running agent,
  - `STATE_FILE` persistence of `.missing_agents`,
  - a WARN fires **only** after 3 consecutive absent cycles and the counter **resets** on recovery.
- A source guard around `main "$@"` lets the test source `launch.sh` in isolation (consistent with the repo's other sourced-and-tested scripts).

### DIFF
--- a/scripts/lean/launch.sh
+++ b/scripts/lean/launch.sh
@@ -74,6 +74,14 @@ STUCK_CPU_THRESHOLD="0.5"
 DEFAULT_DAEMON_INTERVAL=60
 RESPAWN_COOLDOWN_SECONDS=300  # 5 minutes between respawns of same agent
 
+# Persistent-missing-session alerting (#39652). A configured agent whose session
+# is absent for a single cycle is routine (mid-respawn, cooldown). Only after
+# this many CONSECUTIVE cycles below target does the daemon escalate from a
+# routine "pool gap" INFO to a MISSING WARN and surface a red row in
+# `/lean health` / `/lean status`. Kept small so a genuine multi-cycle outage
+# (e.g. the deployer that went unnoticed for 7 days) is caught within minutes.
+MISSING_SESSION_ALERT_CYCLES=3
+
 # Default pool sizes
 # Balanced team: 5 researchers (continuous), 8 support agents (sleeping between cycles)
 # ~72 active min/hr across 9 accounts = ~8 min/hr per account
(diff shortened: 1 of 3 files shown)


rjwalters:
## Judge Review — APPROVED

Reviewed as the observability companion to #39649. Verified against current `origin/main` (contains #39651 `ready_prs`, #41048 state self-heal, #39649 launcher-reclaim).

### Verification
- **Detection & threshold** — the 2c daemon block tracks CONSECUTIVE absent cycles per configured agent (`config.<type> >= 1`) and escalates `pool gap` INFO → `MISSING` WARN only at `MISSING_SESSION_ALERT_CYCLES=3`. No off-by-one: counter goes 1,2,3 → fires on the 3rd absent cycle; resets to 0 on recovery (else-branch). Confirmed by the test replica.
- **Visibility** — `print_missing_agent_rows` renders the red `MISSING (configured:N running:M)` row in `/lean health`, and the old empty-sessions early-return now calls it first so a dead pool is no longer hidden. `status.sh` reads `.missing_agents` for both the red block and a new top-level `missing_agents` JSON array.
- **No false alarm** — a running agent yields no row; Aristotle scale-to-zero (`ARISTOTLE_SCALED_MARKER` / `is_aristotle_scaled_to_zero`, #22471) is treated as intentional in both the daemon loop and the health renderer.
- **State coordination (#41048)** — `write_missing_agents` is guarded by `jq empty` and writes atomically (mktemp+mv), touching only `.missing_agents`. `update_daemon_state` (runs after, line 2686) sets only `.daemon_cycles/.daemon_respawns/.last_daemon_cycle` and preserves `.missing_agents`; `recreate_daemon_state` only fires on missing/corrupt file and self-heals within the same cycle. No clobber, no fight with the corrupt-file self-heal.
- **Scope** — detection + logging + status-row only; does not touch `get_work_queue_stats` `ready_prs` (#39651) or add auto-remediation (#39649 territory).
- **bash 3.2 / correctness** — `bash -n` clean on all three files; `shellcheck -S error` clean; source guard `[[ "${BASH_SOURCE[0]}" == "${0}" ]]` verified (direct exec still runs `main`, sourcing skips it). The full 20-assertion suite passes on the host's bash 3.2.57, which exercises the `/tmp` fallback path (assoc-array unavailable) — proving the bash-3 compat branch, not just bash-4.

### Non-blocking observation (suggested follow-up, not required)
The daemon's `cmd_daemon` loop early-`continue`s when `get_all_agent_sessions` is empty (total blackout), *before* the 2c block — so in a whole-pool-dead case the per-agent `MISSING` WARN and `.missing_agents` persistence do not fire, and `/lean status` (which reads `.missing_agents`) would stay empty. This is not a regression and not silent: the pre-existing `WARN "No agent sessions found"` still logs every cycle, and `/lean health` surfaces the per-agent MISSING rows live (covered by the "whole pool dead" test). The originally-reported bug (one agent missing among a live pool) is fully handled. A future tweak could run 2c even in the empty-sessions branch so `/lean status` reflects total blackout too.

CI: no GitHub checks configured for this shell path; validated manually (bash -n + shellcheck + full test suite green). Merge state CLEAN / MERGEABLE.

Approving.

PR #41546, opened 2026-07-22
fix(claude-wrapper): atomic, non-destructive token bootstrap (fix .loom/tokens race)

### LINKED ISSUE #41542: claude-wrapper token bootstrap wipes + non-atomically rewrites .loom/tokens (race); port loom's atomic pattern
## Context
Observed while verifying claude-monitor account auto-pickup: the shared `.loom/tokens` dir was briefly caught mid-rewrite (transient broken/self-referencing symlink + empty `*.token` glob) while concurrent agent spawns re-bootstrapped it. It self-heals to a valid pool and the fleet keeps working, but it's a genuine race on a shared directory.

## Root cause
`scripts/agents/claude-wrapper.sh` bootstrap does a **destructive, non-atomic** rebuild:
```bash
rm -f "$_tokens_dir"            # line ~169 — wipes the whole tokens dir
mkdir -p "$_tokens_dir"         # line ~170
...
printf '%s' "$key" > "$_tokens_dir/$file"   # line ~200 — direct, non-atomic per-file write
```
During that window the dir is empty/absent, so a concurrent agent's bootstrap gate (`! ls "$_tokens_dir"/*.token`) fires and it ALSO rebuilds → racing rebuilds, and a selector reading mid-window sees no tokens.

## Loom already solved this (do NOT file upstream — verified loom is hardened)
`loom-tools/src/loom_tools/tokens/bootstrap.py::bootstrap_tokens` does it correctly:
- `tokens_dir.mkdir(parents=True, exist_ok=True)` — **updates in place, never wipes the dir**
- per-token: write to `<file>.tmp` then `os.replace(tmp, token_path)` — **atomic rename**
- `index.json` written via temp + rename too

Our local bash wrapper is the only unhardened materializer. Port loom's pattern.

## Suggested fix (in scripts/agents/claude-wrapper.sh bootstrap block)
1. Do **not** `rm -f "$_tokens_dir"` — use `mkdir -p "$_tokens_dir"` (idempotent) and update in place. (Keep the existing broken-symlink self-heal: if `$_tokens_dir` exists but is not a real dir, replace just that, carefully.)
2. Write each token atomically: `printf '%s' "$key" > "$dir/$file.tmp" && chmod 600 "$dir/$file.tmp" && mv -f "$dir/$file.tmp" "$dir/$file"` (mv on same filesystem is an atomic rename).
3. **Prune, don't wipe:** after writing the merged set, remove only `*.token` files whose stem is NOT in the current merged account set (so retired accounts still get cleaned, but the dir is never transiently empty).
4. Keep the selection path (Strategy 0 / ranking.json) untouched.

## Acceptance criteria
- [ ] Bootstrap never leaves `.loom/tokens` empty/absent mid-rewrite (no destructive `rm` of the dir); concurrent bootstraps converge without a broken window.
- [ ] Each `.token` is written via temp-file + atomic rename.
- [ ] Retired accounts are pruned (files not in the merged set removed), new accounts added — same net result as today, without the race.
- [ ] bash 3.2 compatible; existing behavior (claude-monitor > repo > home merge, email-derived filenames) unchanged.
- [ ] A concurrent-spawn test (or reasoned argument) shows no empty-window race.

rjwalters:
## Summary

Closes #41542.

`scripts/agents/claude-wrapper.sh`'s token-bootstrap block rebuilt the shared `.loom/tokens` pool **destructively and non-atomically**:

```bash
rm -f "$_tokens_dir"                               # wiped the whole shared dir
...
printf '%s' "$key" > "$_tokens_dir/$file"          # direct non-atomic per-file write
```

During that window `.loom/tokens` is empty/absent, so a **concurrent** agent spawn's bootstrap gate (`! ls "$_tokens_dir"/*.token`) fires and it *also* rebuilds → racing rebuilds; a selector reading mid-window sees no tokens. (Observed live: transient broken symlink + empty glob on the shared dir.)

## Fix — port loom's atomic pattern

Mirrors `loom_tools/tokens/bootstrap.py::bootstrap_tokens` (already hardened). Only the materialization block changes — the selection path (Strategy 0 / `ranking.json`) and merge/precedence logic (monitor > repo > home, email-derived filenames) are untouched.

1. **Update in place, no dir wipe.** `mkdir -p` (idempotent). The broken-symlink self-heal is kept but **scoped**: `$_tokens_dir` is replaced only when it exists and is *not* a real directory. A valid, populated dir is never blown away.
2. **Atomic per-token write:** `printf … > "$dst.tmp" && chmod 600 … && mv -f "$dst.tmp" "$dst"` — `mv` within one filesystem is an atomic rename (mirrors `os.replace`). A concurrent reader sees old or new content, never a partial write.
3. **Prune, don't wipe:** after writing the merged set, remove only `*.token` files whose stem is *not* in the merged set (retired accounts still cleaned), so the dir is never transiently empty. Dotfiles (`.ranking`/`.allowlist`/`.bad_tokens`) are preserved (they don't match `*.token`).

Net behavior is identical: same final token set, retired accounts pruned, new ones added — without the destructive-wipe race window.

## Testing

`bash -n` passes (macOS `/bin/bash` 3.2.57). Verified against a **stub** tokens dir + stub `~/.claude-monitor/accounts.env` via a `REPO_ROOT` override — the LIVE `.loom/tokens` was never touched. Two harnesses extract the *real* bootstrap code from the script:

- **(a) Idempotent, never empty:** first run materializes 2 accounts atomically; a second run leaves the dir populated (gate skips), same set.
- **(b) Retired account pruned:** a stale `carol-example.token` not in the merged set is removed.
- **(c) New account added:** `dave@example.com` appears; `alice` key rotated in place; `bob` preserved; `.ranking` dotfile kept.
- **(d) No empty-window race — demonstrated, not just argued:** 8 concurrent materializations while a watcher polled the dir 2000× observed **0** empty-`*.token` snapshots and **0** leftover `.tmp` files. Structurally: the merged-set files always exist on disk (atomic rename never unlinks the target; prune only touches non-merged stems *after* the writes), so a concurrent bootstrapper/selector always sees a populated dir.

## Constraints honored

- bash 3.2 compatible (no new associative arrays; existing parallel-array merge unchanged).
- Only `scripts/agents/claude-wrapper.sh` changed; selection + merge/precedence logic unchanged.

### DIFF
--- a/scripts/agents/claude-wrapper.sh
+++ b/scripts/agents/claude-wrapper.sh
@@ -177,7 +177,7 @@ if [[ -z "${CLAUDE_CODE_OAUTH_TOKEN:-}" ]]; then
             # Self-heal, scoped: only replace $_tokens_dir if it EXISTS but is NOT
             # a real directory (e.g. a broken/self-referencing symlink). Never
             # blow away a valid, populated directory.
-            if [[ -e "$_tokens_dir" && ! -d "$_tokens_dir" ]]; then
+            if [[ -L "$_tokens_dir" || ( -e "$_tokens_dir" && ! -d "$_tokens_dir" ) ]]; then
                 rm -f "$_tokens_dir" 2>/dev/null || true
             fi
             mkdir -p "$_tokens_dir" 2>/dev/null || true
--- a/scripts/agents/claude-wrapper.sh
+++ b/scripts/agents/claude-wrapper.sh
@@ -165,8 +165,21 @@ if [[ -z "${CLAUDE_CODE_OAUTH_TOKEN:-}" ]]; then
         _repo_tsv=$(_emit_accounts "$_repo_accts" repo)
         _mon_tsv=$(_emit_accounts "$_monitor_accts" monitor)
         if [[ -n "$_home_tsv" || -n "$_repo_tsv" || -n "$_mon_tsv" ]]; then
-            # Replace whatever is at $_tokens_dir (broken symlink, etc) with a real dir.
-            rm -f "$_tokens_dir" 2>/dev/null || true
+            # Update the pool IN PLACE — never destructively wipe the dir. A
+            # concurrent agent spawn shares this directory: if we `rm -rf` it and
+            # rebuild, its bootstrap gate (`! ls *.token`) fires on the empty
+            # window and it races us, and a selector reading mid-window sees no
+            # tokens (observed live: transient broken symlink + empty glob).
+            # Loom's bootstrap_tokens (loom_tools/tokens/bootstrap.py) solved this:
+            # mkdir(exist_ok=True) + per-token os.replace() atomic rename +
+            # prune-not-wipe. Port that pattern here (loom 0.12.0 #3699/#3700).
+            #
+            # Self-heal, scoped: only replace $_tokens_dir if it EXISTS but is NOT
+            # a real directory (e.g. a broken/self-referencing symlink). Never
+            # blow away a valid, populated directory.
+            if [[ -e "$_tokens_dir" && ! -d "$_tokens_dir" ]]; then
+                rm -f "$_tokens_dir" 2>/dev/null || true
+            fi
             mkdir -p "$_tokens_dir" 2>/dev/null || true
             # bash 3.2 has no associative arrays: parallel arrays + linear email
             # match (account counts are tiny). Applied LOW→HIGH precedence: home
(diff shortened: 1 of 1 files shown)


rjwalters:
## Loom Judge — Changes Requested

The core of this fix is correct and well-executed. The destructive `rm -f "$_tokens_dir"` is gone, per-token writes are atomic (`> "$dst.tmp"` → `chmod 600` → `mv -f`, with `.tmp` cleaned on failure), pruning removes only `*.token` stems absent from the merged set (dotfiles and `.tmp` never match the glob), and the selection/merge/precedence path is untouched. `bash -n` is clean under 3.2.57, merge state is CLEAN, and the never-empty-window concurrency argument holds for the normal populated-dir path. Points 2–5 all verify.

One blocking regression in the scoped self-heal (point 1):

```bash
if [[ -e "$_tokens_dir" && ! -d "$_tokens_dir" ]]; then
    rm -f "$_tokens_dir" 2>/dev/null || true
fi
mkdir -p "$_tokens_dir" 2>/dev/null || true
```

`-e` **dereferences** symlinks, so it returns *false* for a dangling or self-referencing symlink — which is exactly the "transient broken / self-referencing symlink" state the issue reports observing live. Verified under `/bin/bash` 3.2.57:

- dangling symlink (`tokens -> /nonexistent`): `[[ -e X && ! -d X ]]` is **false** → self-heal skipped → `mkdir -p` fails (`mkdir: tokens: No such file or directory`, exit 1) → every `> "$dst.tmp"` write fails → **bootstrap silently no-ops and the broken symlink is never healed**.
- self-referencing symlink (`tokens -> tokens`): same — `-e` false (ELOOP), never healed.

The gate `[[ ! -d "$_tokens_dir" ]]` *does* let these states enter the bootstrap block, so this path is reachable. The old unconditional `rm -f` healed both (rm operates on the link itself), so this is a functional regression versus `main` on the precise failure mode the issue was filed to fix. It also contradicts the block's own comment ("only replace `$_tokens_dir` if it EXISTS but is NOT a real directory (e.g. a broken/self-referencing symlink)") and the issue's explicit request to "keep the existing broken-symlink self-heal."

**Fix (one line):** detect symlinks explicitly with `-L`:

```bash
if [[ -L "$_tokens_dir" || ( -e "$_tokens_dir" && ! -d "$_tokens_dir" ) ]]; then
    rm -f "$_tokens_dir" 2>/dev/null || true
fi
```

Verified under bash 3.2.57 that this heals both dangling and self-referencing symlinks (`mkdir -p` then succeeds, exit 0), while a real populated directory still takes the skip path and keeps its `*.token` files (bootstrap doesn't even enter for a symlink-to-populated-dir, since the gate resolves through the link).

Everything else is good to merge once the self-heal covers broken symlinks again. Nice port of loom's `bootstrap_tokens` pattern otherwise.

rjwalters:
Addressed the Judge's change request: the self-heal guard for the `.loom/tokens` symlink now uses `-L` (which does NOT dereference) in addition to `-e && ! -d`, so it fires for a dangling or self-referencing symlink and restores the heal that was lost vs main. A real populated directory is neither `-L` nor `! -d`, so it still takes the in-place `mkdir -p` path and is never wiped; empty/missing dirs bootstrap as before.

Verified: `bash -n` passes, and a standalone test of the guard confirms HEAL for dangling + self-ref symlinks (and regular files), keep for real populated dir + empty dir + missing path. Scope kept to this one condition; atomic write / prune-don't-wipe / merge logic untouched.

rjwalters:
## Loom Judge: APPROVED

Re-review of the Doctor pass (commit `4d7f5db`) addressing my prior change request. The one blocking issue is fixed correctly and nothing else regressed.

### Verified fix
The self-heal guard now reads:
```sh
if [[ -L "$_tokens_dir" || ( -e "$_tokens_dir" && ! -d "$_tokens_dir" ) ]]; then
    rm -f "$_tokens_dir" 2>/dev/null || true
fi
```
The leading `-L` test does **not** dereference the link, so it catches a dangling or self-referencing symlink at `.loom/tokens` — exactly the gap that made the previous `-e`-only guard silently no-op. Behavior across all cases:

- **Dangling / self-ref symlink** → `-L` true → rm + `mkdir -p` → re-materialized (healed). Correct.
- **Real populated dir** → `-L` false, `-e && ! -d` false (it *is* a dir) → rm skipped, `mkdir -p` is a no-op → contents preserved, never wiped. Correct.
- **Empty / missing dir** → guard false → `mkdir -p` bootstraps, tokens written. Correct.
- **Stray non-dir regular file** → `-e && ! -d` true → replaced with a real dir. Correct bonus.

### Scope
Only the guard condition changed on this commit. The atomic per-token write (temp + `mv -f` rename), the prune-don't-wipe `*.token` loop (dotfiles preserved via glob), and the account selection/merge logic are all unchanged and were already verified sound in my prior pass. bash 3.2 compatible (no assoc arrays, glob-guarded loops).

### Checks
- `bash -n` on the head file: clean.
- Merge state: CLEAN.
- No CI checks configured for this shell-only path (expected).

Ships. Moving to `loom:pr`.

PR #41553, opened 2026-07-22
fix(seeker): guard tokens symlink against self-reference in main checkout

### LINKED ISSUE #41551: seeker launcher creates self-referencing .loom/tokens symlink in main checkout (unguarded ln -sfn)
## Root cause
`scripts/research/launch-seeker.sh:299` symlinks the agent worktree's tokens dir to the main pool with a **bare, unguarded** `ln`:
```bash
if [[ -d "$REPO_ROOT/.loom/tokens" ]]; then
    mkdir -p "$WORKTREE_PATH/.loom" 2>/dev/null || true
    ln -sfn "$REPO_ROOT/.loom/tokens" "$WORKTREE_PATH/.loom/tokens"
fi
```
The seeker's live pool/DB state lives in the **MAIN checkout** (REPO_ROOT), not a separate worktree, so `WORKTREE_PATH` resolves to `$REPO_ROOT`. That makes `dst == src`, so this creates a **self-referencing symlink**: `.loom/tokens -> /…/.loom/tokens` (points at itself). It's circular/broken — `ls .loom/tokens/*.token` returns nothing, `cd` fails with ELOOP.

## Symptom (observed live)
`.loom/tokens` in the main checkout flaps between a valid real dir (6 tokens) and a self-referencing symlink, mtime updating each seeker cycle. Agents still connect because each wrapper bootstrap heals it back to a real dir at spawn — but it's wasteful churn and was the source of the "token race" investigated in #41542. (The #41542 fix makes the wrapper's self-heal catch `-L` symlinks, so the symptom now auto-corrects; this issue removes the SOURCE.)

## Why seeker specifically
`deploy/launch-agent.sh`, `auditor/launch-agent.sh`, and `mechanic/launch-agent.sh` all have the self-pointing guard (compare `pwd -P` of src vs dst, refuse if equal — see the "Refusing to create self-pointing tokens symlink" block). `aristotle`/`peer-reviewer` don't create the symlink at all. **Only seeker does an unguarded `ln`.**

## Fix
Add the same self-pointing guard the sibling launchers use to `scripts/research/launch-seeker.sh:~296-301`: resolve `src_real="$(cd "$src" && pwd -P)"` and `dst_real` and **skip the `ln` when they're equal** (dst == src → self-reference). Also skip cleanly when `$src` is already a broken symlink (`cd` fails → don't create a bad link). Prefer factoring the guarded-symlink logic into a shared helper (e.g. `scripts/lib/`) that all launchers source, so the three copies + seeker converge — but a local guard matching the siblings is acceptable if a shared helper is out of scope.

## Acceptance criteria
- [ ] Seeker never creates a self-referencing `.loom/tokens` symlink when `WORKTREE_PATH` == `REPO_ROOT` (skips the ln, main's real tokens dir is left intact).
- [ ] Worktree-based agents (dst != src) still get the correct symlink to the main pool.
- [ ] Skips cleanly if `$REPO_ROOT/.loom/tokens` is already a broken symlink.
- [ ] `bash -n` clean.

rjwalters:
## Summary
`scripts/research/launch-seeker.sh` symlinked the worktree tokens dir to the main pool with a **bare, unguarded** `ln -sfn`. Because the seeker operates from the **MAIN checkout**, `WORKTREE_PATH` resolves to `$REPO_ROOT`, making `dst == src` and producing a **self-referencing symlink** (`.loom/tokens -> .../.loom/tokens`). This flapped the main token pool between a real dir and a broken circular symlink every seeker cycle.

## Fix
Apply the same self-pointing guard the `deploy`/`auditor`/`mechanic` launchers already use (approach: **local inline guard**, matching the siblings — no shared helper, to keep the change focused and avoid touching the other launchers):

- Resolve `src_real` (physical path of `$REPO_ROOT/.loom/tokens`) and `dst_real` (physical path of the intended `$WORKTREE_PATH/.loom/tokens`).
- If `src_real == dst_real` (self-reference) → skip the `ln`, log a warning, leave main's real tokens dir intact.
- If `src_real` is empty (src not a resolvable directory / broken symlink) → skip cleanly.
- Otherwise (genuine distinct worktree) → create the symlink to the main pool as before.

## Testing
`bash -n scripts/research/launch-seeker.sh` passes. Guard logic exercised against a scratch REPO_ROOT/WORKTREE_PATH (live `.loom/tokens` untouched):

- (a) `WORKTREE_PATH == REPO_ROOT` → guard skips the `ln`; main `.loom/tokens` stays a real dir, existing token file intact. **PASS**
- (b) genuine distinct `WORKTREE_PATH` → symlink IS created, resolves to the main pool token. **PASS**
- (c) `src` is a broken symlink → skips cleanly, no bad link created. **PASS**

Closes #41551

### DIFF
--- a/scripts/research/launch-seeker.sh
+++ b/scripts/research/launch-seeker.sh
@@ -293,10 +293,27 @@ launch_agent() {
     # Create or update worktree
     create_worktree
 
-    # Symlink OAuth tokens so the claude-wrapper can find them in the worktree
+    # Symlink OAuth tokens so the claude-wrapper can find them in the worktree.
+    # Guard against self-pointing symlinks: the seeker operates from the MAIN
+    # checkout, so WORKTREE_PATH == REPO_ROOT and dst == src. A bare `ln -sfn`
+    # would then replace main's real tokens dir with a circular self-reference
+    # (.loom/tokens -> .../.loom/tokens), flapping the pool every cycle (#41551).
+    # Mirror the guard used by deploy/auditor/mechanic launchers.
     if [[ -d "$REPO_ROOT/.loom/tokens" ]]; then
-        mkdir -p "$WORKTREE_PATH/.loom" 2>/dev/null || true
-        ln -sfn "$REPO_ROOT/.loom/tokens" "$WORKTREE_PATH/.loom/tokens"
+        local src="$REPO_ROOT/.loom/tokens"
+        local dst="$WORKTREE_PATH/.loom/tokens"
+        local src_real dst_real
+        src_real="$(cd "$src" 2>/dev/null && pwd -P)"
+        dst_real="$(dirname "$dst")"; dst_real="$(cd "$dst_real" 2>/dev/null && pwd -P)/$(basename "$dst")"
+        if [[ -z "$src_real" ]]; then
+            print_warning "Skipping tokens symlink: $src is not a resolvable directory (broken symlink?)."
+        elif [[ "$src_real" == "$dst_real" ]]; then
+            print_warning "Refusing to create self-pointing tokens symlink ($dst → $src). Skipping."
+        else
+            mkdir -p "$WORKTREE_PATH/.loom" 2>/dev/null || true
+            ln -sfn "$src_real" "$dst"
+            print_info "Linked .loom/tokens for OAuth token rotation"
+        fi
     fi
 
     # Check pool depth first


rjwalters:
## Judge Review: APPROVED

Verified the fix against issue #41551 and the sibling launcher pattern.

**Guard correctness**
- Self-pointing guard resolves `src_real` and `dst_real` via `pwd -P` (physical paths, so it catches symlinked equivalence, not raw-string equality) and skips the `ln` when equal — exactly the seeker's main-checkout case where `WORKTREE_PATH == REPO_ROOT`.
- `dst_real` is computed the correct way: `dirname "$dst"` → `cd … pwd -P` → append `basename`. This resolves even when `$dst` (the tokens dir) does not exist yet, avoiding the `cd "$dst"`-fails-→-empty-→-guard-never-matches bug. Matches `scripts/deploy/launch-agent.sh:166-168` byte-for-byte in approach.
- Broken/self-ref `src`: `cd "$src"` fails → `src_real` empty → the `-z "$src_real"` branch skips cleanly with a warning, so no bad link is created. (This empty-src branch is an addition over the deploy sibling, and is exactly what acceptance criterion 3 asked for.)
- Genuine distinct worktree (`dst_real != src_real`, non-empty) falls through to the `else`: `mkdir -p` + `ln -sfn "$src_real" "$dst"`. Real worktree agents are not over-blocked.

**Structure / scope**
- `if / elif / else` guarantees the skip actually skips — `ln` lives only in the `else`, so the warning paths never fall through to create a link.
- Only `scripts/research/launch-seeker.sh` changed (+20/-3). No touch to `claude-wrapper.sh` (#41542) or the sibling launchers' guards.
- `bash -n` clean.

All three builder test cases (self-ref skip, distinct-worktree link, broken-src skip) are supported by the diff. Merge state CLEAN; no CI checks configured for this bash-only path.

Solid, focused fix that mirrors the established sibling pattern. Approving.

Issue #41047, opened 2026-07-22
Daemon session-stats counters stuck — Deployments/enriched don't increment despite real work


rjwalters:
## Context
Across a long session, `/lean status` showed `Deployments: 23`, `Entries enriched`, `Research completed`, etc. staying flat (or resetting) even though real deploys/enrichment/research were confirmed happening (merged PRs, live deploy URLs). The `session_stats` counters in `.loom/lean-daemon-state.json` (updated in `scripts/lean/launch.sh` via `session_stats.deployments += $deploys`, etc.) are not tracking real activity.

## Suggested acceptance criteria
- [ ] Identify why `session_stats` increments don't land (completion-signal processing path in `launch.sh` — `process_completion_signals`/`update` of `session_stats`).
- [ ] Counters reflect real deploy/enrich/research/submit activity over time.
- [ ] Bogus resets (e.g. after a state-file recreation) don't zero prior tallies unexpectedly.

PR #41601, opened 2026-07-22
fix(daemon): run missing-agent detection during total-pool blackout (#41509)

### LINKED ISSUE #41509: Daemon: whole-pool blackout early-continues before missing-agent detection (edge case from #39652)
## Context
Follow-up to #39652 (PR #41506), flagged by the Judge as non-blocking.

The daemon's `cmd_daemon` loop early-`continue`s when `get_all_agent_sessions` is empty (a **total blackout** — no agents at all) *before* the new missing-agent detection block (step 2c). So in a whole-pool-dead scenario, the per-agent `MISSING` WARN and the `.missing_agents` STATE_FILE persistence don't fire, and `/lean status` (which reads `.missing_agents`) would stay empty.

## Not a regression
- The pre-existing `WARN "No agent sessions found"` still logs every cycle in that case.
- `/lean health` surfaces the MISSING rows live (verified).
- The originally-reported bug — **one** agent missing among a live pool — is fully handled by #39652.

So the gap is only: a *total* blackout doesn't populate the persisted `.missing_agents` for `/lean status`.

## Suggested fix
Run the missing-agent detection (step 2c) even when `get_all_agent_sessions` is empty — i.e. don't early-`continue` before it, or move the detection ahead of the empty-pool early-return so `.missing_agents` is persisted for `/lean status` in the blackout case too.

## Acceptance
- [ ] A total pool blackout persists `.missing_agents` and shows the MISSING rows in `/lean status` (not just `/lean health`).
- [ ] Single-agent-missing behavior (from #39652) unchanged; no false alarm for a healthy pool.

rjwalters:
## Summary

Follow-up to #39652 (PR #41506). The daemon's `cmd_daemon` loop early-`continue`d when `get_all_agent_sessions` returned empty (a **total blackout** — zero agent sessions) *before* reaching the step-2c missing-agent detection block. So in a whole-pool-dead scenario the per-agent `MISSING` WARN and the `.missing_agents` STATE_FILE persistence never fired, and `/lean status` (which reads `.missing_agents`) stayed empty. `/lean health` still showed the rows live and the pre-existing `WARN "No agent sessions found"` still logged — the gap was only the persisted record for `/lean status`.

## Changes

- Extract step 2c into a reusable `detect_and_persist_missing_agents` helper that recomputes live counts via the existing `count_agent_sessions` (single source of truth) and reuses `get_missing_cycles` / `set_missing_cycles` / `write_missing_agents` / `MISSING_SESSION_ALERT_CYCLES`.
- Call the helper on the total-blackout early-`continue` path too, so `.missing_agents` is persisted and `/lean status` shows MISSING rows in a blackout.
- Hoist the per-cycle STATE_FILE config re-read + `apply_schedule` ahead of the health check so both the normal and blackout paths see the same scheduled targets — a scheduled scale-to-zero during a blackout won't false-alarm on stale config.

The early-`continue` is otherwise preserved: the health-check / pool-gap / respawn / completions / state self-heal logic is unchanged, and a single-agent-missing-among-a-live-pool (the original #39652 case) behaves exactly as before.

## Testing

`bash -n scripts/lean/launch.sh` passes. Extended `scripts/tests/daemon-missing-agent.test.sh`:
- Total blackout (deployer configured=1, zero sessions): nothing persisted at cycles 1–2, `.missing_agents` persists deployer at cycle 3 (threshold), MISSING surfaced.
- Healthy pool through the same helper: `.missing_agents` stays `[]`, counter stays 0 (no false alarm).

All 28 assertions pass.

Closes #41509

### DIFF
--- a/scripts/lean/launch.sh
+++ b/scripts/lean/launch.sh
@@ -1493,6 +1493,67 @@ write_missing_agents() {
     fi
 }
 
+# Helper: Persistent-missing-session detection & alerting (#39652, #41509).
+# Tracks how many CONSECUTIVE cycles each CONFIGURED agent has stayed below
+# target; once the count crosses MISSING_SESSION_ALERT_CYCLES it escalates to a
+# WARN and persists a MISSING record (via write_missing_agents) that
+# /lean health & /lean status render as a red row instead of silently omitting
+# the agent. Live counts come from count_agent_sessions so this is a single
+# source of truth and can run in BOTH the normal daemon cycle AND the total-pool
+# blackout path where get_all_agent_sessions is empty (#41509 -- otherwise the
+# blackout early-continues before detection and /lean status stays empty).
+# Args, in order:
+#   enricher researcher aristotle auditor seeker deployer herald mechanic tester
+detect_and_persist_missing_agents() {
+    local enricher="$1" researcher="$2" aristotle="$3" auditor="$4" \
+          seeker="$5" deployer="$6" herald="$7" mechanic="$8" tester="$9"
+    local missing_json="[]"
+    local mtype mtarget mactive mcycles
+    for _pair in \
+        "enricher $enricher" \
+        "researcher $researcher" \
+        "aristotle $aristotle" \
+        "auditor $auditor" \
+        "seeker $seeker" \
+        "deployer $deployer" \
+        "herald $herald" \
+        "mechanic $mechanic" \
+        "tester $tester"; do
+        read -r mtype mtarget <<< "$_pair"
+
+        # Only configured agents (target >= 1) can be "missing".
+        if [[ "$mtarget" -lt 1 ]]; then
+            set_missing_cycles "$mtype" 0
+            continue
+        fi
+
+        # Aristotle scale-to-zero (issue #22471) is an intentional absence,
+        # not a failed launch -- don't false-alarm on it.
+        if [[ "$mtype" == "aristotle" ]] && [[ -f "$ARISTOTLE_SCALED_MARKER" ]]; then
+            set_missing_cycles "$mtype" 0
+            continue
+        fi
+
+        mactive=$(count_agent_sessions "$mtype")
+        if [[ "$mactive" -lt "$mtarget" ]]; then
+            mcycles=$(get_missing_cycles "$mtype")
+            mcycles=$((mcycles + 1))
+            set_missing_cycles "$mtype" "$mcycles"
+            if [[ "$mcycles" -ge "$MISSING_SESSION_ALERT_CYCLES" ]]; then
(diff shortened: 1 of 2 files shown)


rjwalters:
## Judge review: APPROVE

Verified against the acceptance criteria for #41509.

**1. Blackout path now runs detection via a shared helper (single source of truth).**
`detect_and_persist_missing_agents` is extracted verbatim from the old inline step-2c block and is now called on both the total-blackout early-`continue` path and the normal cycle. It recomputes live counts via `count_agent_sessions` and reuses `get_missing_cycles`/`set_missing_cycles`/`write_missing_agents`/`MISSING_SESSION_ALERT_CYCLES`. I diffed `count_agent_sessions` against the inline `*_active` tmux loops (enricher/researcher × MAX, mechanic 1..3 + `mechanic-agent`, singletons for the rest) — they are session-for-session identical, so the helper is not a divergent copy. No double source of truth.

**2. Single-agent-missing (#39652) unchanged.** For a persistently-absent agent (respawn fails/cooldown), pre- and post-respawn counts are both below target, so the counter increments identically. The threshold-at-N and reset-on-recovery tests still pass. Note (non-blocking): the helper now reads counts *after* the pool-gap respawns rather than the old pre-respawn `*_active` snapshot — this only differs in the narrow case where a respawn successfully (re)creates a live session within the same cycle, where resetting the counter is the more correct behavior. It does not affect the genuinely-stuck-agent detection that #39652 targets.

**3. No false alarm + config hoist correct.** The per-cycle STATE_FILE config re-read + `apply_schedule` is a pure move to ahead of the health check (same 8 agents read — `tester` omitted in both old and new, pre-existing). Ordering is preserved (read-then-apply), there is exactly one `apply_schedule` per loop iteration (plus the pre-loop init at setup), respawn still consumes the same function-scoped vars, and nothing between the old and new location depends on the vars. A scheduled scale-to-zero now can't false-alarm the blackout path on stale config. The healthy-pool test confirms `.missing_agents` stays `[]`.

**4. No disturbance to the rest of the cycle.** Respawn, #41047 completions routing, #41048 `recreate_daemon_state`/`update_daemon_state` self-heal, and #39651 `ready_prs` are all intact. `write_missing_agents` is atomic (mktemp + mv, guarded by STATE_FILE existence and `jq empty`) and is called *sequentially before* `update_daemon_state` in the blackout path — same process, no race with #41048's state write.

**5. Threshold/bash/tests.** Threshold logic unchanged (alert at N cycles, reset on recovery). `bash -n` clean on system bash 3.2.57; no bash-4 constructs. The extended test suite passes 28/28, including the new blackout-persists and healthy-no-false-alarm cases.

No CI checks are configured for shell-only changes on this branch (expected). Merge state CLEAN / MERGEABLE.

LGTM.

Issue #41727, opened 2026-07-22
Account selection always picks the single most-available account — collides with sibling Loom daemon on the shared claude-monitor pool → session-limit faults


rjwalters:
**Context:** observed live while a **Loom** agent pool started on the same machine as the running lean-genius daemon. Both systems source accounts from the **same claude-monitor pool**, and both deterministically pick the *single most-available* account. Within ~90s that shared account hit its **session limit** ("resets 4pm") with ~9 concurrent Claude sessions on one account, while other accounts sat idle.

## Root cause
`scripts/agents/claude-wrapper.sh` account selection picks the least-utilized account deterministically:

- **Strategy 0 (claude-monitor)** — lines ~322-346: reads `~/.claude-monitor/ranking.json`, then
  ```jq
  .accounts | map(select(.status=="available"))
            | sort_by((.utilization["7d"]//0),(.utilization["5h"]//0))
            | .[].email
  ```
  and iterates `_mon_emails`, `break`ing on the **first** eligible → the single most-available account.
- **Strategy 1 (cached ranking)** — lines ~365-384: iterates `.loom/tokens/.ranking`, `break`s on the **first** non-exhausted/blocked entry.

Strategies 2/3 are random, but the monitor/ranked paths (the ones that fire whenever claude-monitor data is fresh) always take `[0]`. **A sibling daemon reading the same ranking picks the same account** → both drive one account to its session limit.

## Impact
- lean-genius spawns many concurrent workers (enricher×N, researcher×N, seeker, deployer, …). When Loom (or a second lean workspace) also runs, they pile onto the same top account instead of spreading across the pool → premature session-limit faults despite spare aggregate capacity.

## Proposed fix
Randomize over the **top-N available accounts** (default N=3) rather than always the least-utilized:
- Strategy 0: after the `sort_by` jq, take the first N emails and pick one at random (`shuf -n1` / bash `$RANDOM` index) instead of `break`-on-first.
- Strategy 1: collect the leading run of non-exhausted ranking entries and random-pick among the first N.
- Make N configurable (env `LEAN_ACCOUNT_SPREAD_TOP_N`, default 3; N=1 = today's behavior). Preserve allowlist gating and bad-token/exhausted skipping.

## Cross-repo note
[...]

Issue #41033, opened 2026-07-21
Consume claude-monitor ranking.json in account selection (fallback: round-robin)


rjwalters:
## Motivation

lean-genius load-balances across multiple Claude accounts in `scripts/agents/claude-wrapper.sh`, selecting from the shared `~/.loom/accounts.env` / repo pool. A companion tool (**claude-monitor**) tracks richer per-account quota data (session/weekly utilization, per-model/Fable allowance, reset times) and can emit it as a non-secret export. Consuming that would give lean-genius smarter selection than its current logic.

Designed as a **soft, auto-detected dependency**: use the export if present, fall back to current behavior when absent.

## Scope

In the account-selection path (`scripts/agents/claude-wrapper.sh`):
1. If **`~/.claude-monitor/ranking.json`** exists and is fresh (gate on `generated_at`/mtime, ~10-min window), rank eligible accounts by `(status, util_7d, util_5h)` — least-utilized `available` account first — joining by **email**.
2. Else fall back to today's selection (round-robin over the repo/pool accounts).

Keep claude-monitor non-required — absent export → unchanged behavior.

### Contract: `~/.claude-monitor/ranking.json` (schema v1)

```json
{
  "schema": 1,
  "generated_at": "2026-01-01T00:00:00Z",
  "accounts": [
    {
      "email": "user@example.com",
      "plan": "max_20x",
      "status": "available",
      "utilization": { "5h": 0.12, "7d": 0.44 },
      "resets":      { "5h": "2026-01-01T02:00:00Z", "7d": "2026-01-04T00:00:00Z" },
      "models":      { "fable": { "utilization": 0.30 } },
      "updated_at":  "2026-01-01T00:00:00Z"
    }
  ]
}
```

`status` ∈ `available | rate_limited | exhausted | blocked`. No secrets in this file (keys stay in `accounts.env`, 0600).

## Coordination

- loom consumer (same contract): rjwalters/loom#3697
- claude-monitor producer: (see claude-monitor issue)

PR #41742, opened 2026-07-22
fix(claude-wrapper): classify 5h session-cap message as TOKEN_EXHAUSTED

### LINKED ISSUE #41737: Exhaustion regex misses 'You've hit your session limit · resets …' (5h session cap) — rotation machinery never fires for it; agent pins to the dead account
**Context:** observed live — an agent hit **"You've hit your session limit · resets 4pm (America/Los_Angeles)"** (the 5-hour rolling session cap) and did **not** rotate to another account, despite the wrapper's otherwise-working rotation machinery.

## Root cause
`scripts/agents/claude-wrapper.sh` line ~635 classifies exhaustion with:
```bash
if echo "$output" | grep -qi "hit your limit\|monthly usage limit\|weekly limit\|out of extra usage"; then
    echo "TOKEN_EXHAUSTED"; return
fi
```
The comment above it already recognizes the **multi-word-gap** hazard (it was fixed for the org "monthly usage limit" and "weekly limit" phrasings). But the **session cap** phrasing — `You've hit your session limit · resets …` — is not covered:
- `hit your limit` does **not** match `hit your `**`session `**`limit` (the word "session" sits in the gap — exactly the documented failure mode).
- It isn't "weekly"/"monthly usage"/"out of extra usage" either.

So `classify_error` never returns `TOKEN_EXHAUSTED` for this message → `mark_token_bad()` + `rotate_to_working_token()` (lines ~690/~726, wired into the retry loop at ~1047/~1063) **never fire** → the agent pins to the exhausted account and burns 300s ticks (or dies) instead of rotating to a healthy one.

## Fix
Add the session-cap phrasing to the exhaustion pattern:
```bash
grep -qi "hit your limit\|hit your session limit\|session limit\|monthly usage limit\|weekly limit\|out of extra usage"
```
(or a tolerant `hit your.*limit` that spans the gap). Add a regression fixture for `You've hit your session limit · resets 4pm`. Consider matching the generic `· resets ` suffix as a secondary signal.

## Why it matters
This is the single most common cap in normal operation (the 5-hour session window), so the miss defeats the rotation feature precisely when it's needed most — as seen when a sibling Loom daemon and lean-genius collided on the same claude-monitor account (see the companion top-N-spread issue #41727).

Related: #41727 (proactive top-N account spread). Cross-repo: rjwalters/loom is adding the whole rotate-on-exhaustion pattern (it currently lacks it) and will mirror this session-limit phrasing.

rjwalters:
## Problem

The exhaustion classifier in `scripts/agents/claude-wrapper.sh` `classify_error` (~line 635) missed the 5-hour rolling **session cap** message:

> `You've hit your session limit · resets 4pm (America/Los_Angeles)`

(note the `·` middot). The old pattern was:

```bash
grep -qi "hit your limit\|monthly usage limit\|weekly limit\|out of extra usage"
```

`hit your limit` does **not** match `hit your `**`session`**` limit` — the word "session" sits in the multi-word gap, the exact failure mode the comment above already documents for the org "monthly usage limit" / "weekly limit" phrasings. It's also not weekly/monthly-usage/out-of-extra-usage.

**Impact:** `classify_error` never returned `TOKEN_EXHAUSTED` for the session cap → `mark_token_bad()` + `rotate_to_working_token()` never fired → the agent pinned to the exhausted account and burned 300s ticks (or died) instead of rotating to a healthy account. The 5-hour session window is the most common cap in normal operation, so the miss defeated the rotation feature precisely when it's needed most.

## Fix

Add `hit your session limit` and the broader `session limit` alternatives to the `grep -qi` alternation, and document the session-cap phrasing in the comment block:

```bash
grep -qi "hit your limit\|hit your session limit\|session limit\|monthly usage limit\|weekly limit\|out of extra usage"
```

## Testing

- `bash -n scripts/agents/claude-wrapper.sh` passes (bash 3.2, macOS).
- Scratch harness against the exact regex:
  - `You've hit your session limit · resets 4pm (America/Los_Angeles)` → **TOKEN_EXHAUSTED** (the bug)
  - existing matches (`hit your limit`, weekly, org monthly usage, out of extra usage) still classify as **TOKEN_EXHAUSTED**
  - `session limit reached` variant → **TOKEN_EXHAUSTED**
  - benign non-exhaustion strings (success message, 401 auth, 429 rate limit) → **NOT** matched
- No live fleet disturbed (isolated scratch harness).

## Merge safety

Change is confined to the `classify_error` exhaustion regex + its comment. It does **not** touch the account-selection strategies (~lines 322-384, Strategy 0/1) being edited concurrently by #41727 in a different worktree — the two PRs are on disjoint regions and merge cleanly.

Closes #41737

### DIFF
--- a/scripts/agents/claude-wrapper.sh
+++ b/scripts/agents/claude-wrapper.sh
@@ -631,8 +631,9 @@ classify_error() {
     #   - "hit your limit"                              (legacy short form)
     #   - "You've hit your org's monthly usage limit"   (org cap; multi-word gap defeats `hit.your.limit`)
     #   - "You've hit your weekly limit · resets …"     (weekly cap; same multi-word gap)
+    #   - "You've hit your session limit · resets …"    (5h session cap; same multi-word gap — the word "session" defeats `hit your limit`)
     #   - "You're out of extra usage · resets …"        (session/extra-usage cap)
-    if echo "$output" | grep -qi "hit your limit\|monthly usage limit\|weekly limit\|out of extra usage"; then
+    if echo "$output" | grep -qi "hit your limit\|hit your session limit\|session limit\|monthly usage limit\|weekly limit\|out of extra usage"; then
         echo "TOKEN_EXHAUSTED"
         return
     fi


rjwalters:
## Judge review — changes requested (one small narrowing)

Thanks for the tight, well-documented fix. The core is correct and the bug is real:

- `hit your limit` genuinely can't span the `session` gap, so `You've hit your session limit · resets …` was falling through to the generic non-zero-exit RECOVERABLE branch and never rotating. Confirmed.
- The added **`hit your session limit`** alternative matches the exact live message (the middot `·` is irrelevant — matching the ASCII substring `session limit` is the right approach). ✅
- No 429 regression: a genuine 429 emits `rate limit` / `too many requests` / `429`, none of which contain the substring `session limit`, so the exhaustion branch (which precedes the 429 branch at ~L645) does not swallow a RECOVERABLE 429. ✅
- All four pre-existing phrasings still match; `bash -n` clean; single-file diff disjoint from #41727's selection region (~L322-384) → merges cleanly. ✅

**Requested change — drop the bare `session limit` alternative (keep `hit your session limit`):**

`classify_error` does not short-circuit to SUCCESS on exit 0 — it runs every grep first and only returns SUCCESS at the end (~L674) if nothing matched. So the exhaustion branch fires on **exit-0 output too**, and — unlike the adjacent 429 branch (~L645) — it has **no non-zero-exit guard**. That guard was added specifically so a successful turn that merely *mentions* a limit in prose (the Herald daily-post-limit case) isn't misclassified as an API failure.

The bare `session limit` alternative reintroduces exactly that hazard for a phrase that is currently a very active topic across this fleet (claude-monitor #41033, this issue #41737, rotation spread #41727). An exit-0 success whose transcript contains "session limit" (e.g. a builder/judge/curator turn touching these very issues) would be classified TOKEN_EXHAUSTED → spurious `rotate_to_working_token()` + `consecutive_failures++` on a cycle that actually succeeded.

Since the specific `hit your session limit` already matches the real Anthropic message exactly, the bare `session limit` is redundant and only adds false-positive surface. Please either:

- **(preferred)** remove the bare `session limit`, keeping `hit your session limit`; or
- gate the exhaustion branch on `exit_code -ne 0` the way the 429 branch already is (this would also harden the pre-existing patterns and won't hurt true-positive detection, since a real session cap fails the CLI).

Everything else is good to merge once that's addressed.

rjwalters:
Addressed the Judge feedback (Option A). Dropped the bare `session limit` alternative from the `classify_error` exhaustion regex, keeping `hit your session limit` which matches the real live message (`You've hit your session limit · resets 4pm …`) exactly.

Since `classify_error` has no non-zero-exit guard on the exhaustion branch (it runs on exit-0 output too), the bare `session limit` would have misclassified a successful turn merely mentioning "session limit" in prose (an active fleet topic: #41033/#41737/#41727) as `TOKEN_EXHAUSTED`. The specific phrase is redundant, so removing it eliminates that false-positive surface.

Updated the comment block to document why we match `hit your session limit` rather than the bare substring.

Verification:
- `bash -n scripts/agents/claude-wrapper.sh` clean.
- Live message + all four pre-existing phrasings still match → `TOKEN_EXHAUSTED`.
- Benign exit-0 prose mentioning "session limit" no longer matches.
- Selection-strategy region (~L322-384, owned by #41727/#41744) untouched.

rjwalters:
## Judge re-review — the fix is correct, but it did NOT reach `main` (merge race)

I re-reviewed after the Doctor pass. **The fix commit `3ae71890c5` is exactly what I asked for and is verification-clean** — but PR #41742 was merged **without it**, so `main` currently ships the unfixed broad regex. Details below.

### The fix commit (`3ae71890c5`, on `origin/feature/issue-41737`) — CORRECT
```
grep -qi "hit your limit\|hit your session limit\|monthly usage limit\|weekly limit\|out of extra usage"
```
- [x] bare `session limit` alternative REMOVED; `hit your session limit` retained
- [x] `You've hit your session limit · resets 4pm (America/Los_Angeles)` → TOKEN_EXHAUSTED via `hit your session limit`
- [x] all four pre-existing phrasings still match; benign exit-0 prose mention of "session limit" no longer false-positives
- [x] scope: single file, classify_error exhaustion regex + comment only; selection region (~322–384) untouched
- [x] `bash -n` clean
- [x] comment now documents the no-exit-0-guard rationale explicitly

On its own merits this earns **APPROVE**.

### Problem: the fix is stranded off `main`
PR #41742 is already **MERGED** (merge commit `59e296ad42`), but the merge used the stale head `6cc23684`, not the fix. Timeline (America/Los_Angeles):

| time | commit |
|------|--------|
| 13:26:35 | `6cc23684` original broad regex |
| 13:30:58 | `3ae71890c5` **fix — removes bare `session limit`** |
| 13:31:43 | merge `59e296ad42` into main (took `6cc23684`, stranded the fix) |

`git merge-base --is-ancestor 3ae71890c5 origin/main` → **NOT an ancestor**. Confirmed `origin/main:scripts/agents/claude-wrapper.sh` line 701 still reads:
```
grep -qi "hit your limit\|hit your session limit\|session limit\|monthly usage limit\|weekly limit\|out of extra usage"
```
The bare `session limit` false-positive hazard is **LIVE on main**: any exit-0 successful turn whose transcript merely mentions "session limit" in prose (an active fleet topic — #41033/#41737/#41727) will misclassify as TOKEN_EXHAUSTED and trigger spurious token rotation + `consecutive_failures++`.

### Required action
This is a deployer fast-merge race (late follow-up commit stranded). Since #41742 is merged/closed, the fix cannot be re-merged here. **A fresh follow-up PR must land `3ae71890c5` onto `main`** (cherry-pick that single commit — trivial, no conflicts expected against the current main line 701). Recommend re-opening a Doctor task or a one-line cherry-pick PR. Filing/linking a follow-up is advisable so this doesn't get lost.

Verdict: fix content **approved**, but `main` remains unfixed — needs a follow-up PR to actually deploy the fix.

PR #41744, opened 2026-07-22
fix(wrapper): spread account selection across top-N available accounts (#41727)

### LINKED ISSUE #41727: Account selection always picks the single most-available account — collides with sibling Loom daemon on the shared claude-monitor pool → session-limit faults

rjwalters:
Closes #41727

## Problem
`scripts/agents/claude-wrapper.sh` account selection deterministically picked the **single most-available** account: both the claude-monitor path (Strategy 0) and the cached-ranking path (Strategy 1) `break` on the FIRST eligible entry. With many concurrent workers — plus a sibling **Loom daemon** reading the same `~/.claude-monitor/ranking.json` — everyone piled onto `account[0]`, driving it to its session limit in ~90s while other accounts sat idle.

## Fix
Rank eligible accounts exactly as before, then pick ONE at **random among the top-N** instead of always `[0]`.

- **Strategy 0 (monitor)**: collect all `available`/present/allowlisted/not-bad stems in least-utilized-first order (unchanged `sort_by(util_7d, util_5h)`), then random-pick among the first N. Logs mode `monitor-spread` (N>1) / `monitor` (N=1).
- **Strategy 1 (cached `.ranking`)**: collect the top band of non-exhausted/blocked/bad/missing entries (soonest-reset order), random-pick among the first N. Logs mode `ranked-spread` / `ranked`.
- Shared `_spread_pick` helper.

## Configurable N
- Env var **`LEAN_ACCOUNT_SPREAD_TOP_N`, default 3**.
- N is **clamped to the eligible count**, so the band NEVER widens into exhausted/blocked accounts to fill it — if fewer than N are available, it picks among those.
- **`N=1` reproduces today's exact deterministic least-utilized pick** (a safe rollback). Malformed/`<1` values fall back safely (`0`/`abc` → clamp/default).

## Untouched
Bootstrap/materialization block, ranking.json consumer schema, freshness gates, allowlist/`.bad_tokens` handling, Strategy 2/3 random fallbacks.

## Testing
`bash -n` clean. Verified with a scratch harness (instrumented copy that exits right after selection) against **stub** token dirs and a stub `ranking.json` — the live `.loom/tokens` was not touched:

- (a) N=3, 5 available → picks spread across the top-3, never accounts 4/5, mode `*-spread`.
- (b) N=1 → always the single least-utilized, mode `monitor`/`ranked` (exact old behavior).
- (c) exhausted/blocked/rate_limited accounts never selected, even to fill the band.
- (d) N clamps when fewer accounts are available (e.g. 2 available → band of 2).
- (e) allowlist gating still respected (only allowlisted stems enter the band).
- Both Strategy 0 and Strategy 1 exercised.

### DIFF
--- a/scripts/agents/claude-wrapper.sh
+++ b/scripts/agents/claude-wrapper.sh
@@ -38,15 +38,25 @@ unset CLAUDECODE 2>/dev/null || true
 #
 # Account selection strategy (in priority order):
 #   0. If claude-monitor's ~/.claude-monitor/ranking.json exists and is fresh
-#      (<10 min by generated_at / mtime), pick the least-utilized `available`
-#      account (ordered by status, util_7d, util_5h), joining ranking emails to
-#      on-disk token files via loom's derive_token_filename convention. This is
-#      a SOFT dependency: absent/stale/unparseable export → fall through below.
-#      Mirrors loom 0.12.0 (rjwalters/loom#3697, #3699) + lean-genius #41033.
-#   1. If .loom/tokens/.ranking exists and is <10 min old, pick the first
-#      account that has remaining weekly capacity (sorted by soonest reset).
+#      (<10 min by generated_at / mtime), rank `available` accounts least-utilized
+#      first (ordered by status, util_7d, util_5h) and pick one at RANDOM among
+#      the top-N (env LEAN_ACCOUNT_SPREAD_TOP_N, default 3; #41727), joining
+#      ranking emails to on-disk token files via loom's derive_token_filename
+#      convention. This is a SOFT dependency: absent/stale/unparseable export →
+#      fall through below. Mirrors loom 0.12.0 (rjwalters/loom#3697, #3699) +
+#      lean-genius #41033.
+#   1. If .loom/tokens/.ranking exists and is <10 min old, pick at random among
+#      the top-N accounts (same LEAN_ACCOUNT_SPREAD_TOP_N band) that still have
+#      remaining weekly capacity (sorted by soonest reset).
 #   2. If .loom/tokens/.allowlist exists, random among allowed accounts.
 #   3. Otherwise, random among all .token files.
+#
+# LEAN_ACCOUNT_SPREAD_TOP_N spreads load across the freshest accounts so that
+# many concurrent workers (and a sibling Loom daemon reading the same ranking)
+# don't all collide on the single most-available account and drive it to its
+# session limit (#41727). N is clamped to the eligible count and never widens
+# into exhausted/blocked accounts; N=1 reproduces the old deterministic
+# least-utilized pick exactly (a safe rollback).
 #
 # The ranking file (strategy 1) is written by:
 #   scripts/agents/check-accounts.sh --ranking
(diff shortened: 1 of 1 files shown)


rjwalters:
## Loom Judge — APPROVE

Reviewed against #41727. Clean, well-scoped fix to the greedy account selection. Verified locally on the PR branch.

**Correctness**
- **Both strategies now spread over top-N** instead of `break`-on-first: Strategy 0 collects all eligible stems into `_mon_eligible` (least-utilized-first) then picks among the top band; Strategy 1 does the same into `_rank_eligible` (soonest-reset-first). ✔
- **Critical property — band never widens into ineligible accounts**: every filter (`available` selection, on-disk token presence, allowlist `grep -qxF`, `.bad_tokens`, `exhausted`/`blocked` status, missing token file) runs via `continue` *before* the stem is appended to the eligible array. The band is the first N of an eligible-only array, and N is clamped to `${#eligible[@]}`. An exhausted/blocked/bad/missing account can never be selected even to fill N. ✔
- **N=1 is a faithful deterministic rollback**: tested 200 picks over 5 eligible — always index 0, band 1, mode `monitor`/`ranked` (no `-spread` suffix). `RANDOM % 1 == 0` always yields top-of-band. ✔
- **`$RANDOM` index bounds**: `RANDOM % _band` ∈ `0..band-1`, band ≤ eligible count, so the index into the array is always valid — no off-by-one, no bias into an out-of-band slot. ✔
- **Configurable N** (`LEAN_ACCOUNT_SPREAD_TOP_N`, default 3): tested `0→1`, `abc→3`, `-5→3` (negative fails `^[0-9]+$`), `100 over 4 eligible → 4` (clamped). ✔
- **`_spread_pick` return mechanism**: helper `echo`s `"<index> <band>"`; callers consume via `read -r _pick _spread_top_n <<< "$(_spread_pick …)"`. The here-string runs `read` in the *parent* shell, so both the index and the band size (which drives the mode label) propagate correctly — a global assignment inside the `$()` subshell would indeed have been lost. The builder's concern is handled correctly; `_spread_top_n=1` also provides a safe default. ✔
- **Allowlist + bad-token gating preserved** in both strategies (Strategy 1 never checked the allowlist originally — no regression). ✔

**Scope**
- Single file, +83/-18. All hunks confined to lines 38–450 (header comment + Strategy 0/1 selection). The exhaustion-regex/`classify_error` region (edited by concurrent #41742) and the log-rotation block are untouched — regions disjoint. Bootstrap/materialization block and the ranking.json consumer schema are unchanged. ✔

**Mechanics**
- `bash -n` clean under macOS bash 3.2 (3.2.57); code uses only indexed arrays / `+=` / `[[ =~ ]]` — no bash-4 features. ✔
- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`. No CI checks are triggered for a shell-only change under `scripts/agents/` (expected).

Approving and labeling `loom:pr`.

PR #42453, opened 2026-07-23
fix(daemon): stop misdetecting polling-agent sleeps as COMPLETED

rjwalters:
## Summary

`get_agent_status()` in `scripts/lean/launch.sh` declares a Claude-based
agent `COMPLETED` whenever no `claude` child process is found and the
pane's foreground command is a bare shell name (`bash`/`zsh`/`sh`). But a
non-interactive script never hands off the pty's foreground process group
per subcommand, so the pane still reports `bash` while the wrapper is
inside a `sleep` — e.g. `enforce_cycle_floor`'s `CYCLE_MIN_SECONDS`
cycle-floor sleep (#41507), or the wrapper's own retry/backoff delay.
That's indistinguishable from a truly-finished script under the old
check, so the daemon's health-check loop kills and respawns the agent.

This is why seeker's busy-loop ("Daemon cycle 1" resetting every 1-3
minutes instead of ~30) persisted in `.loom/logs/seeker.log` even after
#41507 merged and the daemon was restarted: the very sleep #41507 added
is what the health check misreads as completion.

Fix mirrors the existing `is_script_based_agent` branch, which already
guards against this with a `has_child_processes` check before declaring
`COMPLETED`. Applied the same guard to the Claude-based branch: if the
pane's shell still has a live child (the `sleep` itself), the agent is
busy, not idle-at-prompt — report `IDLE` for polling agents (seeker,
deployer, auditor, tester, herald, mechanic) or `RUNNING` otherwise.

## Test plan

- [x] `bash -n scripts/lean/launch.sh` — syntax check passes
- [x] Reproduced the exact false-positive locally: sourced the real
      `get_pane_pid`/`find_claude_child`/`get_pane_command`/
      `has_child_processes` helpers from this branch and ran them
      against a `while true; do sleep 30; done` wrapper (mirrors
      `enforce_cycle_floor`'s loop shape) — confirmed `has_child_processes`
      correctly reports a live `sleep` child, so the patched status logic
      now reports `IDLE` instead of the old false `COMPLETED`.
- [ ] Live confirmation: after merge + daemon restart, `seeker.log` cycle
      spacing should widen to ~30min instead of ~1-3min (can be checked
      in a follow-up cycle).

### DIFF
--- a/scripts/lean/launch.sh
+++ b/scripts/lean/launch.sh
@@ -1080,6 +1080,20 @@ get_agent_status() {
     if [[ -z "$claude_pid" ]]; then
         # No claude process - check if shell is at prompt
         if [[ "$pane_cmd" == "zsh" || "$pane_cmd" == "bash" || "$pane_cmd" == "sh" ]]; then
+            # A non-interactive script never changes the pty's foreground process
+            # group per subcommand, so pane_cmd still reports the wrapper's own
+            # shell name (bash) while it's inside a `sleep` — e.g. CYCLE_MIN_SECONDS'
+            # enforce_cycle_floor, or the wrapper's own retry/backoff delay. That
+            # looks identical to "script truly exited to an idle shell prompt"
+            # unless we also check for a live child process.
+            if has_child_processes "$pane_pid"; then
+                if is_polling_agent "$session"; then
+                    echo "IDLE"
+                else
+                    echo "RUNNING"
+                fi
+                return
+            fi
             echo "COMPLETED"
             return
         fi


rjwalters:
## Approved — LGTM

Reviewed the +14/-0 change to `get_agent_status()` in `scripts/lean/launch.sh`. The diagnosis in the PR description is correct and the fix is the right shape.

### Root cause confirmed

Sessions are spawned as an *interactive* shell pane (`tmux new-session -d -s "$session" -c "$worktree_dir"`) and the wrapper is then `send-keys`'d into it. The wrapper is a non-interactive `bash`, so it never hands off the pty foreground process group per subcommand — `pane_current_command` reports `bash` for the wrapper's entire lifetime, including while it sits in `enforce_cycle_floor`'s `CYCLE_MIN_SECONDS` sleep. With no `claude` child during that sleep, the old code fell straight through to `echo "COMPLETED"`, and the daemon loop respawned a perfectly healthy agent.

### Empirical verification (isolated `tmux -L` socket, live agents untouched)

I reproduced the spawn shape exactly and measured both directions:

| Phase | `pane_current_command` | `has_child_processes` | Old verdict | New verdict |
|---|---|---|---|---|
| Wrapper inside its sleep | `bash` (wrapper) | TRUE | `COMPLETED` (false positive) | `IDLE` ✅ |
| Wrapper genuinely exited | `zsh` (interactive shell) | FALSE | `COMPLETED` | `COMPLETED` ✅ |

Note the two phases differ in *which* shell name is reported (`bash` vs `zsh`), but the code accepts `zsh|bash|sh` uniformly — so the shell name alone can never discriminate. The child-process check is exactly the right discriminator.

### Opposite failure mode — checked, and it does not regress

The obvious risk with tightening a completion heuristic is failing to detect a genuinely finished agent, leaving dead agents `IDLE` forever and never respawned. This does not happen:

- `has_child_processes` is `pgrep -P "$pane_pid"`, and `pane_pid` is the **interactive** shell whose only direct child is the wrapper. So the predicate is effectively "is the wrapper still alive" — strictly stronger and more correct than "is there a sleep." Once the wrapper exits, the child count is 0 and `COMPLETED` still fires (Phase 2 above).
- No `send-keys` call site in the file backgrounds a job into the pane (`grep` for trailing `&` on send-keys: zero hits), so there is no lingering-background-child vector that would pin the shell's child count above 0 forever.
- Zombie children would be reaped by the interactive shell within a poll cycle, so any transient miss self-clears.

### Residual gap (non-blocking, pre-existing)

A wrapper that wedges with *no* `claude` child and logs no failures would now be reported `IDLE` indefinitely. The `STUCK` heuristic (elapsed/CPU/network) only runs when a `claude_pid` exists, so it does not cover that case; the backstop is the consecutive-failure check at the top of the daemon loop, which already treats `RUNNING`/`IDLE` + `>= CONSECUTIVE_FAILURE_THRESHOLD` as a forced `kill_and_respawn`. This is the identical trade-off already accepted for the `is_script_based_agent` branch (aristotle/tester) long before this PR, and it is strictly better than the status quo of killing every polling agent on every cycle floor. Worth a follow-up only if a wedged-wrapper case is ever observed in practice — not a blocker here.

### Quality checks

- `bash -n scripts/lean/launch.sh` — passes.
- `shellcheck` — 9 findings on `main`, 9 on the PR branch; **zero** new findings, none in the changed hunk (lines 1075-1105).
- The fix mirrors the existing `is_script_based_agent` branch verbatim in structure, so the two code paths now agree rather than diverge.
- `IDLE` vs `RUNNING` split is consistent with the existing `is_polling_agent` treatment further down in the same function, and both are handled identically by the daemon loop (neither triggers a respawn), so the split is purely observability. (Minor: `tester` appears in `is_polling_agent` but is also `is_script_based_agent`, so it returns from the earlier branch and never reaches this code — harmless dead-ish condition, no action needed.)
- Change is purely additive (+14/-0); no existing behavior is altered other than the specific false positive.

### CI / merge state

- `gh pr checks` reports **"no checks reported on the branch."** This repository has **no CI configured on PR branches** (only `cleanup-branches.yml` and `label-external-issues.yml`, neither of which runs on `pull_request`). To be explicit: this is the expected state for this repo and is **not** a passing CI signal — no automated checks ran, so the verification above is manual and local, not CI-attested.
- `mergeStateStatus`: `CLEAN`, `mergeable: MERGEABLE`, base `main`.

Approving. The one item worth calling out for whoever merges: the live confirmation checkbox in the test plan (seeker cycle spacing widening to ~30min) is still unchecked and can only be validated after merge + daemon restart — worth actually checking `.loom/logs/seeker.log` afterward rather than assuming.

rjwalters:
**Champion Auto-Merge**

---
*Automated by Champion role*

PR #38463, opened 2026-07-13
Adopt configurable worktree root across the lean fleet (LOOM_WORKTREE_ROOT / worktree.root) — resume of #37509

rjwalters:
Closes #37509

## Summary

Ports Loom v0.10.6's worktree-root resolver into the lean fleet's own scripts. Every fleet worktree construction site (enricher, researcher, erdos, aristotle, auditor, deploy, mechanic, peer-reviewer, seeker, deploy temp/batch-rebase worktrees, `launch.sh` respawn paths) now resolves its base via `scripts/lib/worktree-root.sh`:

1. `LOOM_WORKTREE_ROOT` env var (absolute; relative → stderr warning + default)
2. `.loom/config.json` → `worktree.root` (jq-guarded; malformed/missing → soft-fail to default)
3. default `$REPO_ROOT/.loom/worktrees` — byte-for-byte unchanged when no override is set

Overrides are namespaced by repo basename, so `worktree.root = "/Volumes/Stripe"` resolves to `/Volumes/Stripe/lean-genius/<name>` on the dedicated 3.6 TiB volume. Cleanup/GC surfaces (`clean-branches.sh` dual-root scan + outside-pass classification, `clean-research.sh`, `clean-enhancers.sh`, `infra-guardian.sh` reaper allow-pattern) service BOTH the resolved root and the legacy default during the transition.

## Provenance: salvaged from a stranded builder worktree

This is a **resume** of the original #37509 builder, which died 2026-07-11 leaving ~20 files of uncommitted work in `/Volumes/Stripe/lean-genius/issue-37509` (branch `feature/issue-37509`). Its base files were mass-deleted from main hours later (dc9fdffa30) and restored byte-identical in #38423, so the stranded diff was salvaged (`git diff` → patch) and applied with `git apply --3way` onto a fresh branch off current main — cleanly, no conflicts.

**Salvaged (dead builder's work, ~440 lines):** resolver lib, all construction-site wiring, all cleanup/GC wiring, `worktree-cleanup.sh` prose, CLAUDE.md doc section.

**New in this resume:**
- `scripts/tests/worktree-root.test.sh` (deliverable 5, unwritten by the dead builder): 23 assertions — full resolver semantics (env/config/default/relative-fallback/trailing-slash/env-beats-config/malformed-JSON/null-key) + janitor dual-scan, outside-pass INSIDE classification, and infra-guardian reaper allow-pattern replicas.
[...]

PR #43291, opened 2026-07-24
Fix: infra-guardian reaps the deployer's own worktree as "merged"

rjwalters:
## Summary
- `infra-guardian.sh`'s `reap_worktrees()` reaps any worktree under the resolved worktree root whose branch is an ancestor of `origin/main`. The deployer's fixed worktree lives at that root and fast-forwards `feature/deployer` to exactly match `origin/main` every cycle, so the "merged" check is trivially true — the guardian deletes the deployer's own worktree registration mid-cycle, every cycle.
- Observed 2026-07-24: the reap fired between the deploy pipeline's "Sync Branch" step and the PR-merge loop, deregistering `/Volumes/Stripe/lean-genius/deployer` out from under an in-flight run (files intact, `.git/worktrees/deployer` admin dir gone) and requiring manual `git worktree add` recovery mid-cycle.
- Excludes the `deployer` worktree by basename from the merged-reap check — it's a fixed, reused-every-cycle role worktree, not a per-task worktree like `enricher-N`/`researcher-N`.

## Test plan
- [x] `bash -n scripts/lean/infra-guardian.sh` — syntax check passes
- [ ] Next deployer cycle should complete without the worktree being reaped mid-run

### DIFF
--- a/scripts/lean/infra-guardian.sh
+++ b/scripts/lean/infra-guardian.sh
@@ -133,6 +133,19 @@ reap_worktrees() {
         esac
         grep -qxF "$wt" <<<"$locked" && continue
 
+        # Fixed, long-lived role worktrees are reused every cycle rather than
+        # created per-task, and some (the deployer) fast-forward their branch
+        # to exactly match origin/main as part of normal operation -- which
+        # makes the "merged" check below trivially true every cycle. Without
+        # this guard the guardian reaps the deployer's own worktree mid-run
+        # (observed 2026-07-24: fired between the deploy pipeline's "Sync
+        # Branch" step and the rest of the cycle, deregistering the worktree
+        # out from under an in-flight PR-merge loop and requiring manual
+        # `git worktree add` recovery).
+        case "${wt##*/}" in
+            deployer) continue ;;
+        esac
+
         if [[ "$ref" != "DETACHED" ]] && git merge-base --is-ancestor "$ref" origin/main 2>/dev/null; then
             _reap_one "$wt" "merged"
             continue


rjwalters:
## Judge review: APPROVED

Reviewed the full 13-line diff against `origin/main`, traced every consumer of `reap_worktrees()`, and verified the diagnosis end-to-end. This is a correct, well-scoped, low-risk fix to a standing fleet bug.

### The diagnosis is exactly right

`scripts/deploy/launch-agent.sh:106` runs `git checkout -B "$BRANCH_NAME" origin/main` every cycle, so `feature/deployer` is *by construction* identical to `origin/main`. `git merge-base --is-ancestor feature/deployer origin/main` therefore returns 0 unconditionally (a commit is its own ancestor), and the merged-reap arm fires on every guardian pass. Confirmed `WORKTREE_PATH="$WORKTREES_DIR/deployer"` (`launch-agent.sh:55`), so the basename is exactly `deployer` and the new guard matches the real path under every worktree root — `/Volumes/Stripe/lean-genius/deployer`, `.loom/worktrees/deployer`, and `/private/tmp/*` alike.

Verified the guard's matching behavior directly:

```
EXEMPT   /Volumes/Stripe/lean-genius/deployer
EXEMPT   /Users/x/repo/.loom/worktrees/deployer
reapable /Volumes/Stripe/lean-genius/deployer-2
reapable /Volumes/Stripe/lean-genius/my-deployer
reapable /Volumes/Stripe/lean-genius/researcher-1
```

No per-task worktree can be collaterally exempted: `deployer-N`/`my-deployer` still reap, and no other script in `scripts/` creates a worktree whose basename is bare `deployer`.

### Does it fail to reap things it should? (the opposite error)

Worth stating precisely, because the PR description understates the change. The description says it excludes the deployer "from the merged-reap check" — in fact the `continue` exits the entire loop body, so the deployer worktree is now exempt from **both** reap arms: `merged` *and* `stale>${WORKTREE_STALE_HOURS}h`.

**I consider that the correct behavior, not a defect**, and it should not be narrowed:

- On the `merged` arm, exemption loses no information. The deployer's branch is always an ancestor of main by design, so that signal is constant-true for this worktree and never carried any evidence about whether the worktree was actually disposable.
- On the `stale` arm, keeping the check live for the deployer would *reintroduce the same incident under a different trigger*. The staleness test reads `git -C "$wt" log -1 --format=%ct` — the HEAD **commit's** committer date, not worktree activity. A deployer that is running perfectly but sitting on a quiet `origin/main` for longer than `WORKTREE_STALE_HOURS` would be reaped mid-cycle exactly as before. The broad `continue` closes that variant too.

The only real cost: a *stopped* deployer's worktree is now never reclaimed, including via the low-disk emergency path (`check_disk()` → `reap_worktrees()`). That is bounded to a single worktree on a 3.6 TiB dedicated volume, and is plainly the better side of the trade against deregistering a live deployer mid-PR-merge-loop. Please just correct the PR description (or the commit body) to say "excluded from reaping" rather than "from the merged-reap check", so the next reader does not narrow the guard back to the merged arm believing that was the intent.

### Consistent with existing precedent

This is not an ad-hoc carve-out. `scripts/clean-branches.sh:342` already hardcodes the sibling protection for the same reason:

```bash
echo "feature/deployer" >> "$PROTECTED_BRANCHES_FILE"
```

with a comment describing the same failure mode. The two cleanup paths now agree, which is what you want — a fix in only one of them would have left the deployer reapable by the other.

### Verification performed

- `bash -n` on the patched `scripts/lean/infra-guardian.sh` — **passes** (re-ran the PR's own test-plan step independently).
- `case` + `continue` semantics confirmed against the loop; matches the two `case`/`continue` guards already present a few lines above, so it is stylistically in-idiom.
- `git merge-tree origin/main <pr-head>` — **0 conflicts**; `mergeStateStatus` is `CLEAN`. `infra-guardian.sh` has not changed on `main` since this PR's base (#38463), so the 2-week age introduced no drift.
- Confirmed the bug is **still live on `main`** (`git show origin/main:scripts/lean/infra-guardian.sh` has no guard) — this PR is still needed, not superseded.
- **CI: this repository has no checks configured on PR branches.** `gh pr checks` reports *"no checks reported on the 'fix/infra-guardian-deployer-reap' branch"*, and `.loom/config.json` defines no `buildGate`. To be explicit: CI did **not** pass, because there is no CI to pass. Validation here rests on the syntax check and the static reasoning above, not on a green build.
- Test plan item 2 ("next deployer cycle completes without the worktree being reaped") is **observation-only and not executed** — it requires a live deployer cycle. Please confirm it on the next run.

### Non-blocking follow-ups

1. **Generalize via `git worktree lock` instead of per-role basenames.** `reap_worktrees()` already honors locks (`grep -qxF "$wt" <<<"$locked" && continue`), and `scripts/lib/worktree-cleanup.sh:159` already handles `git worktree unlock`. Having each singleton launch-agent lock its worktree at startup would cover every role with no hardcoded names and no further edits to the guardian.
2. **The same class affects the other singletons.** `aristotle`, `seeker`, `auditor-N`, `mechanic-N`, `tester`, `herald`, and `peer-reviewer-N` are all created with `git worktree add "$WORKTREE_PATH" -b "$BRANCH_NAME" main`, so their branch *is* `main` at creation and is an ancestor of `origin/main` until the agent's first commit — reapable by this identical arm during that window. That is the documented "worktree reaped before its first commit" failure mode. This PR deliberately fixes only the observed deployer incident, which I think is the right scope for this change, but the general case deserves its own issue (follow-up 1 would resolve it wholesale).

Neither blocks merge.

### Assessment

Correct root-cause analysis, minimal blast radius (a pure `continue` guard that cannot alter behavior for any other path), consistent with existing precedent in the codebase, and carrying an unusually good explanatory comment that records the incident date and rationale — precisely the comment that stops someone from "simplifying" the guard away later. High value for the deployer fleet.

Approving: `loom:review-requested` → `loom:pr`.

rjwalters:
**Champion Auto-Merge**

---
*Automated by Champion role*

Issue #43033, opened 2026-07-24
Aristotle recovery follow-up: review recovered results, verify SumOfOdds on v4.31, restart agent, add submission dedup


rjwalters:
Follow-up to #43006 (PR pending). The 100 untracked finished Aristotle server projects were retrieved and triaged; tracking is reconciled and wedge detection is in place. The following residual work needs attention.

## 1. Review recovered results saved under `research/aristotle-runs/`

- `recovered-958405df-8534-4c33-9d40-21529cfa14fa.lean` — Vahlen–Capelli T0 submission (tracked job from 2026-07-12). Aristotle did NOT prove the target `two_power_capelli_neg_square` (still sorry) but returned proved supporting lemmas (`quad_repr`, `quad_indep`, `descent`, `two_power_irred`). Relevant to the cube-root-3-irrational-oq-02-oq-03 Vahlen–Capelli vein — a researcher should assess whether the lemma layer is worth incorporating.
- `recovered-212b28eb-...lean` — `TractatusFunctionalCompletenessAristotle`, 0 sorries (fully solved). Check whether its content is already incorporated into the `TractatusOntology*` / `TractatusCompleteness` files on main; integrate if not.
- `recovered-{1efa3c7d,57df7541,5d77873e,e203a760,298662ff}-...lean` — Tractatus family results (Horn / Totality / NGeneration / Decidability / expressibility), 1-2 sorries each. Manual research submissions (server names `ar-*`); companion files were never committed locally. Review against the Tractatus formalization.
- `recovered-3e8027f6-...lean` — `aristotle-v431-probe` (trivial `probe_log_mul`); can be discarded.
- `recovered-3705d2cc-...lean` — AM-GM OQ02 companion, no improvement; can be discarded after a glance.

## 2. Verify the one integrated proof on the v4.31 pin

`proofs/Proofs/SumOfOddsStatementOnly.lean` had its only sorry closed by Aristotle project `09e00bf0` (`induction n <;> simpa [Finset.sum_range_succ] using by linarith`). Integrated with `v431_verification: "pending"` per the standard retrieve-integrate.sh convention. Run:

    ./scripts/aristotle/verify-v431-build.sh Proofs.SumOfOddsStatementOnly

and flip `v431_verified` in `research/aristotle-jobs.json` on success.

## 3. Restart the live aristotle-agent after the recovery PR merges

[...]

Issue #43006, opened 2026-07-24
[Epic #43004] Recover the Aristotle pipeline: 99 untracked finished server projects, submissions blocked


rjwalters:
**Epic**: #43004 - Restore research capacity
**Phase**: 1 of 3
**Phase Goal**: Recover & rebalance existing capacity

## Description

Aristotle — the engine's only scalable proof-search arm — has been wedged since at least 2026-07-12 and nobody noticed because its agent loop shows green in health checks.

Current state (from the agent's own cycle output):
- Server active projects: 0 (limit 5), but **99 finished (IDLE) projects on the server are not tracked locally**
- The backlog guard sees the untracked projects and **skips all submissions** ("Recovery pipeline may be broken")
- `research/aristotle-jobs.json` tracks exactly 1 job (submitted 2026-07-12, now expired); lifetime yield: 0 wins of 0 terminal jobs
- 989 candidates remain queued with nothing flowing

The 99 finished projects may contain completed proofs that were never retrieved.

## Tasks

1. Run `./scripts/aristotle/retrieve-integrate.sh` (per the agent's own warning) to pull the 99 finished server projects; triage results — integrate real proofs, archive junk. Note `research/aristotle-jobs-archive-zombies.json` exists from a prior cleanup; follow that pattern for stale projects.
2. Reconcile `research/aristotle-jobs.json` with server state so the backlog guard's untracked-project count returns to 0 and submissions unblock.
3. Verify `./scripts/aristotle/submit-batch.sh` flows again (target ≥3 active jobs from the 989-candidate queue).
4. Add wedge detection: the agent/health check should flag (not just log) when untracked-server-projects > 0 or when no job has reached a terminal state in N days, so this failure mode surfaces in `launch.sh health` instead of showing RUNNING.

## Acceptance Criteria

- [ ] 99 untracked server projects triaged: results retrieved, integrated or explicitly archived with reasons
- [ ] `aristotle-jobs.json` reconciled with server; backlog guard no longer blocks submissions
- [ ] ≥3 active submissions flowing; `check-jobs.sh --update` reflects live status
- [ ] `launch.sh health` (or the agent status block) reports a warning state when the pipeline is wedged

## Dependencies

Part of Epic #43004. Phase 1 issue with no blocking dependencies.

---
*Created by Epic skill from Epic #43004*

PR #43649, opened 2026-08-04
aristotle: add submission dedup guard (Part of #43033)

rjwalters:
## Summary

Part of #43033 (Aristotle recovery follow-up). This PR closes item 4 (submission dedup guard) with tests; items 1, 2, and 3 are documented below with recommended next steps rather than force-completed, for reasons explained per item.

## Item 4 — submission dedup guard (done)

Root cause of the 90-duplicate incident: after `research/aristotle-jobs.json` tracking was lost, the existing "already submitted" filters (`get_submitted_files`/`get_blocked_files`), which key off the *current* status of the *most recent* job, had nothing to dedupe against.

- Added `get_repeat_offender_files()` to `scripts/aristotle/find-candidates.sh`: a file submitted `ARISTOTLE_DEDUP_MAX_ATTEMPTS`+ times (default 3, any status) across the whole jobs.json history with no `"integrated"` job among them is excluded from candidate selection — independent of, and a backstop for, the current-status filters.
- Verified against the live `research/aristotle-jobs.json`: correctly flags `ChebyshevBounds` and `SchroederBernstein` (the two repeat offenders from the incident) and does not flag `SumOfOddsStatementOnly` (which did eventually integrate).
- The other half of item 4 — excluding 0-sorry files from candidate selection — was already implemented in `collect_candidates()` (`total_sorry -eq 0 → skip`); no change needed there.
- New isolated unit test: `scripts/tests/aristotle-dedup-guard.test.sh` (5/5 passing), independent of the live proof corpus or a running Aristotle pipeline so it stays fast and deterministic. Run: `bash scripts/tests/aristotle-dedup-guard.test.sh`.

## Item 2 — verify SumOfOddsStatementOnly on v4.31 (deferred, not this PR)

Attempted but deferred: the host's Docker daemon was unresponsive (`docker info` timed out) under heavy concurrent load from other Loom sweep workers at the time of this session. Per CLAUDE.md, `lake build` must never run outside the Docker wrapper, so I did not force it. `research/aristotle-jobs.json` already correctly reflects this as `v431_verified: false`, `v431_verification: "pending"` (the script's own documented convention for a skipped gate) — no data was falsely marked verified. Follow-up: `./scripts/aristotle/verify-v431-build.sh Proofs.SumOfOddsStatementOnly` once Docker is healthy.

## Item 1 — triage of recovered results under `research/aristotle-runs/` (reviewed, no integration in this PR)

`research/aristotle-jobs-archive-recovered.json` already carries a full per-project triage classification from the #43006 recovery. I reviewed the 9 files still worth a look and confirmed/extended that record:

- `recovered-212b28eb-...lean` (`TractatusFunctionalCompletenessAristotle`, 0 sorries, self-contained, Mathlib-only) — genuinely new content, **not yet present** in `proofs/Proofs/` (checked: no `TractatusFunctionalCompleteness*` file exists). Ready for a Researcher/Enricher to add as a new proof once a build can be verified.
- `recovered-{1efa3c7d,57df7541,5d77873e,e203a760,298662ff}-...lean` (Tractatus Horn/Totality/NGeneration/Decidability/Expressibility family) — each still carries 1-2 unresolved sorries; these are exploratory leads toward the companion `rjwalters/tractatus` repo, not mergeable as-is. Left for research follow-up.
- `recovered-958405df-...lean` (Vahlen–Capelli) — target theorem `two_power_capelli_neg_square` still sorry; contains proved supporting lemmas (`quad_repr`, `quad_indep`, `descent`, `two_power_irred`). Mathematical-value judgment call for a Researcher, not made here.
- `recovered-3e8027f6-...lean` (trivial `probe_log_mul`) and `recovered-3705d2cc-...lean` (AM-GM OQ02, no improvement over target) — confirmed discardable, no action needed.

I did not add any new `.lean` files to `proofs/Proofs/` in this PR, since Docker was unavailable to verify a build (see item 2) and CLAUDE.md's Axiom Integrity Policy requires careful, verified gallery status assignment that a build failure would undermine.

## Item 3 — restart the live aristotle-agent (operator action, out of scope for a diff)

Not applicable to a code PR — restarting a running process is an operational action for whoever/whatever manages `aristotle-agent`, to be done after this PR merges (`./scripts/aristotle/launch-agent.sh --stop && ./scripts/aristotle/launch-agent.sh`).

## Item 5 — server backlog beyond newest 100

Per the issue body, no action required (submit-batch guard only inspects the newest 100; no overlap with previously downloaded results).

## Test plan

- [x] `bash scripts/tests/aristotle-dedup-guard.test.sh` — 5/5 passing
- [x] `bash -n scripts/aristotle/find-candidates.sh` — syntax OK
- [x] Manual jq check of the new query against the live `research/aristotle-jobs.json` — correctly identifies the two known repeat offenders and does not falsely flag the successfully-integrated file
- [ ] Full `./scripts/aristotle/find-candidates.sh --count` smoke test against the live 6000+ file corpus — started but did not finish within this session due to heavy host I/O contention (many concurrent Loom sweep workers); recommend a reviewer re-run if host load has settled

### DIFF
--- a/scripts/aristotle/find-candidates.sh
+++ b/scripts/aristotle/find-candidates.sh
@@ -44,6 +44,16 @@ JOBS_FILE="$PROJECT_ROOT/research/aristotle-jobs.json"
 OPEN_CONJ_REGISTRY="$PROJECT_ROOT/research/open-conjectures.json"
 ARISTOTLE_RUNS_DIR="$PROJECT_ROOT/research/aristotle-runs"
 
+# Repeat-submission dedup guard (issue #43033). A file submitted this many
+# times or more (any status, summed across the whole jobs.json history)
+# without ever reaching "integrated" is a repeat offender: further
+# resubmission is unlikely to make progress and just burns server capacity.
+# This is a backstop independent of get_submitted_files()/get_blocked_files()
+# above, which key off the *current* status of the *most recent* job(s) and
+# can be fooled by a jobs.json history that lost or never recorded
+# intermediate attempts (the root cause of #43006's 90-duplicate incident).
+ARISTOTLE_DEDUP_MAX_ATTEMPTS="${ARISTOTLE_DEDUP_MAX_ATTEMPTS:-3}"
+
 # Parse arguments
 COUNT_ONLY=false
 BEST_N=0
@@ -120,6 +130,23 @@ get_blocked_files() {
     done | sort -u
 }
 
+# Files submitted ARISTOTLE_DEDUP_MAX_ATTEMPTS+ times across all of jobs.json
+# history (any status) with no "integrated" job among them. Excluded from
+# candidate selection regardless of the current-status filters above.
+get_repeat_offender_files() {
+    if [[ ! -f "$JOBS_FILE" ]]; then
+        return
+    fi
+
+    jq -r --argjson max "$ARISTOTLE_DEDUP_MAX_ATTEMPTS" '
+        .jobs
+        | group_by(.file)
+        | map(select(length >= $max and all(.[]; .status != "integrated")))
+        | map(.[0].file)
+        | .[]
+    ' "$JOBS_FILE" 2>/dev/null | xargs -I{} basename {} .lean | sort -u
+}
+
 # Analyze a single file. Returns score -1 for hard rejects.
 # Args: file, tier (0, 1 or 2)
 analyze_file() {
@@ -165,7 +192,8 @@ collect_candidates() {
     local tier="$1"
     local submitted_files="$2"
     local blocked_files="$3"
-    shift 3
+    local repeat_offender_files="$4"
+    shift 4
     local files=("$@")
 
     for file in "${files[@]}"; do
@@ -180,6 +208,12 @@ collect_candidates() {
 
         # Skip if blocked (repeatedly failed, not modified)
         if echo "$blocked_files" | grep -q "^${basename}$"; then
+            continue
+        fi
+
+        # Skip repeat offenders (ARISTOTLE_DEDUP_MAX_ATTEMPTS+ submissions,
+        # never integrated) — see get_repeat_offender_files().
+        if echo "$repeat_offender_files" | grep -q "^${basename}$"; then
             continue
         fi
 
@@ -331,6 +365,9 @@ main() {
     local blocked_files
     blocked_files=$(get_blocked_files)
 
+    local repeat_offender_files
+    repeat_offender_files=$(get_repeat_offender_files)
+
     local candidate_data=()
 
     # Tier 0: StatementOnly files (*StatementOnly.lean) — Harmonic format
@@ -342,7 +379,7 @@ main() {
     if [[ ${#tier0_files[@]} -gt 0 ]]; then
         while IFS= read -r line; do
             [[ -n "$line" ]] && candidate_data+=("$line")
-        done < <(collect_candidates 0 "$submitted_files" "$blocked_files" "${tier0_files[@]}")
+        done < <(collect_candidates 0 "$submitted_files" "$blocked_files" "$repeat_offender_files" "${tier0_files[@]}")
     fi
 
     # Tier 1: Companion files (*Aristotle.lean)
@@ -354,7 +391,7 @@ main() {
     if [[ ${#tier1_files[@]} -gt 0 ]]; then
         while IFS= read -r line; do
             [[ -n "$line" ]] && candidate_data+=("$line")
-        done < <(collect_candidates 1 "$submitted_files" "$blocked_files" "${tier1_files[@]}")
+        done < <(collect_candidates 1 "$submitted_files" "$blocked_files" "$repeat_offender_files" "${tier1_files[@]}")
     fi
 
     # Tier 2: Regular proof files (unless --tier1-only).
@@ -380,7 +417,7 @@ main() {
         if [[ ${#tier2_files[@]} -gt 0 ]]; then
             while IFS= read -r line; do
                 [[ -n "$line" ]] && candidate_data+=("$line")
-            done < <(collect_candidates 2 "$submitted_files" "$blocked_files" "${tier2_files[@]}")
+            done < <(collect_candidates 2 "$submitted_files" "$blocked_files" "$repeat_offender_files" "${tier2_files[@]}")
         fi
     fi
 
--- a/scripts/tests/aristotle-dedup-guard.test.sh
+++ b/scripts/tests/aristotle-dedup-guard.test.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+# Guards the repeat-submission dedup guard in
+# scripts/aristotle/find-candidates.sh (get_repeat_offender_files).
+#
+# Root cause (issue #43033): after research/aristotle-jobs.json tracking was
+# lost, the existing "already submitted" filters (which key off the *current*
+# status of the most recent job) had nothing to dedupe against, and the
+# candidate queue re-served the same top files every cycle — 90 of 100
+# recovered projects were duplicate submissions of just three files. This
+# test verifies the independent repeat-offender backstop: a file submitted
+# ARISTOTLE_DEDUP_MAX_ATTEMPTS+ times (any status, summed across all of
+# jobs.json history) with no "integrated" job among them is excluded from
+# candidate selection, regardless of what the status-based filters allow.
+#
+# Run: bash scripts/tests/aristotle-dedup-guard.test.sh
+# Exits non-zero if any assertion fails.
+set -u
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+FIND_CANDIDATES="$SCRIPT_DIR/../aristotle/find-candidates.sh"
+
+PASS=0
+FAIL=0
+
+pass() { echo "  ok: $1"; ((PASS++)); }
+fail() { echo "  FAIL: $1"; ((FAIL++)); }
+
+TMPDIR_TEST="$(mktemp -d)"
+trap 'rm -rf "$TMPDIR_TEST"' EXIT
+
+JOBS_FILE="$TMPDIR_TEST/aristotle-jobs.json"
+
+# Extract get_repeat_offender_files() so it can be sourced in isolation
+# without pulling in main()'s side effects (it invokes `main` unconditionally
+# at the bottom of the real script).
+guard_src="$(awk '
+    /^get_repeat_offender_files\(\) \{/ { capture=1 }
+    capture { print }
+    capture && /^\}$/ { exit }
+' "$FIND_CANDIDATES")"
+
+if [[ -z "$guard_src" ]]; then
+    fail "could not locate get_repeat_offender_files() body in $FIND_CANDIDATES"
+    echo ""
+    echo "PASS=$PASS FAIL=$FAIL"
+    exit 1
+fi
+
+run_guard() {
+    local max="$1"
+    (
+        ARISTOTLE_DEDUP_MAX_ATTEMPTS="$max"
+        eval "$guard_src"
+        get_repeat_offender_files
+    )
+}
+
+# 1) A file submitted 3+ times with no "integrated" job is a repeat offender.
+cat > "$JOBS_FILE" <<'EOF'
+{"jobs": [
+  {"file": "proofs/Proofs/ChebyshevBounds.lean", "status": "failed"},
+  {"file": "proofs/Proofs/ChebyshevBounds.lean", "status": "expired"},
+  {"file": "proofs/Proofs/ChebyshevBounds.lean", "status": "submitted"}
+]}
+EOF
+out="$(run_guard 3)"
+if grep -qx "ChebyshevBounds" <<<"$out"; then
+    pass "3 non-integrated submissions -> flagged as repeat offender"
+else
+    fail "3 non-integrated submissions should be flagged (got: $out)"
+fi
+
+# 2) A file submitted 3+ times that eventually reached "integrated" is NOT
+#    excluded — the whole point of the guard is stopping unproductive churn,
+#    not permanently banning files that did eventually succeed.
+cat > "$JOBS_FILE" <<'EOF'
+{"jobs": [
+  {"file": "proofs/Proofs/SumOfOddsStatementOnly.lean", "status": "failed"},
+  {"file": "proofs/Proofs/SumOfOddsStatementOnly.lean", "status": "expired"},
+  {"file": "proofs/Proofs/SumOfOddsStatementOnly.lean", "status": "integrated"}
+]}
+EOF
+out="$(run_guard 3)"
+if grep -qx "SumOfOddsStatementOnly" <<<"$out"; then
+    fail "a file that eventually integrated must not be a permanent repeat offender (got: $out)"
+else
+    pass "file with an 'integrated' job in its history is never flagged"
+fi
+
+# 3) Below the threshold: 2 submissions with max=3 should not be flagged.
+cat > "$JOBS_FILE" <<'EOF'
+{"jobs": [
+  {"file": "proofs/Proofs/SchroederBernstein.lean", "status": "failed"},
+  {"file": "proofs/Proofs/SchroederBernstein.lean", "status": "expired"}
+]}
+EOF
+out="$(run_guard 3)"
+if [[ -z "$out" ]]; then
+    pass "2 submissions below default threshold (3) -> not flagged"
+else
+    fail "2 submissions should be below threshold (got: $out)"
+fi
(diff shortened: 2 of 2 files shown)


rjwalters:
## Changes Requested

Reviewed at head `876776560a` in an isolated worktree (`/Volumes/Stripe/lean-genius/pr-43649`). The scoping of this PR is good — the item-by-item write-up of what was *not* done (items 1/2/3) is honest and appropriately conservative, and "Part of #43033" is the right framing for a partial increment. The dedup guard itself is clean, well-commented, correctly wired through all three `collect_candidates` call sites, and reproduces exactly as claimed on the live data. One blocker below.

### Verification performed

- `bash scripts/tests/aristotle-dedup-guard.test.sh` → **5/5 PASS**
- `bash -n` on both changed files → OK
- `shellcheck -S warning` on both files → no *new* findings (the `SC2155` hits are pre-existing style in untouched code; the `SC2034` in the test is a false positive — the var is consumed by the `eval`'d function body)
- Replayed the guard's `jq` against the live `research/aristotle-jobs.json` (100 jobs): flags `ChebyshevBounds` (31 jobs, never integrated) and `SchroederBernstein` (31 jobs, never integrated), does not flag `SumOfOddsStatementOnly` (28 jobs, one integrated). Matches the PR description exactly.
- **CI: "no checks reported on the `feature/issue-43033` branch."** This repo has no `pull_request`-triggered workflow (`.github/workflows/` contains only `cleanup-branches.yml` and `label-external-issues.yml`), so this is the expected absence of CI — it is **not** a green CI signal, and nothing about this PR was validated by CI. The verification above is entirely local. Merge state is `CLEAN`.
- Body scanned for auto-closing keywords: **none present**. `closes item 4` is prose, not a `closes #N` reference, and the commit subject carries a bare `(#43033)`. Nothing here will auto-close #43033 on merge, which is correct for a partial increment.

### Blocker: the idempotency key has no un-exclusion path

`get_repeat_offender_files()` keys on *"count of jobs for this `.file` in jobs.json, with no `integrated` among them"*. That key is monotonically increasing and never re-evaluated against the file's current content, so **crossing the threshold excludes a file from candidate selection permanently**. Reproduced with fixtures:

| Scenario | jobs.json | Result |
|---|---|---|
| A | 3x `failed`, file subsequently repaired/rewritten | `Foo` still excluded — forever |
| B | 2x `completed` + 1x `submitted` (zero failures) | `Bar` excluded anyway |
| C | jobs.json corrupt/unparseable | empty list — guard silently **fails open** |
| D | history pruned below threshold | silently un-excluded |

Scenario A is the problem. The adjacent `get_blocked_files()` — 40 lines above, same file — already solves precisely this, and deliberately: it blocks on 2+ failures **but unblocks when the file's mtime is newer than `last_submitted`**, because "resubmission won't help" is only true while the file is unchanged. The new guard has no such check, and because it runs as an independent `continue` in `collect_candidates()`, it **silently defeats that existing escape hatch** for any file with 3+ total jobs. A Researcher who does the documented thing — complete the definitions, convert `axiom` to `theorem … := by sorry`, resubmit — cannot get the file back into the queue. The only recoveries are hand-editing `research/aristotle-jobs.json`, or raising `ARISTOTLE_DEDUP_MAX_ATTEMPTS` globally, which also disables the guard for the genuine offenders it exists to stop.

Two aggravating factors:

1. **The threshold counts any status, not failures.** `submitted` and `completed` count toward it (scenario B), so a file that never failed once can be permanently excluded while its results are still in flight to `retrieve-integrate.sh`. `get_blocked_files()` deliberately counts only `failed`/`expired`. Given the default of 3, "any status" is aggressive.
2. **The skip is silent.** `collect_candidates()` just `continue`s. With no log line, an operator watching a file quietly vanish from `--count` output has no way to attribute it to this guard — the same observability gap that let the original 90-duplicate storm run unnoticed.

**Requested fix** (small — reuse the pattern already written in this file): gate the repeat-offender exclusion on the same mtime comparison `get_blocked_files()` uses, so a file modified since its last submission is *not* treated as a repeat offender. Emitting a one-line skip notice (behind the existing verbose/log convention) would close the observability gap. A test for the "repaired after N attempts → eligible again" case should accompany it.

### Non-blocking notes

- **The backstop shares a single point of failure with what it backstops.** The PR body describes the guard as "independent of, and a backstop for, the current-status filters." It is independent of the *filters*, but not of the *data source* — scenario C shows a `2>/dev/null` swallow that yields an empty list on any malformed jobs.json. Since #43033's root cause was *jobs.json tracking being lost*, this guard is equally blind in that exact scenario. Consistent with the existing filters, so not a regression — but worth not over-claiming in the comment header.
- **`unknown` phantom entry.** 7 live jobs have `"file": "unknown"`; they group together, clear the threshold, and emit a bogus `unknown` offender. Harmless (no file matches), but those 7 attempts are also unattributed, so the guard undercounts for whichever files they really belong to. Pre-existing data quality issue, not caused by this PR.
- **Test covers the function, not the wiring.** The test `awk`-extracts and `eval`s `get_repeat_offender_files()` in isolation, so the `shift 3` → `shift 4` change and the three updated call sites are untested. I verified all three by inspection and they are correct; a missed one would have corrupted the `files` array silently. Worth a smoke assertion.
- Test plan item 4 (full `find-candidates.sh --count` against the live corpus) is still unchecked. I did not run it either — with the blocker above outstanding it is better re-run after the fix.

Happy to re-review promptly once the mtime escape hatch is in.

rjwalters:
Doctor: mtime escape hatch + skip notice implemented and committed locally; tests now 14/14. Running the full `find-candidates.sh --count` against the live corpus (test-plan item 4) before pushing — 5,941 proof files, takes several minutes per run, and I'm running the pre-fix revision alongside it for a before/after comparison. Still treating.

rjwalters:
## Feedback addressed — un-exclusion path, skip notice, and wiring tests

Head is now `bf49620ab2`. Thanks for the precise diagnosis; the blocker was real and the fix is the one you pointed at.

### 1. Blocker: the idempotency key now has an un-exclusion path

`get_repeat_offender_files()` adopts the same mtime rule `get_blocked_files()` already uses. The jq stage now emits `file|count|last_submitted` (taking the newest `submitted` in the group) instead of just the path, and the shell loop drops a file from the offender list when its mtime is newer than that timestamp — i.e. a file repaired after N attempts is a candidate again, exactly as it is after 2 failures under `get_blocked_files()`. Scenario A from your table now resolves; the guard no longer shadows the existing escape hatch.

**One deliberate asymmetry, called out because it is load-bearing:** the check fails *closed* when "modified since" cannot be established — no parseable `submitted` value, or the recorded path no longer exists on disk. An unknown timestamp is not evidence of repair, and this is not hypothetical: **99 of the 100 live jobs carry `"submitted": "unknown"`** (the #43006 recovery records; only one job in the file has a real timestamp). Failing *open* there would un-exclude every legacy offender on the first `touch`, disabling the backstop precisely in the corrupt-history case #43033 exists for. Going forward `submit-batch.sh` writes `date -u +"%Y-%m-%dT%H:%M:%SZ"`, the exact format `get_blocked_files()` parses, so the hatch works for every job recorded from now on; repairing a record's `submitted` field re-enables it retroactively. The rationale is in the function's header comment so the next reader does not have to rediscover it.

### 2. Observability: the skip is no longer silent

Added `dedup_log()` — one line to **stderr** (stdout stays machine-readable; every existing caller already invokes the script with `2>/dev/null`: `submit-batch.sh:618`, `aristotle-agent.sh:165,270`, `launch.sh:2086`, `status.sh:95`), suppressible with `ARISTOTLE_QUIET=true` per the `prune-zombies.sh` / `drift-repair.sh` convention. `collect_candidates()` now attributes every repeat-offender skip, and the guard logs each un-exclusion. Live example from the `--count` run below:

```
find-candidates: ChebyshevBounds (tier 2): skipped by the repeat-submission dedup guard — >= 3 jobs in research/aristotle-jobs.json, none integrated, and no modification recorded since the last submission (modify the file, or raise ARISTOTLE_DEDUP_MAX_ATTEMPTS, to requeue)
```

### 3. Tests: 5 → 14 cases

New, on top of the original five (all still passing, unmodified):

| # | Case |
|---|---|
| 6 | repaired after 3 attempts (mtime > `last_submitted`) → **eligible again** (your requested case) |
| 7 | still unmodified → still excluded |
| 8 | `"submitted": "unknown"` + freshly touched file → fails closed, still excluded |
| 9 | the un-exclusion notice goes to stderr, stdout stays empty |
| 10 | **wiring:** the real `collect_candidates()` emits the probe file with all three filter lists empty (proves `shift 4` left the file args intact) |
| 11 | **wiring:** each filter list filters from its own positional slot (loop over slots 2/3/4) |
| 12 | the repeat-offender skip emits an attributable notice |
| 13 | `ARISTOTLE_QUIET=true` silences the notices |
| 14 | **wiring (static):** exactly 3 `collect_candidates` call sites pass `"$repeat_offender_files"`, and exactly one `shift 4` |

Cases 10–13 `eval` the real `collect_candidates` / `analyze_file` / `count_matches` / `dedup_log` bodies (the extractor is now a generic `extract_fn`), so the wiring you verified by inspection is now executed, not just grepped. I mutation-tested the new assertions rather than trusting them:

| Mutation | Result |
|---|---|
| `shift 4` → `shift 3` | case 14 FAILs |
| escape-hatch condition → `if false` | cases 6 and 9 FAIL |
| skip `dedup_log` line deleted | case 12 FAILs |
| repeat-offender `if` block deleted from `collect_candidates` | cases 11, 12, 13 FAIL |

### 4. Your non-blocking item: counting only `failed`/`expired` — declined, with data

I checked this before changing it, and it would break the guard: **every repeat offender in the live `research/aristotle-jobs.json` has status `archived`** — 31× `ChebyshevBounds`, 31× `SchroederBernstein`, 7× the `unknown` phantom. Zero `failed`, zero `expired`. Narrowing the threshold to `failed`/`expired` would flag *nothing* on the real corpus and make the guard a strictly weaker duplicate of `get_blocked_files()`. Your scenario B (in-flight `submitted`/`completed` counting toward the threshold) is now much narrower anyway: any file whose in-flight work is accompanied by an edit escapes via the mtime hatch, and a genuinely in-flight file is already held by `get_submitted_files()`. Left as-is deliberately; no test fixtures changed for it.

The other non-blocking notes I have left alone as out of scope for this PR: the `2>/dev/null` fail-open on a malformed `jobs.json` (pre-existing, shared with the other filters — I did not change the behaviour, but the header comment now says explicitly that the guard is independent of those *filters*, not of their *data source*, so the over-claim you flagged is gone), and the 7 `"file": "unknown"` records (pre-existing data quality).

### Verification

- `bash scripts/tests/aristotle-dedup-guard.test.sh` → **14/14 PASS** (5 original + 9 new)
- `bash -n` on both changed files → OK
- `shellcheck -S warning` → **no new findings.** `find-candidates.sh` is byte-identical in findings to the pre-fix revision (10× `SC2155` + 1× `SC2034`, all pre-existing in untouched code); the test file is now **clean** (0 findings) — the `SC2034` false positive you noted is suppressed with an inline `# shellcheck disable` naming the reason.
- Branch applies cleanly to current `origin/main` (`git merge-tree --write-tree` exits 0), so I left the history alone — no rebase, one new commit on top.
- **Test-plan item 4 (your unchecked item), now run.** Full `find-candidates.sh --count` against the live corpus (5,941 files under `proofs/Proofs/`), from an isolated worktree, read-only:

  ```
  $ bash scripts/aristotle/find-candidates.sh --count
  find-candidates: ChebyshevBounds (tier 2): skipped by the repeat-submission dedup guard — >= 3 jobs ...
  find-candidates: SchroederBernstein (tier 2): skipped by the repeat-submission dedup guard — >= 3 jobs ...
  992
  ```

  **992 candidates.** I ran the pre-fix revision (`876776560a`) side by side in the same worktree for a controlled comparison: it also returns **992**. So the fix is behaviour-preserving on the current corpus — the two legacy offenders are still excluded (their `submitted` is `"unknown"`, so the hatch correctly does not fire), the only difference being that the two exclusions are now visible and attributable instead of silent. Each run takes ~4 minutes; note in passing that `launch.sh:2086` / `status.sh:95` wrap this call in `timeout 10`, so those two dashboards have been silently reading `0` — pre-existing, unrelated to this PR, not touched here.

Nothing outside `scripts/aristotle/find-candidates.sh` and `scripts/tests/aristotle-dedup-guard.test.sh` was modified.

rjwalters:
## Approved — the re-judge blocker is fixed

Re-review of head `bf49620ab2` (one Doctor commit on top of the previously reviewed `876776560a`). The prior verdict had exactly one blocker: `get_repeat_offender_files()` excluded a file **permanently and silently** once it crossed the attempt threshold, with no mtime escape hatch. That is resolved.

### Blocker verification

1. **Escape hatch is the same rule `get_blocked_files()` uses.** Both resolve `$PROJECT_ROOT/$file`, read mtime via `stat -f '%m'` → `stat -c '%Y'`, parse the last recorded `submitted` via `date -j -f '%Y-%m-%dT%H:%M:%SZ'` → `date -d`, and compare the two. `get_blocked_files()` blocks when `mtime <= submit_epoch`; the new guard lifts the exclusion when `submit_epoch > 0 && mtime > submit_epoch`. Same comparison, same fallback chain.
2. **It fails closed where the adjacent filter fails open — deliberately, and documented.** With an unparseable timestamp `submit_epoch=0`, so `get_blocked_files()` un-blocks (`mtime > 0`) while the new guard keeps the exclusion; a missing path is likewise `continue` (un-block) there vs. still-excluded here. For a backstop whose whole reason to exist is a jobs.json that lost history, "unknown timestamp is not evidence of repair" is the right direction, and the header comment states it plus the remedy. Test 8 pins it.
3. **Skips are no longer silent.** `dedup_log()` writes to `>&2`; every repeat-offender skip in `collect_candidates()` and every lifted exclusion emits one line. `ARISTOTLE_QUIET=true` suppresses. Verified stdout stays machine-readable both by tests 9/12/13 and live: a `--count` run emitted two notices on stderr with stdout empty. All four consumers already discard stderr — `aristotle-agent.sh:165,270`, `launch.sh:2086`, `status.sh:95` (`--count 2>/dev/null`) and `submit-batch.sh:618` (`--json ... 2>/dev/null`).

### Checks run

- `bash scripts/tests/aristotle-dedup-guard.test.sh` → **PASS=14 FAIL=0** (was 5 tests). Coverage now includes the repaired-after-N-attempts case, the unparseable-timestamp fail-closed case, both stderr notices, `ARISTOTLE_QUIET`, and real-`collect_candidates()` wiring (per-slot filter assertions + the static `shift 4` / three-call-site check, which is the class of defect a positional-arg change actually risks).
- `bash -n` clean on both files. `shellcheck -S warning` on `find-candidates.sh` is **byte-identical in findings to `origin/main`** (10× SC2155, 1× SC2034, all on pre-existing lines — the new function correctly splits `local` from assignment); the new test file is shellcheck-clean.
- `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`. `gh pr checks` reports no checks — this repo has only `cleanup-branches.yml` / `label-external-issues.yml` and no `buildGate.command` in `.loom/config.json`, so the scoped test run above *is* the gate, not a skipped one.

### The declined suggestion holds up

Doctor declined the prior non-blocking "count only failed/expired" note with data, and the data checks out: both live offenders are 31 jobs of status `archived` with `"submitted": "unknown"`, zero `failed`/`expired` — restricting the count to failed/expired would flag nothing at all.

I also verified the "992 before and after" claim mechanically rather than taking it on assertion: `ChebyshevBounds.lean` and `SchroederBernstein.lean` both contain **zero** `sorry` occurrences, so `analyze_file()` hard-rejects them regardless; the guard therefore removes no candidate that would otherwise have been submitted today, and its live effect is the two notices.

### Non-blocking observations (no re-work requested)

- Every *currently live* offender carries `"submitted": "unknown"`, so the mtime escape hatch does not apply to any of them today — their escape routes are repairing the `submitted` field or raising `ARISTOTLE_DEDUP_MAX_ATTEMPTS`. That is the documented fail-closed path and it is now attributable via the stderr notice, which is what the blocker asked for; worth knowing if someone later adds sorries to one of those two files and wonders why it stays out of the queue.
- `([.[].submitted // empty] | sort | last)` sorts lexicographically, so a group mixing real ISO timestamps with `"unknown"` picks `"unknown"` (`u` > digits) and fails closed even though a parseable timestamp exists. Conservative in the safe direction; a `select(test("^[0-9]{4}-"))` filter would be tighter if this ever matters.
- A non-numeric `ARISTOTLE_DEDUP_MAX_ATTEMPTS` makes `--argjson` fail and `2>/dev/null` then disables the guard silently. Same swallow-stderr idiom as the surrounding functions, so not a regression.

Nothing here is the original blocker recurring, and nothing here is a new defect worth another cycle. LGTM.

PR #107, opened 2025-12-26
Add proof card for Schroeder-Bernstein theorem (Wiedijk #25)

rjwalters:
## Summary

- Adds proof card data for the Schroeder-Bernstein theorem (Wiedijk #25)
- Includes comprehensive annotations explaining the orbit-based proof strategy
- Documents the theorem's connection to cardinality theory and Cantor's theorem

The Lean file was already merged in PR #87. This PR completes issue #102 by adding the proof card UI data.

## Test plan

- [x] Lean file compiles successfully
- [x] TypeScript/Vite build passes
- [ ] Verify proof card renders correctly in the UI

Closes #102

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #84, opened 2025-12-26
Add Fermat's Sum of Two Squares theorem proof

rjwalters:
## Summary
- Add Fermat's Sum of Two Squares theorem: A prime p can be expressed as p = a² + b² iff p = 2 or p ≡ 1 (mod 4)
- Complete Lean 4 proof using Mathlib's implementation based on Zagier's elegant one-sentence proof
- Full frontend integration with meta.json, annotations, and tactic states

## Changes
- `proofs/Proofs/FermatTwoSquares.lean` - Complete proof with pedagogical examples
- `src/data/proofs/fermat-two-squares/` - Frontend data files
- `src/data/proofs/index.ts` - Add proof to navigation

## Test plan
- [ ] Verify frontend build succeeds (`pnpm build`)
- [ ] Verify Lean proof compiles (`lake build Proofs.FermatTwoSquares`)
- [ ] Check proof appears in site navigation
- [ ] Review annotations for mathematical accuracy

Closes #68

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #43682, opened 2026-08-04
chore(repo-skills): sync installed files to source c925fb2 (v0.7.0 line)

rjwalters:
Restores the gitignored .install-local.json sidecar (deleted by a pull, rjwalters/repo#96) and commits the tracked-file deltas from the same sanctioned installer run: post-tag doc fixes for all/followups/release/reset/tidy plus a guard-destructive.sh update. Part of the /repo:all update-tools stage.

https://claude.ai/code/session_01NNqnwQi5KJhHV3hpQB1eU1

### DIFF
--- a/.claude/commands/repo/all.md
+++ b/.claude/commands/repo/all.md
@@ -50,14 +50,87 @@ report. This surfaces everything before anything is touched.
 Offer to fix gitignore findings here. Leave README, link, and documentation
 fixes for the Docs stage next — don't apply them twice.
 
-### 2. Docs (see [[docs]])
+### 2. Sync early, if and only if nothing can be lost (see [[reset]])
+
+[[reset]] is two separable halves, and only one of them is safe to move:
+
+- **Sync-and-switch** (reversible): `git fetch`, check out the default branch,
+  `git pull --ff-only`. Nothing is removed.
+- **Pruning** (gated): stash review, branch & worktree deletion. Can
+  permanently remove work, so it keeps its gates and stays last (stage 6).
+
+Running the whole of Reset last assumes the working branch is the right place
+for the later stages to be looking — which holds when it carries unpushed work
+those stages should see. When the branch is **fully pushed and behind the
+default branch** that assumption inverts: there is no local-only content to
+protect, and Docs/Tidy either report drift that is already fixed upstream or
+edit a copy that pollutes an open PR's diff and blocks the branch switch Reset
+performs at the end. So check the branch state here, before Docs runs:
+
+```bash
+current=$(git symbolic-ref --short HEAD 2>/dev/null) || current=""
+default=$(git symbolic-ref --short refs/remotes/origin/HEAD 2>/dev/null | sed 's|^origin/||')
+git fetch origin --quiet
+
+eligible=no
+if [ -n "$current" ] && [ -n "$default" ] && [ "$current" != "$default" ] \
+   && [ -z "$(git status --porcelain)" ]; then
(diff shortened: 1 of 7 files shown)


rjwalters:
## Approved — Repo Skills sync verified against source c925fb2

Evaluated per request (PR carried no loom labels; adding `loom:pr` on approval).

### 1. Scope: only Repo Skills-managed paths
All 7 changed files live under installer-managed roots — 5 under `.claude/commands/repo/`, 2 under `.claude/skills/repo/`. No source, build, workflow, or gallery files touched. File mode on the hook is preserved (`100755` → `100755`).

### 2. Fidelity: byte-identical to upstream
Diffed every file in the PR head against `/Users/rwalters/GitHub/repo` @ `c925fb28b6a9` (`fix(guard): make ml_segment() heredoc-aware so body lines are not command words (#107)`):

```
commands/repo/{all,followups,release,reset,tidy}.md   -> no diff
hooks/repo/guard-destructive.sh                        -> no diff
```

Baseline sanity: `main`'s copies of `tidy.md` / `all.md` / `release.md` are likewise byte-identical to source @ `bc9fe24`, confirming the installer applies no template substitution to these files — so byte-identity is the correct expectation, not a coincidence.

### 3. `guard-destructive.sh` — sane shell, and the guard is NOT weakened
- `bash -n` clean; `shellcheck -S warning` clean.
- Upstream suite run against the exact bytes this PR installs (`hooks/repo/tests/test-guard-destructive.sh`): **536/536 passed, 0 failed**, avg 32ms (<200ms perf gate).
- Hand-probed adversarial heredoc shapes directly against the PR's file:

| Case | Result |
|---|---|
| dangerous command AFTER a heredoc terminator | `deny` |
| dangerous command inside a quoted-delimiter body | `deny` (raw catastrophic scan still fires) |
| `$(...)`-smuggled payload in a bare-delimiter body | `deny` (the expansion-capable carve-out holds) |
| `cat <<EOF \| git reset --hard origin/main` | `ask` (opener line still segments at the pipe) |
| benign `gh issue create --body "$(cat <<'EOF' … )"` (the #84 false positive) | allowed |
| plain `git status` | allowed |

The change is strictly a false-positive fix: heredoc bodies stop producing phantom top-level segments, while the `ALWAYS_BLOCK_PATTERNS` scan still reads the raw command string, and unquoted-delimiter bodies carrying `$(`/backtick revert to the legacy separator-active floor. The `<<<` here-string, arithmetic left-shift (`$((1 << 3))`), comment (`# <<EOF`), and line-continuation edge cases are each explicitly handled and covered by the suite.

### 4. `install-metadata.json` — valid and consistent
Parses cleanly (`python3 -m json.tool`). Only the `commit` field moves, `bc9fe24` → `c925fb2`, matching source HEAD. `version` stays `0.7.0`, matching the source `VERSION` file; `dev: false` is correct for an installed (non-symlinked) copy. The 19-entry `commands` array exactly matches the 19 files in `.claude/commands/repo/` — the one apparent gap versus source (`tests`) is a directory in `commands/repo/`, not a command, so its absence is correct.

### CI
This repo has no `pull_request`-triggered workflows (only `cleanup-branches.yml` and `label-external-issues.yml`), so `no checks reported` is expected rather than a gap; there is no `check:ci` script in `package.json`. Merge state was `CLEAN`. The changed paths are markdown and a hook script outside the `eslint`/build pipeline, so no repo-side check applies.

No issue-linking keyword is required — this is a chore sync from a sanctioned installer run, not issue-tracked work.

Approving.

PR #43617, opened 2026-07-27
chore(repo): hygiene pass — untrack the reap sentinel, fix broken `make help` / `make status`, dangerous build docs

rjwalters:
Findings from a `/repo:all` audit sweep. Every item below was reproduced before fixing.

---

## 1. Root cause of the recurring worktree-reaping incidents

`.loom-managed` was **committed to `main`**, carrying a stale marker for issue 41675.

`.loom/scripts/worktree.sh:88` writes this sentinel into worktrees Loom creates. `agent-destroy.sh:147` (and `merge-pr.sh`) then treat its **presence** as authorization to `git worktree remove`, and its **absence** as "user-owned — refuse to remove" (issue #3334). The file's own header states this contract.

`.gitignore:33` already lists `.loom-managed`, but an ignore rule is inert against an already-tracked path. So every `git worktree add … origin/main` checks the sentinel out into the new worktree root, **the ownership guard never fires, and any worktree becomes reapable.**

Measured on this host: **44 of 60 worktrees carry it** — including the one created minutes earlier purely to hold this PR's edits.

Untracking restores the intended contract. Loom-created worktrees still get a sentinel from `worktree.sh` and are still cleaned up normally; hand-made worktrees are protected again.

## 2. Functional fixes

**`make help` exited 2 on every invocation.** An unclosed double-quote in the `Options:` echo made `/bin/sh` die with `unexpected EOF`. The `Options:` block never printed, and since `help` is `.DEFAULT_GOAL`, bare `make` failed too. `CLAUDE.md` directs every agent to "Run `make help`". Broken since 96d25016b9 (2026-07-12).

**`make status` threw a shell error on every run.** `scripts/lean/status.sh` used `_live_agents=$(… | grep -cE '…' || echo 0)`. With no match, `grep -c` prints `0` *and* exits 1, so `||` fired and the value became `"0\n0"` — then `[[ "$_live_agents" -gt 0 ]]` threw `syntax error in expression`. Fallback moved off the pipeline.

## 3. Safety fixes

`CLAUDE.md` carries a boxed DANGER banner — direct `lake build` "can consume 100GB+ memory in seconds and crash the host system" — and `proofs/bin/` ships a wrapper to block it. Two committed docs told readers to do exactly that:

- `proofs/README.md` — the Build step and the macOS Sequoia workaround
- `CONTRIBUTING.md` — onboarding step 5

Both now use `./proofs/scripts/docker-build.sh` with the safety note.

[...]

PR #43683, opened 2026-08-04
chore(loom): update Loom install 0.12.0 -> 0.18.0 (closes #43675)

### LINKED ISSUE #43675: Loom install stale: daemon dispatch_sweep broken (missing spawn-worker.sh, unresolvable loom_tools) — patched with untracked local copies
## Context

During the 2026-08-04 `/sweep all` session, every `mcp__loom__dispatch_sweep` into this repo failed with an opaque "failed to spawn sweep child". Two-layer root cause:

1. The current loom-daemon (v0.10+/0.12, built 2026-08-03) resolves its spawn binary as `.loom/scripts/spawn-worker.sh` — this repo's Loom scripts (installed Jul 21) predate that dispatcher and only have `spawn-claude.sh`.
2. The older python-based `spawn-claude.sh` needs the `loom_tools` package at `<repo>/loom-tools/src` (script-relative resolution); it is not installed anywhere importable on this host.

Patched in-session with **untracked** workarounds: `spawn-worker.sh` copied from `/Users/rwalters/GitHub/loom/defaults/scripts/`, and a repo-root `loom-tools` symlink to `/Users/rwalters/loom-wt-ratelimit/loom-tools` (excluded via `.git/info/exclude`). These survive only on this machine and will silently vanish on a fresh clone. `.loom/install-metadata.json` claims `loom_version: 0.12.0` but the scripts on disk are older.

## Suggested acceptance criteria

- [ ] Re-run the Loom installer (or otherwise sync `.loom/scripts/` with the loom repo's `defaults/scripts/`) so `spawn-worker.sh` and its `lib/` deps are properly installed
- [ ] Daemon `dispatch_sweep` into this workspace succeeds without the untracked symlink/copy
- [ ] Remove the temporary `loom-tools` symlink and `.git/info/exclude` entry once unneeded

rjwalters:
## Summary

Re-run of the Loom installer from `/Users/rwalters/GitHub/loom` (0.18.0, commit 6ff58955), closing #43675: the 0.12-era install lacked `spawn-worker.sh`, which 0.18 daemons resolve as the sweep spawn binary — every `dispatch_sweep` into this workspace failed until patched with untracked copies. This lands the update properly (~300 files: scripts, roles, commands, docs, hooks).

Local divergences kept on top of the installer output (all in the final commit):

- **`.gitignore` keeps `.loom/config.json` ignored** — the installer removed the rule (upstream now ships a default config), but this fleet's copy is host-local runtime state (`worktree.root = /Volumes/Stripe`).
- **`.loom/tokens` tracked symlink removed + `.loom/tokens/` ignored** — main carried a self-referential symlink at that path while the real directory holds live OAuth tokens; without the ignore, a `git add -A` after removing the symlink could commit secrets.
- `.claude/settings.json` merge verified purely additive (worktree-paths guard, Stop hook, `/private/tmp` permission aliases); the custom permission allowlist is preserved.
- Rebased onto current main and ignore rules deduped; `git ls-files -i -c --exclude-standard` is empty.

⚠️ Note for reviewer: the installer targeting this **linked worktree** also staged deletions in the **primary checkout** (bug, being filed upstream); the primary was fully restored from HEAD before this PR.

https://claude.ai/code/session_01NNqnwQi5KJhHV3hpQB1eU1


rjwalters:
## APPROVED — Loom 0.12.0 -> 0.18.0 vendored update

Reviewed at `df3210368c` (includes the follow-up bare-`.lake` gitignore commit). Bulk vendored-tooling update, so this was a risk-surface review rather than a 308-file line review.

### 1. Security — PASS

- `.loom/tokens` deleted from the index. It was mode `120000` pointing at its own absolute path (`.loom/tokens -> /Users/.../.loom/tokens`) — a self-referential loop, so no real token directory was ever reachable through it. `git ls-files -- .loom/tokens` is now empty.
- `.loom/tokens/` is ignored (`.gitignore:9`) and `.loom/config.json` is ignored (`.gitignore:32`) — both survive the update. Worth noting for the record: the trailing slash is exactly why the symlink escaped the ignore in the first place (dir patterns don't match symlinks), so removing it closes that hole rather than relying on the ignore.
- `git ls-files -i -c --exclude-standard` is empty on the branch.
- `.loom-managed` is ignored and untracked (the #43617 regression has not returned).
- No token/secret material anywhere in the diff. The only secret-shaped strings are the obviously-synthetic `ghs_supersecrettoken...` fixtures in the upstream test `.loom/scripts/tests/test-github-app-token.sh`, byte-identical to upstream.
- Zero new `/Users/rwalters` host paths in added lines; `install-metadata.json`'s `loom_source` line is unchanged from main.

### 2. `.claude/settings.json` — PASS, purely additive

Verified structurally (parsed JSON, not eyeballed diff):

- `permissions.allow`: 60 -> 62 entries, **zero removals**. The only additions are `Read(/private/tmp/**)` and `Edit(/private/tmp/**)`, the macOS aliases of the pre-existing `/tmp` rules. No permission broadened beyond that.
- `hooks.UserPromptSubmit` and `hooks.SessionStart`: unchanged.
- `hooks.PreToolUse`: all pre-existing entries preserved; one `Edit|Write` matcher added.
- `hooks.Stop`: newly added.
- All three referenced hook scripts exist and are tracked at mode `100755`: `guard-loom-workflow.sh`, `guard-worktree-paths.sh`, `guard-background-subagents.sh`.

### 3. `.gitignore` — PASS

Purely additive: **zero removed lines**. No duplicate entries. `.loom/config.json`, `.loom/tokens/`, `.loom-managed` all preserved.

The bare `.lake` restore is correct and I verified the rationale empirically rather than taking it on faith: the primary checkout has `.lake -> proofs/.lake` and currently reports `?? .lake` in `git status`, proving `.lake/` does not match it. Keeping both `.lake` and `.lake/` is intentional, not a duplicate.

### 4. Installed-file integrity — PASS (full comparison, not a spot-check)

Compared against `/Users/rwalters/GitHub/loom` @ `6ff58955`:

| Tree | Identical | Differs | Missing | Extra |
|---|---|---|---|---|
| `.loom/scripts` | 215 | 0 | 0 | 0 |
| `.loom/hooks` | 18 | 0 | 0 | — |
| `.loom/roles` | 12 | 0 | 0 | — |
| `.loom/docs` | 27 | 0 | 0 | — |
| `.claude/commands/loom` | 32 | 0 | 0 | — |

The five requested scripts (`spawn-worker.sh`, `merge-pr.sh`, `worktree.sh`, `lib/worktree-root.sh`, `sweep-checkpoint.sh`) are byte-identical with no substitutions applied. `bash -n` is clean on all **185** changed shell scripts. `check-label-descriptions.sh` passes (max description 99/100 chars).

### 5. Deletions — PASS, all upstream-intentional

Three deletions total: `.loom/tokens` (intended), plus `.loom/scripts/clean-labels.sh` and `.claude/agents/loom-shepherd.md`. Neither exists in current upstream defaults — `clean-labels.sh` survives only inside stale loom worktree branches, and `loom-shepherd.md` appears nowhere in the loom repo. `.claude/README.md` documents the v0.10.0 shepherd deprecation. Remaining `loom-shepherd` mentions are help text and migration notes; the one genuinely dangling reference (`.claude/commands/shepherd.md` -> `.loom/scripts/loom-shepherd.sh`) is pre-existing debris on main and untouched here.

### 6. Critical outcome for #43675 — PASS

`.loom/scripts/spawn-worker.sh` exists, tracked at mode `100755`, byte-identical to upstream, `bash -n` clean.

### Also checked

- `AGENTS.md` 413 -> 3 lines is safe: the old file was generated boilerplate (it literally said `**Loom Version**: unknown`) with **zero** lean-genius-specific content — no mention of lean, proofs, erdos, mathlib, enricher, aristotle, docker-build, or pnpm. The full guide now lives at `.loom/AGENTS.md` (155 lines). Project `CLAUDE.md` is untouched.
- Merge state `CLEAN`, branch 0 commits behind main, working tree clean.
- **CI**: `gh pr checks` reports none. This is expected, not a gap — the repo has only `cleanup-branches.yml` and `label-external-issues.yml`, neither PR-triggered, and there is no `check:ci` script in `package.json`. The PR touches zero files under `src/` or `proofs/`.

---

## Three findings — none blocking, but read #3 before merging

**1. `.github/labels.yml` now has duplicate definitions (follow-up).** 50 entries, only 31 unique names. The installer appended a marker-delimited `# BEGIN LOOM LABELS` block with the canonical set, but the repo's 19 pre-existing unmarked `loom:` / `tier:` definitions above the marker were not removed. 12 of the 19 pairs have *conflicting descriptions*.

The current effective outcome is benign — `sync-labels.sh` parses top-to-bottom and the in-block copy comes later, so the canonical description wins — but it costs 19 redundant API calls per sync and guarantees permanent divergence, since every future upgrade refreshes the block while the stale copies above it never move. Suggested follow-up: delete only the duplicated `loom:`/`tier:` entries above the BEGIN marker, keeping the genuinely repo-own labels (the Research Pipeline block) outside it, which is exactly what the block's own header instructs.

**2. One merge-time blocker in the primary checkout (operational).** `/Users/rwalters/GitHub/lean-genius/.loom/scripts/spawn-worker.sh` currently exists as an **untracked** manual patch, byte-identical to the branch's tracked copy. I confirmed empirically in a scratch repo that git refuses the merge even when the untracked content matches exactly. The GitHub-side merge is fine, but `merge-pr.sh`'s subsequent `git pull --ff-only` in the primary checkout will abort with "untracked working tree files would be overwritten".

Fix before pulling: `rm /Users/rwalters/GitHub/lean-genius/.loom/scripts/spawn-worker.sh`. I enumerated every file added by this branch against the primary checkout — this is the **only** such blocker.

**3. `.loom/config.json` is missing from the primary checkout — fleet-critical, please restore.** This is outside the diff (the file is gitignored) but is a direct casualty of the installer run this PR represents, and the PR body assumes the host-local copy survived. It did not:

- The primary checkout has **no** `.loom/config.json` at all.
- The copy the installer wrote into the worktree is the **stock upstream default** — it has no `worktree` key whatsoever, so no `worktree.root` and no `guards.forceScope`.
- The primary's `.loom/config/` directory was written at 00:08, inside the installer-bug window the PR body describes. Because the file is gitignored, the `git checkout HEAD -- .` restore of the primary could not have brought it back.

Per CLAUDE.md, `worktree-root.sh` resolves `LOOM_WORKTREE_ROOT` > `.loom/config.json` -> `worktree.root` > default `$REPO_ROOT/.loom/worktrees`. With the config gone, every newly created agent worktree (enricher-N, researcher-N, erdos-N, aristotle, sweep worktrees) lands on the **boot disk** instead of the 3.6 TiB `/Volumes/Stripe` volume — the documented precursor to the disk-full SIGBUS 135/139 failure mode. Recreate the host-local config with `worktree.root = "/Volumes/Stripe"` (and `guards.forceScope`) before the fleet spawns anything new.

---

Blocking the merge would not fix findings 2 or 3 — both are host state that already diverged — and #43675 is actively breaking daemon dispatch, so the update should land. The diff itself is disciplined and correct on every risk axis: additive config, no permission broadening, no secrets, no ignore-rule loss, and byte-exact vendored content. Approving.

PR #43693, opened 2026-08-07
security: remove infrastructure identifiers and real addresses from tracked files

rjwalters:
Found by a disclosure sweep across all 73 public repos (2AMLogic/2am#88 — private). Four leaks live at HEAD in this public repo:

- `scripts/deploy/check-account.sh` pinned the Cloudflare account ID **and** the account holder's personal email as literals, printing both on failure. This is the **third copy** of this guard in the estate (after 2AMLogic/klayout-tools and rjwalters/kicad-tools) — it propagated by copying, so the pattern may exist in repos not yet scanned.
- `scripts/deploy/sync-and-deploy.sh` printed the account ID in an error hint.
- `wrangler.toml` named the account ID and email in a comment.
- `scripts/agents/{check,sync}-token-registry.sh` used three real addresses as worked examples of the email→token-filename derivation.

## Approach

The account expectation is now a **SHA-256 digest**. Reading it from the environment would be the weaker fix: this guard exists to catch a wrong account that is *already logged in*, so an expectation the environment can move is one a wrong environment moves with it. The ID is 128 bits of hex, so the digest does not expose it. Failure output prints digest prefixes.

Token-registry examples become reserved `example.com/.net/.org` addresses, teaching the identical derivation rule.

## Verification

Used the script's own `WRANGLER_WHOAMI_OUTPUT` hook — correct account passes, wrong account exits 1, `--help` works. `bash -n` clean on all four scripts; `wrangler.toml` still parses.

An earlier pass renamed the variable without updating every reference, which under `set -u` failed with an unbound variable instead of comparing. Caught by running it, not by reading it.

## Deliberately unchanged

`wrangler.toml`'s D1 `database_id` (wrangler needs the literal to bind D1) and `SUBMISSION_EMAIL` (where the app actually sends submissions). Both are functional values, not incidental leaks, and are flagged for a separate decision.

### DIFF
--- a/.loom/docs/token-pool.md
+++ b/.loom/docs/token-pool.md
@@ -175,8 +175,8 @@ Behavior notes:
 - Only `is_active = 1` rows are imported; `expires_at` is **not** used as a filter
   (observed rows carry stale timestamps while still authenticating — health comes
   from `loom-daemon tokens check`).
-- Token filenames use the same derivation as `bootstrap` (`robb@2amlogic.com` →
-  `robb-2amlogic.token`), so an account keeps one identity across both paths and
+- Token filenames use the same derivation as `bootstrap` (`alice@example.com` →
+  `alice-example.token`), so an account keeps one identity across both paths and
   re-importing overwrites in place.
 - Idempotent: unchanged tokens are left untouched. `index.json` records
   `source: monitor-db` (distinct from the `monitor` snapshot) and, as always,
--- a/scripts/agents/check-token-registry.sh
+++ b/scripts/agents/check-token-registry.sh
@@ -9,7 +9,7 @@
 # claude-monitor migration (#41033/#41044/#41045) the PRIMARY source is
 # ~/.claude-monitor/accounts.env, which carries EMAIL + KEY only:
 #
-#     ACCOUNT_EMAIL_N       account email  (e.g. robb@2amlogic.com)
+#     ACCOUNT_EMAIL_N       account email  (e.g. alice@example.com)
 #     ACCOUNT_KEY_N         the OAuth token value                    (SECRET)
 #     ACCOUNT_TOKEN_FILE_N  OPTIONAL explicit token filename (legacy .env only)
 #
@@ -18,9 +18,9 @@
 # strip '.' and '-' from the email local-part, append '-<first-domain-label>',
 # lowercase, drop unsafe chars, then '.token'. Examples:
 #
-#     robb@2amlogic.com             -> robb-2amlogic.token
-#     r.j.walters@gmail.com         -> rjwalters-gmail.token
-#     robb@integratedplasmonics.com -> robb-integratedplasmonics.token
+#     alice@example.com             -> alice-example.token
+#     a.b.carol@example.net         -> abcarol-example.token
+#     dave@example.org              -> dave-example.token
 #     agent-1@2amlogic.com          -> agent1-2amlogic.token
 #     agent-2@2amlogic.com          -> agent2-2amlogic.token
 #
--- a/scripts/agents/sync-token-registry.sh
+++ b/scripts/agents/sync-token-registry.sh
@@ -10,9 +10,9 @@
 # the email local-part, append '-<first-domain-label>', lowercase, drop unsafe
 # chars, then '.token'. Examples:
 #
-#     robb@2amlogic.com             -> robb-2amlogic.token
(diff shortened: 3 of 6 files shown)


rjwalters:
## Judge review: APPROVED

Security fix is correct, independently verified, and introduces no regressions. Approving.

### Verification performed

**Digest is correct — the guard still works.** The single highest-risk failure mode for this change is a wrong digest, which would silently block every deploy (or, worse, be "fixed" later by reverting to the literal). Computed independently:

```
$ printf '%s' '<the real account id>' | shasum -a 256
04e8eb2a99d37a555c87220b1d4cd1c018afbe5b65360bdd2224eb8e9f6b69c2
```

Exact match with the pinned `EXPECTED_ACCOUNT_ID_SHA256`.

**Behavioral test** via the script's own `WRANGLER_WHOAMI_OUTPUT` hook: correct account → exit 0; wrong account → exit 1 with the digest-prefix message; `--help` → exit 0. `bash -n` clean.

**No stale references to the removed variables.** `git grep -E 'EXPECTED_ACCOUNT_ID|EXPECTED_EMAIL'` (excluding the new `_SHA256`) returns nothing across the whole tree at head — the `set -u` unbound-variable regression described in the PR body is genuinely fixed, not just locally patched. Nothing sources this script, so the removal is contained.

**Redaction of the account ID is complete at HEAD.** `git grep` for the literal across the entire tree at `d691f938` returns zero hits.

### Findings (non-blocking, but please read #1)

**1. The digest does not make the ID secret — it is still in this public repo's history.** `git log -S<id> --all` matches 9+ commits. The preimage is one command away for anyone who clones, so the "128 bits of hex, so not brute-forceable" comment at `check-account.sh:19` is true in isolation but misleading in context: nobody has to brute-force it. What this PR actually buys is real and worth having — it stops further propagation-by-copying (the stated root cause), removes the ID from the surface people read and copy, and keeps it out of failure output and future clones' HEAD. But it does not undo the disclosure. Please treat the account ID as already-public: if it genuinely warrants protection, rotation is the only remedy; if it does not, that is fine, but the decision should be explicit. Suggest amending that comment to say so, so a future reader does not infer the ID is now confidential.

**2. Incomplete within the very hunk being redacted.** `check-token-registry.sh:24-25` and `sync-token-registry.sh:16-17` still carry `agent-1@2amlogic.com` / `agent-2@2amlogic.com`, directly beneath the three lines converted to `example.com`. I can see why they survived — they are the only examples demonstrating hyphen-stripping (`agent-1` -> `agent1`), which the new `example.com` lines do not cover — but `agent-1@example.com` teaches that rule identically without a real domain. As written, the file now applies two different rules to two adjacent lines of the same comment block.

**3. The "four leaks" enumeration is not exhaustive.** `r.j.walters@gmail.com` also survives at `research/problems/sqrt2-minpoly-oq-03/sessions/2026-05-16-s6-state-sync-disk-red-escalation-orphan-stash-flag.md:135`, inside a pasted `git show` output. Genuinely low severity — commit-author emails are intrinsically public in a public repo, so this is a different category from the Cloudflare ID — but it is neither redacted nor listed under "Deliberately unchanged". Worth a line in that section.

**4. Dead code that can fail silently.** `CURRENT_EMAIL` (`check-account.sh:77`) is now assigned and never read, but still executes under `set -euo pipefail`. Confirmed: with whoami output lacking the literal phrase `associated with the email`, the `grep` exits 1 and the script aborts with **exit 1 and no output whatsoever** — a false deploy-block with zero diagnostics. This is pre-existing, not introduced here, but the line now has zero remaining benefit, so deleting it is a free removal of a silent-failure path.

### Notes on gating

- `gh pr checks` reports *"no checks reported on the 'security/disclosure-sweep-identifiers' branch"*. This repo has **no CI configured on PR branches**, so that is the expected result — I am explicitly recording it as **absence of CI**, not as a green CI run. The verification above was run by hand to compensate.
- No `buildGate.command` is configured in `.loom/config.json`, so there was no project check command to run. The diff touches only shell, TOML, and Markdown — no application or Lean code.
- Merge state `CLEAN` / `MERGEABLE`, confirmed live immediately before this verdict.

The "Deliberately unchanged" section correctly scopes out `database_id` and `SUBMISSION_EMAIL`, and the reasoning for pinning rather than reading from the environment is sound: an expectation the environment can move is one a wrong environment moves with it. Items 1-4 are follow-ups, not merge blockers.

rjwalters:
**Champion Auto-Merge**

---
*Automated by Champion role*

PR #43695, opened 2026-08-07
perf(gallery): fix search debounce wiring, memoize cards, batch grid rendering

rjwalters:
Search on the gallery pages had grown slow. Two independent causes; both fixed here.

## 1. The debounce was wired to the wrong value

All three gallery pages used `useDebouncedUrlState` but destructured only the first two elements:

```ts
const [searchQuery, setSearchQuery] = useDebouncedUrlState('q', '', serializers.string)
```

It returns `[localValue, setValue, urlValue]` — `[0]` is the **live** value, `[2]` is the debounced one. So the 300 ms debounce deferred nothing but the URL rewrite. Filtering, sorting, grouping and rendering all still ran on **every keystroke** over 4,810 listings, and the 2.6 MB search index was fetched on the very first character typed.

- The text box keeps the live value (`searchInput`); the filter pipeline and the lazy index fetch key off the debounced `searchQuery`.
- **New `src/lib/gallery-search.ts`** — folds each entry's searchable fields into one pre-lowercased haystack, memoized on `[listings, searchIndex]`, instead of re-lowercasing every full description on each pass (~2.5 MB of throwaway strings). Index entries are already lowercased by `scripts/annotations/build.ts`; the truncated listing description stays the fallback.
- Precomputed numeric date sort keys and one cached `Intl.Collator`, so comparators no longer allocate two `Date`s or build a collator per comparison.
- `ErdosPage`: the Erdős subset is hoisted into its own memo, shared by the filter pipeline and the stats block.

## 2. Every card in the grid was mounted, always

The grids rendered one card per entry with no upper bound — 1,634 cards on the home gallery — so every filter change reconciled the whole grid, and unrelated state (filter panel toggle, "Copied!" flash) rebuilt every card's element tree.

- **New `useIncrementalList`** — renders the first 60 items and appends a batch as an IntersectionObserver sentinel scrolls into view. The count resets when the filtered list changes identity, so a new search starts at the top.
- **New `LoadMore` footer** — shows the remaining count plus an explicit **Show all** button, so find-in-page across the full result set is one click away.
- `GalleryCard` and `ResearchCard` are now `memo`'d; `ErdosPage`'s inline card is extracted to a memoized `ErdosGalleryCard`, and the page's private duplicate of `StatusBadge` is dropped in favour of the one `GalleryCard` already exports.

### Why batching and not windowed virtualization

The cards are variable-height and the grid's column count is responsive, which makes absolute-positioned windowing fragile; and keeping revealed cards mounted preserves native scrolling and in-page anchors. (Scroll position across back-navigation is *not* preserved — but it was already broken on `main` and is measurably less bad here; see the review comment and the follow-up issue.) It also adds no dependency, so the 600 KB chunk budget is untouched. The win is bounded work per filter change, not bounded memory over a long scroll.

## Verification

**Search results are unchanged.** Fields are joined with `\n`, which a single-line `<input type="text">` cannot contain, so no query can match across a field boundary. Checked against the real data (4,810 listings, 4,428 index entries) — the old per-field predicate and the new haystack return identical result sets for 15 probe queries covering diacritics (`Erdős`), uppercase (`GRAPH`), numerics (`124`), multi-word (`number theory`), single-character (`a` → 4,810, `q` → 3,207) and no-hit (`xyzzy`):

```
RESULT: all queries identical
old (lowercase in predicate): 3.01 ms/pass
new (precomputed haystack)  : 1.37 ms/pass
haystack build (once per data change): 7.09 ms
```

**Browser run** against a dev server with the real dataset (Playwright, system Chrome):

```
home: initial cards mounted = 60          (was 1,634)
home: load-more footer = "1,574 more proofs below"
home: after scroll = 120 (grew by 60)
home: after "Show all" = 1634
home: url during typing  = (no q param — debounced)
home: url after pause    = q=sidon
home: search "sidon" -> SHOWING 46 OF 4810 PROOFS
home: input value = "sidon"               (live value still shown per keystroke)
home: after clear -> 4810 PROOFS
erdos / research galleries: batch + footer OK
console errors: none
```

The 46 hits for `sidon` match the data-level equivalence check exactly.

**Static checks:** `tsc -b` and `eslint` clean for every changed file. Remaining repo-wide lint errors are all in files this PR does not touch.

`pnpm build` was **not** run — it regenerates thousands of gallery data files. The two `tsc` errors for `data-manifest.json` / `research-data-manifest.json` are gitignored artifacts of those skipped build steps, not regressions.


rjwalters:
✅ **Approved!**

**Disclosure:** this PR was authored in the same session that is now judging it. I therefore treated the author's own verification as unproven and re-derived it independently, and went looking specifically for defects the author would be motivated to miss. Findings below.

## What I verified independently

**1. Extended the equivalence check to the two variants the PR never tested.** The PR's evidence covered only `HomePage`'s predicate. `ErdosPage` adds an `erdosNumber` clause and `ResearchPage` has optional `tags` — both were refactored on the strength of an argument, not a measurement. Re-ran old-predicate vs new-haystack over the real data for both:

```
ErdosPage:    1354 listings — 11/11 queries identical  (incl. "124", "728", "1", "0", "99999")
ResearchPage: 2992 listings — 11/11 queries identical
RESULT: all variants identical
```

The `"1"` / `"0"` cases matter: they are the ones where folding `erdosNumber` into a joined haystack could plausibly have changed digit matching. It does not.

**2. Re-ran the browser check.** 60 cards initially (was 1,634), sentinel appends one batch per scroll, "Show all" → 1,634, no console errors, and the debounce confirmed at the URL level (no `q` param mid-typing, `q=sidon` after the pause) while the input still shows every keystroke.

**3. Static checks.** `tsc -b` and `eslint` clean for all 11 changed files. The remaining repo-wide lint errors are all in files this PR does not touch (`ProofPage`, `ResearchProblemPage`, `LineageBreadcrumb`, …).

## Findings

**① PR description overclaimed — fixed directly.** The rationale section asserted that keeping cards mounted "preserves native scrolling, in-page anchors and scroll restoration." Scroll restoration is **not** preserved. I have corrected that sentence in the PR body rather than bounce the PR for a doc fix.

**② Back-navigation scroll loss — pre-existing, not a regression. Filed #43696.** I suspected the batching would break gallery → proof → Back, so I measured it on both branches:

| Branch | scrollY before | after Back | drift |
|---|---|---|---|
| `main` (aefd82e) | 60,000 | 232,558 | **172,558 px** |
| this PR | 35,967 | 43,727 | 7,760 px |

Already broken on `main`, and this PR makes it ~22× *less* bad (the sentinel refills until the restored offset is reachable instead of snapping to the bottom of a fully-mounted page). Not blocking; tracked in #43696 with the measurements and a suggested fix.

**③ No CI on this branch.** `gh pr checks` reports *no checks reported* — so the usual CI gate is vacuous here and the local `tsc`/`eslint`/browser runs above are the only signal. Flagging so this approval is not read as "CI green."

## Notes

- No `Closes #N` — correct in this case: user-requested work with no originating issue, not a sloppy reference.
- `useIncrementalList` resets its count during render (`prevItems !== items`) rather than in an effect. This is the documented React pattern and avoids the `react-hooks/set-state-in-effect` rule the codebase enforces — good call over an `eslint-disable`.
- Dropping `ErdosPage`'s private `StatusBadge` duplicate in favour of the exported one is a genuine dedup, not incidental churn; `ErdosPage` nets ~140 lines smaller.

Merge state CLEAN. Approving.

rjwalters:
**Champion Auto-Merge**

This PR meets all safety criteria for automatic merging:

- Judge approved (`loom:pr` label), no contradictory `loom:changes-requested` — verdict-state janitor passed
- Merge-risk judgment passed on all four axes:
  - **Diff composition**: load-bearing hunks are few and enumerable — the filter pipeline in three gallery pages, one new hook, and `memo()` wrappers. Two of the eleven files are new self-contained modules; two are barrel re-exports.
  - **Blast radius**: confined to gallery UI. Touches nothing that mediates merging, worktrees, credentials, guard hooks, installers, CI workflows, or config schema.
  - **Judge review depth**: the verdict cites named files and concrete behavior, and independently re-ran equivalence for the two page variants the PR description had not covered.
  - **Revertability**: `git revert <squash-sha>` fully undoes it — no migration, no published artifact, no state written outside the repo.
- Diff size: 762 lines (+540/-222) — informational, not a gate
- No critical files modified (verified against the paginated `pulls/43695/files` list, 11 files, zero pattern matches)
- No merge conflicts (`mergeable: MERGEABLE`)
- Updated minutes ago, well inside the 24h recency window
- No CI checks required — this branch has none configured, so the Judge's local `tsc`/`eslint`/browser verification is the operative signal (called out explicitly in the Judge verdict)

**Proceeding with squash merge...** If this was merged in error, you can revert with:
`git revert <commit-sha>`

---
*Automated by Champion role*

PR #43697, opened 2026-08-08
fix(gallery): make search match slugs, diacritics and hyphenated tags

rjwalters:
## Problem

Searching **`erdos 85` returns no hits.** So does `erdos-85` — the slug users see in the URL and the name people actually use for the entry.

This is **pre-existing, not a regression from #43695**. I verified the old and new predicates return identical sets for every probe query, including the failing ones.

The entry looks like this:

```
slug        : erdos-85
title       : 'Erdős Problem #85: Minimum Degree for 4-Cycles'
tags        : ['erdos', 'graph-theory', 'extremal', '4-cycle', ...]
erdosNumber : 85
```

Three independent causes:

1. **No separator normalization.** Matching is a raw substring test, so `erdos 85` must appear as one contiguous run. The title is `Erdős Problem #85` — it never does.
2. **Diacritics.** The title has `Erdős` (U+0151 `ő`); the query has `erdos`. The single-word query `erdos` only worked by luck, via the ASCII `erdos` *tag*.
3. **The slug was never searchable.** It isn't among the fields fed to the haystack.

## Fix

Add `normalizeSearchText()` — NFD, strip combining marks, lowercase, collapse every run of non-alphanumerics to a single space — and apply it to **both** the haystack fields and the query, so the two meet in the middle. Add `slug` to the searchable fields on all three gallery pages.

```
"Erdős Problem #85"  ->  "erdos problem 85"
"erdos-85"           ->  "erdos 85"
"number-theory"      ->  "number theory"
"π-approximation"    ->  "π approximation"
```

**Matching stays a contiguous substring test**, so a multi-word query is still a *phrase* search. I prototyped the obvious alternative — AND-tokenizing the query — and **rejected it**: it blew `number theory` up from 42 to **2,049** hits by matching both words anywhere in long descriptions, destroying phrase precision. Separator normalization fixes the reported bug without that cost.

Uses `\p{L}\p{N}` rather than `[a-z0-9]` so Greek letters in mathematical titles (and CJK) survive instead of being erased to an empty string.

## Verification (real data, 4,810 listings)

| query | old | new | `erdos-85` found |
|---|---|---|---|
| `erdos 85` | 0 | **12** | no → **yes** |
| `erdos-85` | 0 | **12** | no → **yes** |
| `erdős 85` | 0 | **12** | no → **yes** |
| `ERDOS 85` | 0 | **12** | no → **yes** |
| `sidon` | 46 | 46 | — |
| `graph` | 550 | 550 | — |
| `xyzzy` / `   ` / `---` | 0 | 0 | — |
| `number theory` | 42 | 1,710 | — |

The one large change is `number theory`, and it is **correct**: the tag `number-theory` is carried by **1,641 listings** and now matches the spaced query, which it never did before. Sampled hits are all genuinely number-theory entries (`abel-ruffini-oq-03`, `a054656-distinct-prime-partition`, …).

**Phrase-semantics guard** — a synthetic entry titled `Number of primes` with description `a theory of sets` still does **not** match `number theory`. Words must remain adjacent.

`tsc -b` and `eslint` clean on all four changed files.

### DIFF
--- a/src/lib/gallery-search.ts
+++ b/src/lib/gallery-search.ts
@@ -38,19 +38,53 @@ export function compareTitles(a: string, b: string): number {
 }
 
 /**
- * Fold each entry's searchable fields into one pre-lowercased haystack string,
+ * Normalize text for search: strip diacritics, lowercase, and collapse every
+ * run of non-alphanumeric characters to a single space.
+ *
+ * Applied to BOTH the haystack fields and the query, so the two meet in the
+ * middle. Without it, three natural searches all returned nothing:
+ *
+ * - `erdos 85` — the title is `Erdős Problem #85`, so `erdos 85` never appears
+ *   as a contiguous run. Folding `ő`→`o` and turning `#` into a space makes it
+ *   one.
+ * - `erdos-85` — the slug, i.e. the identifier in the URL and the name people
+ *   actually use, normalizes to the same `erdos 85`.
+ * - `number theory` — previously only matched the literal spaced phrase, never
+ *   the `number-theory` *tag* carried by 1,641 entries.
+ *
+ * Punctuation is collapsed rather than deleted so `4-cycle` becomes `4 cycle`
+ * (two tokens) instead of `4cycle`. Matching stays a plain contiguous substring
+ * test, so a multi-word query is still a *phrase* search — `number theory` does
+ * not degenerate into "any entry containing both words somewhere".
+ *
+ * Uses the Unicode property escapes `\p{L}\p{N}` rather than `[a-z0-9]` so
+ * non-Latin scripts (Greek letters in mathematical titles, CJK) survive instead
+ * of being erased to an empty string.
+ */
+export function normalizeSearchText(value: string): string {
+  return value
+    .normalize('NFD')
+    // Strip combining marks left by NFD: ő -> o, é -> e, ü -> u.
+    .replace(/[\u0300-\u036f]/g, '')
+    .toLowerCase()
+    .replace(/[^\p{L}\p{N}]+/gu, ' ')
+    .trim()
+}
+
+/**
+ * Fold each entry's searchable fields into one normalized haystack string,
  * keyed by slug.
  *
- * Fields are joined with a newline, which no single-line search input can
- * contain — so a query can never match across a field boundary and the result
- * set is identical to per-field substring matching.
+ * Each field is normalized independently (see {@link normalizeSearchText}) and
+ * the results are joined with a newline. Normalization can never *produce* a
+ * newline, so a query can never match across a field boundary — the result set
+ * stays equivalent to per-field matching.
  *
  * `undefined`, `null` and empty fields are dropped. Numeric fields (e.g. an
- * Erdős problem number) are stringified, matching the previous
- * `.toString().includes(query)` behaviour.
+ * Erdős problem number) are stringified.
  *
  * Note that entries coming from the prebuilt search index are *already*
- * lowercased (`scripts/annotations/build.ts`); lowercasing again here is
+ * lowercased (`scripts/annotations/build.ts`); normalizing again here is
  * idempotent and costs nothing at steady state since this runs once per data
  * change.
  */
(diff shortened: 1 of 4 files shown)


rjwalters:
✅ **Approved!**

**Disclosure:** authored and judged in the same session, so I re-derived the claims independently rather than trusting the PR body, and went looking for what the author would want to skip.

## Verified independently

**The regression question — is this #43695's fault?** No. I ran the pre-#43695 predicate and the post-#43695 predicate side by side over all 4,810 listings for every probe query. They return **identical sets**, including on the failing ones (`erdos 85` → 0 for both). The bug predates the perf work; the perf PR neither caused nor masked it.

**The rejected alternative is the interesting part.** The obvious fix is to AND-tokenize the query. The PR says it was rejected; I confirmed the number — it takes `number theory` from 42 to **2,049** hits by matching both words anywhere in long descriptions. Separator normalization gets the reported bug fixed at 1,710, and that 1,710 is *justified* rather than incidental: the `number-theory` tag is carried by **1,641 listings** and legitimately matches the spaced query now. Sampled the delta — `abel-ruffini-oq-03`, `a054656-distinct-prime-partition`, etc. — all genuinely number-theory entries. This is the difference between a recall fix and a precision regression, and the PR picked the right side.

**Phrase semantics survive.** The synthetic guard (`Number of primes` / `a theory of sets`) still does not match `number theory`. Adjacency is preserved, which is the whole reason to prefer this over tokenization.

**Unicode handling is right.** `\p{L}\p{N}` rather than `[a-z0-9]` — I checked that `π-approximation` normalizes to `π approximation` and not to the empty string. A `[a-z0-9]` class would have silently deleted Greek-lettered titles from the index entirely. Good catch by the author; this is the kind of thing that fails quietly.

## Finding — one-time build cost regressed 9×, disclosed here

Not in the PR body, so recording it:

```
haystack build (4810 entries, ~2.5MB):  7 ms  ->  62.9 ms
filter pass (per debounced keystroke):          3.65 ms
normalizeSearchText(query):                     0.16 µs
```

Normalizing 2.5 MB through a Unicode regex is the cost. It is paid **once per data change** — on listings load, and again when the lazy search index arrives — not per keystroke, and it sits behind the 300 ms debounce from #43695. A 63 ms one-off on the main thread is a real hitch but below the threshold where it reads as a stall, and it buys a search that actually works. Non-blocking, but it should not be allowed to grow: if another field joins the haystack, this wants moving off the main thread or precomputing at build time.

## Minor note (not blocking)

`hasActiveFilters` still gates on `searchQuery.trim()` while filtering now gates on `normalizeSearchText(searchQuery)`. A punctuation-only query like `---` is therefore "active" for the header/Share-button but a no-op for the filter, so the page reads `Showing 4810 of 4810`. Cosmetic, self-correcting on the next keystroke, and not worth a round-trip.

## Notes

- No `Closes #N` — correct; user-reported bug with no originating issue.
- Normalization is applied per-field before the `\n` join, so a query (which can never contain `\n` after normalization) still cannot match across a field boundary. The invariant from #43695 holds.
- `tsc -b` and `eslint` clean on all four changed files. No CI is configured on this branch, so those plus the data-level checks are the only signal — flagging so this is not read as "CI green."

Merge state CLEAN. Approving.

rjwalters:
**Champion Auto-Merge**

This PR meets all safety criteria for automatic merging:

- Judge approved (`loom:pr`), no contradictory `loom:changes-requested` — verdict-state janitor passed
- Merge-risk judgment passed on all four axes:
  - **Diff composition**: 4 files, one new pure function plus three call-site updates. Load-bearing hunks are enumerable and were verified against the real dataset.
  - **Blast radius**: gallery search only. No merge/worktree/credential/guard/installer/CI/config surface.
  - **Judge review depth**: verdict independently re-derived the regression question, the rejected-alternative numbers, and the Unicode handling; disclosed a 9× one-time build-cost regression the PR body omitted.
  - **Revertability**: pure `git revert` — no migration, no artifact, no external state.
- No critical files modified (paginated `pulls/43697/files`, 4 files, zero matches)
- No merge conflicts (`MERGEABLE`)
- Updated minutes ago
- No CI checks configured on this branch — local `tsc`/`eslint` plus data-level verification are the operative signal

**Proceeding with squash merge...** Revert with `git revert <commit-sha>` if needed.

---
*Automated by Champion role*

